PR #53, opened 2014-09-06
Sorted dict 1

StephenVavasis:
Pull request for SortedDict and BalancedTree23 taking into account Kevin Squire's feedback from the past 12 days.


aviks:
Replaced by #70

PR #68, opened 2014-12-30
Implement nlargest and nsmallest

cpitclaudel:
Python's heapq module offers two convenient functions, [`nlargest` and `nsmallest`](https://docs.python.org/2/library/heapq.html#heapq.nlargest). This patch replicates part of that functionality (the `key` argument is missing − it should probably be a feature of the underlying heap). 

For performance reasons I added an `extract_all_rev!` method, but since `extract_all` wasn't documented in `README.md` I didn't document it either. I also added relevant tests in `test_binheap.jl`.

### DIFF
--- a/src/heaps.jl
+++ b/src/heaps.jl
@@ -72,3 +72,58 @@ function extract_all!{VT}(h::AbstractHeap{VT})
     end
     r
 end
+
+function extract_all_rev!{VT}(h::AbstractHeap{VT})
+    n = length(h)
+    r = Array(VT, n)
+    for i = 1 : n
+        r[n + 1 - i] = pop!(h)
+    end
+    r
+end
+
+# Array functions using heaps
+
+function nextreme{T, Comp}(comp::Comp, n::Int, arr::AbstractVector{T})
+    if n <= 0
+        return T[] # sort(arr)[1:n] returns [] for n <= 0
+    elseif n >= length(arr)
+        return sort(arr, lt = (x, y) -> compare(comp, y, x))
+    end
+    
+    buffer = BinaryHeap{T,Comp}(comp)
+    
+    for i = 1 : n
+        @inbounds xi = arr[i]
+        push!(buffer, xi)
+    end
+    
+    for i = n + 1 : length(arr)
+        @inbounds xi = arr[i]
+        if compare(comp, top(buffer), xi)
+            # This could use a pushpop method
+            pop!(buffer)
+            push!(buffer, xi)
+        end
+    end
+
+    return extract_all_rev!(buffer)
+end
+
+@doc """
+Returns the `n` largest elements of `arr`.
+
+Equivalent to `sort(arr, lt = >)[1:min(n, end)]`
+""" ->
+function nlargest{T}(n::Int, arr::AbstractVector{T})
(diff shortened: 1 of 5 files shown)


aviks:
Should this be merged?

cpitclaudel:
@aviks I think it could be merged!

src/heaps.jl
```diff
@@ -72,3 +72,58 @@ function extract_all!{VT}(h::AbstractHeap{VT})
...
+        r[n + 1 - i] = pop!(h)
+    end
+    r
+end
+
+# Array functions using heaps
+
+function nextreme{T, Comp}(comp::Comp, n::Int, arr::AbstractVector{T})
+    if n <= 0
+        return [] # sort(arr)[1:n] returns [] for n <= 0
```
hayd: Perhaps this should this be `eltype(arr)[]` ?

kmsquire: Definitely, yes.

cpitclaudel: Wouldn't `T[]` work just the same? I've pushed that.

kmsquire:
@cpitclaudel, thanks--that works better, of course.  

Right now, this doesn't actually merge cleanly, so it would need to be rebased against the current master.  However, note that similar functionality exists in base as the `select`/`select!` function, which uses an algorithm based on quicksort.

```julia
julia> select([0,21,-12,68,-25,14], 1:3)
3-element Array{Int64,1}:
 -25
 -12
   0

julia> select([0,21,-12,68,-25,14], 1:3, rev=true)
3-element Array{Int64,1}:
 68
 21
 14
```

`select!` is the same, except that it rearranges the array in place (and is therefore more efficient).

I'm (slightly) concerned about duplicating functionality in base with different names but the same (or similar) functionality--at the least, I'd rather converge on the same name/set of names.  To me, the function names here would seem clearer to anyone who isn't familiar with [`quickselect`](http://en.wikipedia.org/wiki/Quickselect).  But that might just depend on your background.

On another note, because of the heap construction, I would guess that this algorithm is slightly slower than `select` (and `select!`), although that would have to be tested.

@StefanKarpinski, any thoughts?

kmsquire:
Actually, I was somewhat wrong about the speed (which is why it's always good to test).  For getting just a few elements from large arrays, this algorithm is faster than `select` (see https://gist.github.com/kmsquire/6ea27036b815c9a56101).

So, now it's just naming.  If no one pipes up with some thoughts soon, I'll merge.

kmsquire:
(After it's rebased, of course.)

cpitclaudel:
I'll rebase it now; the original names were inspired by the corresponding Python functions.

cpitclaudel:
All done!

kmsquire:
I'm going to merge this, but I'm thinking of the following:

1. creating `heapselect` and `heapselect!` functions, which matches the signature and features of `select`/`select!`.
2. have `nlargest` and `nsmallest` call either of these functions, based either on problem size or user-selection.

Thoughts?  If this is okay, I'll create an issue, and may try to implement something.

cpitclaudel:
This sounds great. I think the heap class could also use a `pushpop` method; that's what the corresponding Python method uses. I might look into it (I left a comment in the source)

PR #69, opened 2014-12-30
Get coverage data generation back up for Coveralls

garborg:
Perhaps better to update the default Travis script, but for now.


garborg:
Bump. (Since the emails come whether or not coveralls has a number to return.)

Issue #67, opened 2014-12-23
OrderedDict and OrderedSet constructor does not work properly on variable tuple tupes


yuyichao:
Best shown with the example below.

```julia
julia> Set{(Symbol...)}([(:a,)])
Set((Symbol...,)[(:a,)])

julia> OrderedSet{(Symbol...)}([(:a,)])
ERROR: `convert` has no method matching convert(::Type{(Symbol...,)}, ::Array{(Symbol,),1})
Closest candidates are:
  convert(::Type{T}, ::T)
  convert(::(Type{T<:Top},Type{T<:Top}...), ::(Any,Any...))
  convert(::Type{Ptr{Void}}, ::Array{T,N})
  ...

 in setindex! at /usr/share/julia/site/v0.4/DataStructures.jl/src/hashdict.jl:366
 in union! at /usr/share/julia/site/v0.4/DataStructures.jl/src/orderedset.jl:31
 in call at /usr/share/julia/site/v0.4/DataStructures.jl/src/orderedset.jl:12

julia> Dict{Symbol, (Symbol...)}(:a => (:a,))
Dict{Symbol,(Symbol...,)} with 1 entry:
  :a => (:a,)

julia> OrderedDict{Symbol, (Symbol...)}(:a => (:a,))
ERROR: `convert` has no method matching convert(::Type{DataStructures.OrderedDict{Symbol,(Symbol...,)}}, ::Pair{Symbol,(Symbol,)})
Closest candidates are:
  convert(::Type{T}, ::T)
  convert(::Type{Nullable{T}}, ::T)
  convert(::Type{Scical.optics.Focus}, ::Any)
  ...

 in call at ./base.jl:34
```

kmsquire:
Fixed in the latest release.  Sorry this took so long!

Issue #64, opened 2014-10-25
Dict-like structures' constructors should track Dict changes 


garborg:
(I'm guessing.)
Leave them in, if `@deprecate` depending on `VERSION`, add missing methods?

kmsquire:
I believe this is addressed for `OrderedDict`s and `OrderedSet`s, but not yet for `DefaultDicts`.

PR #77, opened 2015-03-04
RFC: Update OrderedDict, OrderedSet constructors to take iterables (fixes #50, #64, #67)

### LINKED ISSUE #50: Make OrderedSet constructor work on iterables, like base
For v0.3, the `Set` interface has been changed, so that `Set`s are now constructed from iterables.  `OrderedSet` should make the same change.

(Need to check if this is true for `Dict`s as well.)

```julia
julia> a = [1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> Set(a)
Set{Int64}({2,3,1})

julia> OrderedSet(a)
OrderedSet{Array{Int64,1}}([1,2,3])
```

This last one should be:

```julia
julia> OrderedSet(a)
OrderedSet{Int64}(1,2,3)
```

Assigning myself, but I may not get to it soon, so anyone else should feel free to change this.  It should be an easy change.

kmsquire:
* Plus minor related changes

Eventually, these data structures might be made moot, if https://github.com/JuliaLang/julia/pull/10116 is merged.  But for now, it would at least be worthwhile matching `Dict` and `Set` construction in base in Julia v0.3.

I plan to merge this in a day or two, pending any comments.

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -5,7 +5,7 @@ module DataStructures
     import Base: length, isempty, start, next, done,
                  show, dump, empty!, getindex, setindex!, get, get!,
                  in, haskey, keys, merge, copy, cat,
-                 push!, pop!, shift!, unshift!,
+                 push!, pop!, shift!, unshift!, insert!,
                  union!, delete!, similar, sizehint,
                  isequal, hash,
                  map, reverse,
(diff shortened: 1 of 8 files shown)


kmsquire:
I guess I should add some more tests...

kmsquire:
That's all I plan to do for now.  I added tests for `OrderedDicts` and `OrderedSets`, mostly by copying and modifying `Dict` and `Set` tests from `Base`.  This actually uncovered a couple of bugs (tests are useful!), which I fixed as well.

This is a breaking change, in the sense that some constructors for the old version of `OrderedDicts` and `OrderedSets` no longer work.  This does bring those mostly up to par with the constructors in v0.3, though, so I doubt many people will be affected.

At any rate, I'm going to bump the minor number, as opposed to the patch number, and make an announcement on julia-users.

kmsquire:
Actually, I take that back--this isn't a breaking change.  The old functionality is deprecated, and the new functionality is available.  Does that warrant a bump in minor number?

kmsquire:
I just bumped the patch.

PR #78, opened 2015-03-04
WIP: Fix #34, implement `==` instead of `isequal` in places

### LINKED ISSUE #34: provide == as well as isequal for Julia 0.3
Currently, you are providing an `isequal` method for various types.   In Julia 0.3, you will *also* need to provide `==`.  See JuliaLang/julia#6833.

(In Julia 0.2, `==` called `isequal` by default, so providing `isequal` was enough.  In Julia 0.3, however, these two functions are swapped: `isequal` calls `==` by default.  So overriding `isequal` is not enough to make things like `!=` or `==` work.   In order to remain backward-compatible with Julia 0.2, you will need to provide *both* methods.)

garrison:


lostanlen:
Looks good to me. Let's merge it eventually.

kmsquire:
@garrison, can this be merged as is, even if it's incomplete?

garrison:
@kmsquire, sure!

kmsquire:
Sorry, I should have actually looked at the PR, but I was on my phone... it is pretty trivial. ;-)

Issue #83, opened 2015-04-15
Proposal: SuffixTree


gideonite:
I'm implementing suffix trees in Julia for fun and was wondering whether I should cleanup the code and give it a home here.

kmsquire:
I think it depends.  I think it would be great to have an implementation of suffix trees in DataStructures.  For it to make it in to DataStructures generally requires someone to do code review.  I try to do so, and others do as well, but most of us are pretty busy, so review can sometimes take a while.

Anyway, I would encourage you to make a pull request, but to keep the process smooth:

1. Try to create and update the pull request in manageable chunks.  It's easier to follow along as you make changes that way.  Dumping a large body of code all at once generally makes code review challenging.
2. If you want review of the code, say so in the pull request or in comments as you progress.  If you don't get a response, bump the thread after a day or two.
3. You might know this, but in Julia development, marking a PR as `WIP:` (work in progress) generally means that you aren't ready for code review, but let's interested parties follow along. Changing the title to `RFC` (request for comments) says it's ready for review (but for this repo, as suggested in 2, it's probably a good idea to bump the thread or add a comment saying that you want review).

Anyway, looking forward to seeing what you have!

oxinabox:
Since #84 is abandoned and closed.
I think this issue should be also.

gideonite:
I'm sorry for going AWOL on this. Aside from my not having pushed this
through, is there any interest in seeing this get merged at some point?

Thanks!

kmsquire:
Hi @gideonite, as with my previous discussion in #84, my time is pretty limited, although others might have more time to review.  

If you're interested in updating #84, it can be reopened, or you can open up a new pull request.  I do think a suffix tree would be a nice data structure to include here.

Be aware that there has been some discussion about splitting up DataStructures.jl, although that work hasn't started yet.

PR #84, opened 2015-04-17
WIP: getting started with a SuffixTrie.

gideonite:
Hopefully to become a SuffixTree. The plan is to do this by compressing the SuffixTrie once it has been created. See #83

### DIFF
--- a/src/suffixtree.jl
+++ b/src/suffixtree.jl
@@ -1,0 +1,112 @@
+using Base.Test
+import Base.insert!
+
+abstract SuffixTrie
+
+type SuffixTrieNode <: SuffixTrie
+  value
+  children::Array{SuffixTrieNode}
+  index::Int64
+end
+
+SuffixTrieNode(value, children) = SuffixTrieNode(value, children, -1)
+SuffixTrieNode(value) = SuffixTrieNode(value, [], -1)
+
+type SuffixTrieRoot <: SuffixTrie
+  children::Array{SuffixTrieNode}
+end
+
+SuffixTrieRoot() = SuffixTrieRoot([])
+
+function is_leaf(node::SuffixTrieNode)
+  return node.index == -1
+end
+
+function is_prefix_match(s1::String, s2::String)
+  if length(s1) == length(s2)
+    return s1 == s2
+  elseif length(s1) < length(s2)
+    return s1 == s2[1:length(s1)]
+  else
+    return s2 == s1[1:length(s2)]
+  end
+end
+
+function find_match(nodes::Array{SuffixTrieNode}, suffix::String)
+  for node in nodes
+    if is_prefix_match(string(node.value), suffix)
+      return node
+    end
+  end
+
+  return None
+end
+
+function find_or_create_next_node(nodes::Array{SuffixTrieNode}, suffix::String)
+  next = find_match(nodes, suffix)
+  if next == None
+    next = SuffixTrieNode(suffix[1])
+    push!(nodes, next)
+  end
+  next
+end
+
+function insert!(suffix_trie::SuffixTrieRoot, suffix::String, index::Int)
+  next = find_or_create_next_node(suffix_trie.children, suffix)
+  insert!(next, suffix[2:end], index)
+end
+
+function insert!(node::SuffixTrieNode, suffix::String, index::Int)
+  if isempty(suffix)
+    node.index = index
+  else
+    next = find_or_create_next_node(node.children, suffix)
+    insert!(next, suffix[2:end], index)
+  end
+end
+
+function construct_suffix_trie(text::String)
+  text = string(text, '$')
+  suffixes = [text[i:end] for i in 1:length(text)]
+
+  ret = SuffixTrieRoot()
+
+  for (suffix, index) in zip(suffixes, 1:length(suffixes))
+    insert!(ret, suffix, index)
+  end
+
+  ret
+end
+
+construct_suffix_trie("abc")
+
+for node in construct_suffix_trie("panamabanana").children
+  println(node)
+end
+
+##
+## Tests
+##
+
+function type_tests()
+  leaf = SuffixTrieNode(10, [])
+  @test is_leaf(leaf) == true
+end
+
+type_tests()
+
+function construct_suffix_trie_test()
+#   @test BioTrieModule.BioTrieRoot == typeof(construct_suffix_trie("abcd123"))
+end
+
+construct_suffix_trie_test()
+
+function is_prefix_match_test()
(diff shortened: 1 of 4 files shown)


kmsquire:
Thanks for starting this!  Maybe obvious, but just some quick immediate feedback: the test code should be moved its their own file under `tests` (and `test/runtests.jl` should be updated to run it).

src/suffixtree.jl
```diff
@@ -0,0 +1,112 @@
+using Base.Test
+import Base.insert!
+
+abstract SuffixTrie
+
+type SuffixTrieNode <: SuffixTrie
+  value
```
kmsquire: For types, any value without a type defaults to `Any`.  It would be better to parameterize both `SuffixTrie` and `SuffixTrieNode` (i.e., as `SuffixTrie{T}` and `SuffixTrieNode{T}`) and make this `value::T`.

src/suffixtree.jl
```diff
@@ -0,0 +1,112 @@
...
+  children::Array{SuffixTrieNode}
```
kmsquire: Without a dimension `Array{SuffixTrieNode}` is actually an abstract type, and in general, it's almost always better to use concrete types for members of types.  So this should either be `Vector{SuffixTrieNode}` or `Array{SuffixTrieNode,1}` (which mean exactly the same thing).

src/suffixtree.jl
```diff
@@ -0,0 +1,112 @@
...
+  index::Int64
+end
+
+SuffixTrieNode(value, children) = SuffixTrieNode(value, children, -1)
+SuffixTrieNode(value) = SuffixTrieNode(value, [], -1)
+
+type SuffixTrieRoot <: SuffixTrie
+  children::Array{SuffixTrieNode}
```
kmsquire: The `Array` comment above applies here (as does the parameterization comment)

src/suffixtree.jl
```diff
@@ -0,0 +1,112 @@
...
+end
+
+SuffixTrieRoot() = SuffixTrieRoot([])
```
kmsquire: In Julia 0.3, `[]` is an `Array{None,1}` (which is a totally useless type).  In v0.4, it's `Array{Any,1}`, which is more useful, but it might be good to change this if/when you change the types to be parameterized.

gideonite:
Now that I'm looking more at the project, I think that it makes more sense for me to basically compress the already existing `Trie

kmsquire:
General comment: for anything belonging to JuliaLang, at least (and most other packages), the style is to use 4 spaces.

Adding functionality to the existing `Trie` would be welcome.  I haven't looked at that code, but would still be willing to review.

BTW, even if you jump right to modifying or extending `Trie`s, this is a very good first pull request--just the right size, and easy to review.

gideonite:
@kmsquire Sorry, I clicked comment a bit too eagerly. I updated the comments above with a longer comment.

Ah I didn't know that. Okay, I can easily reformat.

Excellent, I'll see if reading through the code inspires any changes.

wlbksy:
would save more space if the ``ties`` is constructed with ``Double-Array Trie`` 
reference url : http://linux.thai.net/~thep/datrie/datrie.html

kmsquire:
> I see your point about the parameterized types. I definitely will do that, but are Suffix Trees ever used for datatypes other than string? Not from what I've been able to find.

They're sometimes used in DNA sequencing.  In general these would also be strings, but e.g., in BioSeq.jl, they're not, mainly because the author wanted easy mutation, and Strings in Julia are immutible.

> So that the `SuffixTree` data structure is just a wrapper around the `Trie`. Should `Suffixtree <: Trie`? There is an extra property which is specific to leaf nodes: the suffix's index in the string.

This is a point where Julia differs from a lot of other languages: `SuffixTree` can certainly wrap a `Trie`,  but the way the type hierarchy works, it can't be a subclass of `Trie` (you can only have subclasses of abstract types).  Sometimes this seems restrictive (to me), but my understanding is that allowing general subclassing would make the implementation a bit more complex and possibly cause problems with multiple dispatch.

One way to work around this is to create an `AbstractTrie` type, and define functions common to all `Trie`s on this type.  Another option is to wrap and delegate (see, e.g., ordereddict.jl and delegate.jl in this package).  In this case, the wrapping class and the wrapped class still have a common superclass.

(I wrote a few classes using delegation here, but am still somewhat dissatisified with that solution, and would love to find a different/better solution.  Still, it works.)

gideonite:
I'm stuck on whether I should:
* add some sort of `compress` function to `Trie` or
* whether I need to create a new file for `SuffixTree` as a new data structure that implements an "on-line" suffix tree construction, see [this] (http://www.cs.helsinki.fi/u/ukkonen/SuffixT1withFigs.pdf).

I could start by adding `compress` but does this addition even make sense? How often do people want to construct a Trie and then compress it?

Another interesting feature coming out of this discussion is extending `Trie` to support multiple data types for keys. Right now it only supports `String`. For example, perhaps it would be interesting to implement `insert!` for `Array{Nucleotide}`.

gideonite:
I've decided to go for adding compress functionality to the current Trie code. In order to do this I need to make it more generalized so that the underlying children map, `children::Dict{Char, Trie{T}}`, becomes `children::Dict{K,Trie{K,T}}`.

Once this is done I'll have to implement a number of functions for `Trie{String, T}` and then I'll be able to write a function `compress(trie::Trie{Char,T})` which returns a trie of type `Trie{String,T}`.

This also opens up the field for others to add their own Trie types. For example, `Trie{Nucleotide, T}`.

gideonite:
So I've started generalizing Trie to take two parameter types instead of one. Now Trie is of type `Trie{K,V}` where `K` defaults to `Char` for now. The next step is to build out the code for the type `Trie{String, V}` which in fact corresponds to a compressed Trie since the string key values are created by concatenating the char string values when there is only one child.

My next step is to implement `compress(Trie{Char, V}) -> Trie{String,V}`

kmsquire:
Thanks for the continued updates!  I've been following, although my Julia
time has been limited to short spurts recently.

test/test_trie.jl
```diff
@@ -1,7 +1,7 @@
-using DataStructures
+include("../src/trie.jl")
```
hayd: The include should happen in src/DataStructures.jl :)

There's also some tests and stuff still in the suffixtrie file.

SuffixTrie is a neat datastructure, would be great to have!

kmsquire:
This seems to have been forgotten.  Closing, but feel free to ping back here or open another PR in the future.

PR #90, opened 2015-05-02
add Multidict

wildart:
Multi-valued dictionary implementation based on `Dict{K,Vector{V}}`. It behaves as normal dictionary only
* `setindex!` appends new elements to the value vector
* `pop!` removes the last element from the value vector & if the value vector is empty removes key

Some additional functions were added
* `enumerateall` creates enumerator through all (k,v) pairs in the  dictionary
* `count` counts all values in the dictionary ( `length` returns number of keys)
* `size` returns number of keys & values in the dictionary

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -43,6 +43,8 @@ module DataStructures
     export excludelast, tokens
     export orderobject, Lt
 
+    export MultiDict
+
 
     if VERSION < v"0.4.0-dev"
         using Docile
@@ -67,6 +69,8 @@ module DataStructures
     include("list.jl")
     include("balancedTree.jl")
     include("tokens.jl")
+
+    include("multidict.jl")
 
     import .Tokens: Token, IntSemiToken, semi, container, assemble
     import .Tokens: deref_key, deref_value, deref, status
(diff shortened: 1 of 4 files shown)


wildart:
I wonder maybe `size` would be better way present element count information, so it would return

    (#key, #value) =  size(d)

IainNZ:
It'd be good to somehow knock this and #88 at the same time

StephenVavasis:
@IainNZ, @wildart, @kmsquire 

I can add analogous functions to mimic this functionality to SortedMultiDict, but I have a few concerns:

m[k] = newitem: the proposal here is that this operation acts like a 'push!' instead of an assignment statement, which seems to me inconsistent with the use of assignment elsewhere in Julia.  Perhaps '+=' instead of '='.  What do other readers think?

MultiDict <: Associative: MultiDict does not implement the Associative interface, so I think this subtype declaration should be omitted

malmaud:
Ping @wildart, would be useful to have this. I vote for using `insert!(m, k, newitem)` (as SortedMultiDict does)  instead of `m[k]=newitem` to be more consistent with expectations of assignment.

malmaud:
Superceded by https://github.com/JuliaLang/DataStructures.jl/pull/115

PR #99, opened 2015-06-05
On 0.4 allow SortedDicts to be constructed like regular Dicts

Keno:


kmsquire:
Cc: @StephenVavasis

StephenVavasis:
@Keno and @kmsquire,

I can merge this into my current PR (#88) for the new datatypes SortedSet and SortedMultiDict, but can you send me some example invocations of the new SortedDict constructors so that the test set can include the new constructors?


Could

kmsquire:
Hi Stephen, 

Since this is already a pull request, the most standard way to deal with this is to merge it as is, and (if necessary) for you to rebase your PR on it (although this PR is so minimal that it probably isn't necessary).  

It would be better not to conflate it with your current PR, which is already long enough.

It could use some tests (or updates to the current tests) though (@Keno?). 

As of Julia v0.4, the standard `Dict` notation in `Base` is

```julia
julia> Dict{String, Int}()
Dict{AbstractString,Int64} with 0 entries
```

The previous notation is deprecated.  Presumably, this PR allows a similar notation for `SortedDicts` (I haven't tested it.)

kmsquire:
Superceded by #113.

Issue #97, opened 2015-06-02
Method ambiguity error in 0.4-dev


yuyichao:
Is this sth that should be fixed (or can be reasonably fixed) in the package or should this wait for better solution for method ambiguity landed in the language?

```julia
julia> using DataStructures
Warning: New definition 
    serialize(Any, DataStructures.HashDict) at /usr/share/julia/site/v0.4/DataStructures/src/hashdict.jl:91
is ambiguous with: 
    serialize(Base.SerializationState, Any) at serialize.jl:411.
To fix, define 
    serialize(Base.SerializationState, DataStructures.HashDict)
before the new definition.
Warning: New definition 
    serialize(Any, DataStructures.HashDict) at /usr/share/julia/site/v0.4/DataStructures/src/hashdict.jl:91
is ambiguous with: 
    serialize(Base.IO, Any) at serialize.jl:433.
To fix, define 
    serialize(Base.IO, DataStructures.HashDict)
before the new definition.
Warning: New definition 
    deserialize(Any, Type{DataStructures.HashDict{#K<:Any, #V<:Any, #O<:Any}}) at /usr/share/julia/site/v0.4/DataStructures/src/hashdict.jl:100
is ambiguous with: 
    deserialize(Base.SerializationState, DataType) at serialize.jl:666.
To fix, define 
    deserialize(Base.SerializationState, Type{DataStructures.HashDict{#K<:Any, #V<:Any, #O<:Any}})
before the new definition.
```

yuyichao:
Fixed by #102

PR #102, opened 2015-06-23
fix serialization ambiguity warnings

jakebolewski:
also fix the version for the serialization api change


kmsquire:
Thanks!  I had just made a similar change yesterday, but hadn't pushed it.  Yours is slightly more complete.  

Minor nit: since it's only written 3 times, I would prefer `SerializationState` over `SerState`--easier to understand at a glance.  

Otherwise, please merge!

kmsquire:
(Fixes #97)

PR #105, opened 2015-07-14
RFC: Fix constructor for OrderedDict

kmsquire:
* When constructing a dictionary literally using Pair syntax, construction would fail if
  any of the Pairs were not the same type as the dictionary (including
  `OrderedDict{Any,Any}`.  Loosened this restriction.
* Also allow constructing `OrderedDict{K,Any}` when the keys are the same,
  but the values have different types
* Note that constructing and `OrderedDict{Any,V}}` using Pairs is not
  currently supported

~~(Still need to add a test for the second change)~~

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -48,6 +48,7 @@ module DataStructures
         using Docile
     end
 
+    include("ds_compat.jl")
     include("delegate.jl")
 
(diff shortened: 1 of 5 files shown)


kmsquire:
Rather out of date, and hopefully obsoleted by #170.

Issue #103, opened 2015-06-26
keys don't seem to be properly typed


atteson:
When I run "typeof(collect(keys(SortedDict( Dict( ["a","b","c"], [1,2,3] ) ))))", I get "Array{Any,1}" whereas I would expect to get "Array{ASCIIString,1}" which is what "typeof(collect(keys(SortedDict( Dict( ["a","b","c"], [1,2,3] ) ))))" gives me.

I'm using Julia version 0.3.8 and seem to be on commit 79e517 of the DataStructures package.

kmsquire:
Cc: @StephenVavasis

StephenVavasis:
Thanks for pointing out this omission.  I have fixed this in the latest commit in the SortedContainers branch, which will, in the near future, become the master branch.  You can find this commit in my personal repository.  In the mean time, you can also work around this problem by using the two-argument form of the 'collect' function: collect(T, itr)

kmsquire:
@StephenVavasis, thanks for fixing this.  

Generally, the preferable action would be to submit a separate, small pull request containing just this fix (against master).  Is that something you would be willing to do?  

You would then have to rebase your SortedContainers branch agains master (which you'll have to do before it is merged anyway.)

StephenVavasis:
Kevin,

I just submitted a pull request that fixes just this one issue

StephenVavasis:
This patch that I just submitted (#106) address the issue that keys and values objects for SortedDict ought to support other functions besides just start/done/next in order to be compatible with Dict.  In particular, the patch addresses the problem that keys and values objects previously didn't support ``collect`` correctly.   

I'm studying dict.jl, and I see that there is at least one other function that ``keys`` is supposed to support but that it currently does not.

```julia
    in(k, keys(s))
```
where s is a SortedDict, will be slow because this operation will be done in the naive way (iterate over all keys of s, see if k is one of them).  The haskey() function is ok -- it uses efficient key lookup.

Fixing this is more complicated, because it is not clear what in(k, keys(a:b)) ought to do, where a, b are tokens.  (In the SortedContainers branch, this would b in(k, keys(inclusive(s,a,b))), where a, b are semitokens.)

There are several possibilities:

(1) Leave the code alone, but in the documentation, advise against using in(k,keys(..)).  This would contradict the Julia manual, which says that in(k,keys(..)) is the same as haskey(..).

[...]

PR #106, opened 2015-07-18
Include missing eltype functions so that collect works properly

StephenVavasis:
Address issue #103 by including many new eltype declarations.  Also, fixed the existing eltype declaration to be compatible with 0.4.

### DIFF
--- a/src/sortedDict.jl
+++ b/src/sortedDict.jl
@@ -231,6 +231,8 @@ immutable ExcludeLast{K, D, Ord <: Ordering}
     pastlast::Int
 end
 
+eltype(e::ExcludeLast) = eltype(e.m)
+
 
 ## This holds an object describing an include-last
 ## (i.e., colon operator) iteration.
@@ -240,6 +242,8 @@ immutable IncludeLast{K, D, Ord <: Ordering}
     first::Int
     last::Int
 end
+
+eltype(e::IncludeLast) = eltype(e.m)
 
 
 ## The basic iterations are either over the whole sorted dict, an
@@ -256,21 +260,54 @@ immutable SDKeyIteration{T <: SDIterableTypesBase}
     base::T
 end
 
+eltype{K,D,Ord <: Ordering}(s::SDKeyIteration{SortedDict{K,D,Ord}}) = K
+eltype{K,D,Ord <: Ordering}(s::SDKeyIteration{ExcludeLast{K,D,Ord}}) = K
+eltype{K,D,Ord <: Ordering}(s::SDKeyIteration{IncludeLast{K,D,Ord}}) = K
+
 immutable SDValIteration{T <: SDIterableTypesBase}
     base::T
 end
+
+eltype{K,D,Ord <: Ordering}(s::SDValIteration{SortedDict{K,D,Ord}}) = D
+eltype{K,D,Ord <: Ordering}(s::SDValIteration{ExcludeLast{K,D,Ord}}) = D
+eltype{K,D,Ord <: Ordering}(s::SDValIteration{IncludeLast{K,D,Ord}}) = D
 
 immutable SDTokenIteration{T <: SDIterableTypesBase}
     base::T
 end
 
+eltype{K,D,Ord <: Ordering}(s::SDTokenIteration{SortedDict{K,D,Ord}}) = 
+        @compat Tuple{SDToken{K,D,Ord}, Tuple{K, D}}
+eltype{K,D,Ord <: Ordering}(s::SDTokenIteration{ExcludeLast{K,D,Ord}}) = 
+        @compat Tuple{SDToken{K,D,Ord}, Tuple{K, D}}
+eltype{K,D,Ord <: Ordering}(s::SDTokenIteration{IncludeLast{K,D,Ord}}) = 
+        @compat Tuple{SDToken{K,D,Ord}, Tuple{K, D}}
+
+
 immutable SDTokenKeyIteration{T <: SDIterableTypesBase}
     base::T
 end
 
+eltype{K,D,Ord <: Ordering}(s::SDTokenKeyIteration{SortedDict{K,D,Ord}}) = 
+        @compat Tuple{SDToken{K,D,Ord}, K}
+eltype{K,D,Ord <: Ordering}(s::SDTokenKeyIteration{ExcludeLast{K,D,Ord}}) = 
+        @compat Tuple{SDToken{K,D,Ord}, K}
+eltype{K,D,Ord <: Ordering}(s::SDTokenKeyIteration{IncludeLast{K,D,Ord}}) = 
+        @compat Tuple{SDToken{K,D,Ord}, K}
+
+
 immutable SDTokenValIteration{T <: SDIterableTypesBase}
     base::T
 end
+
+eltype{K,D,Ord <: Ordering}(s::SDTokenValIteration{SortedDict{K,D,Ord}}) = 
+        @compat Tuple{SDToken{K,D,Ord}, D}
+eltype{K,D,Ord <: Ordering}(s::SDTokenValIteration{ExcludeLast{K,D,Ord}}) = 
+        @compat Tuple{SDToken{K,D,Ord}, D}
+eltype{K,D,Ord <: Ordering}(s::SDTokenValIteration{IncludeLast{K,D,Ord}}) = 
+        @compat Tuple{SDToken{K,D,Ord}, D}
+
+
 
 typealias SDCompoundIterable Union(SDKeyIteration,
                                    SDValIteration, SDTokenIteration,
@@ -399,7 +436,10 @@ function in{K,D,Ord <: Ordering}(pr::(@compat Tuple{Any,Any}), m::SortedDict{K,D
     return exactfound && isequal(m.bt.data[i].d,convert(D,pr[2]))
 end
 
-eltype{K,D,Ord <: Ordering}(m::SortedDict{K,D,Ord}) = (K,D)
+eltype{K,D,Ord <: Ordering}(m::SortedDict{K,D,Ord}) = @compat Tuple{K,D}
+eltype{K,D,Ord <: Ordering}(::Type{SortedDict{K,D,Ord}}) = @compat Tuple{K,D}
+keytype{K,D,Ord <: Ordering}(::SortedDict{K,D,Ord}) = K
+datatype{K,D,Ord <: Ordering}(::SortedDict{K,D,Ord}) = D
 
 orderobject(m::SortedDict) = m.bt.ord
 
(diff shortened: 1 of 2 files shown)


src/sortedDict.jl
```diff
@@ -256,22 +260,55 @@ immutable SDKeyIteration{T <: SDIterableTypesBase}
     base::T
 end
 
+eltype{K,D,Ord <: Ordering}(s::SDKeyIteration{SortedDict{K,D,Ord}}) = K
+eltype{K,D,Ord <: Ordering}(s::SDKeyIteration{ExcludeLast{K,D,Ord}}) = K
+eltype{K,D,Ord <: Ordering}(s::SDKeyIteration{IncludeLast{K,D,Ord}}) = K
```
kmsquire: These can be simplified to
```julia
eltype{K}(s::SDKeyIteration{SortedDict{K}}) = K
eltype{K}(s::SDKeyIteration{ExcludeLast{K}}) = K
eltype{K}(s::SDKeyIteration{IncludeLast{K}}) = K
```

Actually, though, did you check that these definitions did the right thing?  In the past, I believe the parser ignored the inner set of parenthesis, although that might not be the case any more.

src/sortedDict.jl
```diff
@@ -256,22 +260,55 @@ immutable SDKeyIteration{T <: SDIterableTypesBase}
...
+
 immutable SDValIteration{T <: SDIterableTypesBase}
     base::T
 end
 
+eltype{K,D,Ord <: Ordering}(s::SDValIteration{SortedDict{K,D,Ord}}) = D
+eltype{K,D,Ord <: Ordering}(s::SDValIteration{ExcludeLast{K,D,Ord}}) = D
+eltype{K,D,Ord <: Ordering}(s::SDValIteration{IncludeLast{K,D,Ord}}) = D
```
kmsquire: Same notes as above; here, you would need both `K` and `D`, but not `Ord`, since without `K`, there's no way to specify the second type parameter.

src/sortedDict.jl
```diff
@@ -399,7 +436,10 @@ function in{K,D,Ord <: Ordering}(pr::(@compat Tuple{Any,Any}), m::SortedDict{K,D
     return exactfound && isequal(m.bt.data[i].d,convert(D,pr[2]))
 end
 
-eltype{K,D,Ord <: Ordering}(m::SortedDict{K,D,Ord}) = (K,D)
+eltype{K,D,Ord <: Ordering}(m::SortedDict{K,D,Ord}) = @compat Tuple{K,D}
+eltype{K,D,Ord <: Ordering}(::Type{SortedDict{K,D,Ord}}) = @compat Tuple{K,D}
+keytype{K,D,Ord <: Ordering}(::SortedDict{K,D,Ord}) = K
+datatype{K,D,Ord <: Ordering}(::SortedDict{K,D,Ord}) = D
```
kmsquire: As above, you can remove at least `Ord` from these, which is slightly less noisy.

src/sortedDict.jl
yuyichao: I think this (and most of the ones below) cannot be simplified for the same reason `Array{Complex}` won't match `Array{Complex128}`

src/sortedDict.jl
yuyichao: To preserve the meaning, only these (except the `::Type` one) can be simplified.

src/sortedDict.jl
```diff
@@ -241,6 +243,8 @@ immutable IncludeLast{K, D, Ord <: Ordering}
     last::Int
 end
 
+eltype{K,D,Ord<:Ordering}(e::IncludeLast{K,D,Ord}) = eltype(e.m)
+
```
yuyichao: On the other hand, this one (and the one above) doesn't need any type parameters because they are not used.

The constraint on the `Ord` parameter can also be removed for the ones below.

src/sortedDict.jl
kmsquire: While that's true, unspecified type parameters are actually equivalent to `ANY`, in which case, `Array{Complex}` is covariant to `Array{ANY}`:
```julia
julia> Array{Complex} <: Array{Any}
false

julia> Array{Complex} <: Array{ANY}
true
```

StephenVavasis:
Removing the constraint on Ord in the eltype declarations did not seem to work.  Here is an example of the error message.  This is a 6-day old master.

ERROR: LoadError: LoadError: LoadError: TypeError: SDKeyIteration: in T, expecte
d T<:Union{DataStructures.SortedDict{K,D,Ord<:Base.Order.Ordering},DataStructure
s.IncludeLast{K,D,Ord<:Base.Order.Ordering},DataStructures.ExcludeLast{K,D,Ord<:
Base.Order.Ordering}}, got Type{DataStructures.SortedDict{K,D,Ord}}
 in include at boot.jl:254
 in include_from_node1 at loading.jl:133
 in include at boot.jl:254
 in include_from_node1 at loading.jl:133
 in include at boot.jl:254
 in include_from_node1 at loading.jl:133
while loading c:/Users/vavasis/Documents/GitHub/DataStructures.jl/src/sortedDict
.jl, in expression starting on line 263
while loading c:/Users/vavasis/Documents/GitHub/DataStructures.jl/src/DataStruct
ures.jl, in expression starting on line 75
while loading C:\Users\vavasis\Documents\Projects\qmg21\Julia\sctest1_oldDS.jl,
in expression starting on line 6

julia>

src/sortedDict.jl
yuyichao: My point was that `Array{Complex}` is a subtype of with `Array` but `Array{Complex{Float64}}` is not a subtype of `Array{Complex}`.

```julia
julia> type A{T1, T2}
       end

julia> type B{T}
       end

julia> f{T}(::B{A{T}}) = T
f (generic function with 1 method)

julia> f(B{A{Int, Int}}())
ERROR: MethodError: `f` has no method matching f(::B{A{Int64,Int64}})

julia> f(B{A{Int}}())
Int64
```

yuyichao:
It shouldn't. See my comment above https://github.com/JuliaLang/DataStructures.jl/pull/106#discussion_r34954535

src/sortedDict.jl
kmsquire: As @yuyichao pointed out above, I was wrong here: you do need the parameters (but you can remove `<: Ordering`).

kmsquire:
Yep, @yuyichao is right.  Sorry for the misdirection!  (If you only have one level of braces, you can remove the type parameters in the way that I mentioned, but not here.)

src/sortedDict.jl
kmsquire: Okay, got it--thanks for the explanation.  (As you pointed out, the `Ord` constraint can be removed.)

StephenVavasis:
Kevin and Yichao,

I pushed a new commit on this branch fix_eltype1 that removes the needless type parameters in two statements and tests the new eltype definitions in test_sorteddict.  With regard to all the other type parameters and similar stylistic issues, I'd like to suggest that we leave that topic aside because this PR is supposed to be a stopgap measure anyway until we can finish the SortedContainers branch.  I used a somewhat different style in that branch -- I defined small functions to extract type parameters.

I'd also appreciate any comments you have regarding my most recent posting on issue #103 -- I am unclear on the best way to deal with in(k, keys(...)) for SortedDict and SortedMultiDict.

Thanks,
Steve

kmsquire:
That seems fine.  Can you squash the commits here?  After that I (or someone else) will merge.

StephenVavasis:
Sorry, I guess I'm never going to master git... I squashed the last two commits with the command git rebase -i HEAD~2, which seemed to work (I accepted the default commit message), but then my 'git push' command didn't do anything ("Everything up to date") so I don't think I properly squashed the commits.

kmsquire:
Hi Stephen, if you do a `git log`, do you see both commits?  If not, then you properly squashed them.  

After squashing, when you push, you need to force push (`git push -f ...`) for the squashed change to appear on github.

yuyichao:
> Sorry, I guess I'm never going to master git... I squashed the last two commits with the command git rebase -i HEAD~2, which seemed to work (I accepted the default commit message), but then my 'git push' command didn't do anything ("Everything up to date") so I don't think I properly squashed the commits.

Feels like you didn't actually change anything.

When you are doing `git rebase -i`, the first editor window that pops up is for letting you automatically select the action you want to take on each commit (and **NOT** the commit message). (The `-i` is for interactive).

Each non-comment line of the file should start with `pick` followed by the SHA of the commit and the commit message. If you change the `pick` in front of the latest commit, to `s` (or `squash`), it should try to squash the two commits together.

If you have done that correctly and save&exit your editor, git should pop up another window for you to select the commit message. This is when you should edit the commit message to whatever you want and save&exit the editor. If everything goes right, you should have successfully squashed the two commits.

As @kmsquire said, when you want to push the result of a squash, be sure to add `-f` since you are overwriting the history (instead of simply appending to it)

StephenVavasis:
Yichao, 

Sorry, this did not work either.  I typed git rebase -i HEAD~2 again, and as you said, I got an editor window containing two lines beginning with the word 'pick'.  I changed both instances of 'pick' to 'squash' and closed the editor window.  Then I received an error message:

   ***
Cannot 'squash' without a previous commit.
  ***

Now I appear to be stuck: 

'git status' tells me that a rebase is in progress:

 ***
rebase in progress; onto 27f9c92
You are currently editing a commit while rebasing branch 'fix_eltype1' on '27f9c
92'.
  (use "git commit --amend" to amend the current commit)
  (use "git rebase --continue" once you are satisfied with your changes)

nothing to commit, working directory clean
  ***

 but git rebase --continue gives an error message:  

  ***
cat: c:/Users/vavasis/Documents/GitHub/DataStructures.jl/.git/rebase-merge/stopp
ed-sha: No such file or directory
Cannot 'squash' without a previous commit
  ***

'git log' does not show either of my recent commits on this branch, and the source-code files themselves have been reverted to their 'master' versions (i.e., prior to my two recent rounds of editing).

Some day I hope to join the ranks of millions of happy 'git' users around the world, but not today!

yuyichao:
> Sorry, this did not work either. I typed git rebase -i HEAD~2 again, and as you said, I got an editor window containing two lines beginning with the word 'pick'. I changed both instances of 'pick' to 'squash' and closed the editor window. Then I received an error message:

No you need to only change the one for the **last** commit

yuyichao:
So you should not do a `git rebase --abort` and do the `git rebase -i` again.

StephenVavasis:
Yichao,

Thanks for the help!  I think it worked correctly this time.

-- Steve

kmsquire:
Looks like this was handled by #113 (or maybe a later commit?).

PR #109, opened 2015-07-28
added CircularBuffer with tests and docs

tbreloff:
This is a fairly generic structure that I have in my QuickStructs.jl package, but probably belongs here.  Tests passed for me on 0.4 and 0.3.10.  Comments welcome.

### DIFF
--- a/src/circularbuffer.jl
+++ b/src/circularbuffer.jl
@@ -1,0 +1,104 @@
+
+# this represents a circular buffer, where new items are pushed to the back of the list, 
+# overwriting values in a circular fashion
+# iterating occurs from the "start" until the "end", and will return "length(buffer)" nodes
+
+type CircularBuffer{T} <: AbstractVector{T}
+  capacity::Int
+  startIdx::Int
+  buffer::Vector{T}
+
+  CircularBuffer{T}(::Type{T}, capacity::Int) = new(capacity, 1, T[])
+end
+
+CircularBuffer{T}(::Type{T}, capacity::Int) = CircularBuffer{T}(T, capacity)
+
+function CircularBuffer{T}(::Type{T}, capacity::Int, fillval::T)
+  cb = CircularBuffer(T, capacity)
+  for i in 1:capacity
+    push!(cb, copy(fillval))
+  end
+  cb
+end
+
+function CircularBuffer{T}(::Type{T}, capacity::Int, v::Vector{T})
+  cb = CircularBuffer(T, capacity)
+  for x in v
+    push!(cb, copy(x))
+  end
+  cb
+end
+
+
+
+function Base.print{T}(io::IO, cb::CircularBuffer{T})
+  print(io, "{")
+  print(io, join(collect(cb), ", "))
+  print(io, "}")
+end
+
+function Base.show{T}(io::IO, cb::CircularBuffer{T})
+  print(io, "CircularBuffer(cap=$(cb.capacity) start=$(cb.startIdx) n=$(length(cb)))")
+  print(io, cb)
+end
+
+
+# cap = 8
+# n = 6
+# start = 6
+
+# i = 1
+# x = 6
+
+# i = 4
+# x = 1
+
+# 4 5 6 _ _ 1 2 3
+
+
+
+
+function bufferIndex(cb::CircularBuffer, i::Int)
+  if i < 1 || i > length(cb)
+    error("CircularBuffer out of range. cb=$cb i=$i")
+  end
+  (i + cb.startIdx - 2) % cb.capacity + 1
+end
+
+Base.getindex(cb::CircularBuffer, i::Int) = cb.buffer[bufferIndex(cb, i)]
+function Base.setindex!{T}(cb::CircularBuffer, data::T, i::Int)
+  cb.buffer[bufferIndex(cb, i)] = data
+  nothing
+end
+
+
+
+function Base.push!{T}(cb::CircularBuffer{T}, data::T)
+
+  if length(cb) == cb.capacity
+    # if the buffer is full, increment the startIdx, and add to the end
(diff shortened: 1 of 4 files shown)


tbreloff:
Seems the build is now failing because of the recent change from Tuple --> Pair for associative containers...  failures in test_ordereddict.jl.

tbreloff:
A recent thread on julia-dev reminded me that I submitted this months ago without response.  Should I bother making this compatible with current 0.4, or should I just close the PR and forget about it?

src/circularbuffer.jl
```diff
@@ -0,0 +1,104 @@
...
+# n = 6
+# start = 6
+
+# i = 1
+# x = 6
+
+# i = 4
+# x = 1
+
+# 4 5 6 _ _ 1 2 3
```
danielarndt: Were these comments intentionally left?

tbreloff: Partially to show how it works... they could certainly be removed though.

src/circularbuffer.jl
```diff
@@ -0,0 +1,104 @@
+
+# this represents a circular buffer, where new items are pushed to the back of the list, 
+# overwriting values in a circular fashion
+# iterating occurs from the "start" until the "end", and will return "length(buffer)" nodes
+
+type CircularBuffer{T} <: AbstractVector{T}
+  capacity::Int
```
simonster: nit: use 4 spaces for indentation

src/circularbuffer.jl
```diff
@@ -0,0 +1,104 @@
...
+  startIdx::Int
+  buffer::Vector{T}
+
+  CircularBuffer{T}(::Type{T}, capacity::Int) = new(capacity, 1, T[])
```
simonster: There is no need to write the type twice, since this is an inner constructor for a parametrized type. This could just be `CircularBuffer(capacity::Int) = new(capacity, 1, T[])`.

src/circularbuffer.jl
```diff
@@ -0,0 +1,104 @@
...
+
+CircularBuffer{T}(::Type{T}, capacity::Int) = CircularBuffer{T}(T, capacity)
+
+function CircularBuffer{T}(::Type{T}, capacity::Int, fillval::T)
+  cb = CircularBuffer(T, capacity)
+  for i in 1:capacity
+    push!(cb, copy(fillval))
+  end
+  cb
+end
```
simonster: Not sure this constructor is necessary. `fill!(CircularBuffer(T, n), fillval)` ought to work and is more idiomatic. If performance is an issue you could create a specialized method to do this. (Not sure it should be copying `fillval` though.)

src/circularbuffer.jl
```diff
@@ -0,0 +1,104 @@
...
+
+function CircularBuffer{T}(::Type{T}, capacity::Int, v::Vector{T})
+  cb = CircularBuffer(T, capacity)
+  for x in v
+    push!(cb, copy(x))
+  end
+  cb
+end
```
simonster: Similarly, this is `copy!(CircularBuffer(T, n), v)`. This should almost certainly not be copying `x`.

src/circularbuffer.jl
```diff
@@ -0,0 +1,104 @@
...
+# 4 5 6 _ _ 1 2 3
+
+
+
+
+function bufferIndex(cb::CircularBuffer, i::Int)
+  if i < 1 || i > length(cb)
+    error("CircularBuffer out of range. cb=$cb i=$i")
+  end
+  (i + cb.startIdx - 2) % cb.capacity + 1
```
simonster: Possibly faster to branch here or use `ifelse` than to use `%`, since `%` is expensive

src/circularbuffer.jl
```diff
@@ -0,0 +1,104 @@
...
+end
+
+Base.getindex(cb::CircularBuffer, i::Int) = cb.buffer[bufferIndex(cb, i)]
+function Base.setindex!{T}(cb::CircularBuffer, data::T, i::Int)
+  cb.buffer[bufferIndex(cb, i)] = data
+  nothing
```
simonster: Usually `setindex!` returns the AbstractArray (i.e. `cb`)

src/circularbuffer.jl
```diff
@@ -0,0 +1,104 @@
...
+    push!(cb.buffer, data)
+
+  end
+end
+
+function Base.append!{T}(cb::CircularBuffer{T}, datavec::AbstractVector{T})
+  for d in datavec
+    push!(cb, d)
+  end
+end
```
simonster: This could possibly be more efficient if specialized instead of using `push!`

tbreloff:
Thanks for all the comments.  I'm sure I wouldn't do this the same way today.  Does this mean someone would actually like this to be merged?  If so I can fix it up and re-submit, but I don't want to waste my time.

simonster:
I've had a use for this before, and I think it would be a good thing to have here.

src/circularbuffer.jl
tbreloff: I think I `copy` because this may be holding "reference types" like Array, and I don't want a buffer of the same object.  Is there a better way to enable this?  Or just require the user to add this functionality themselves?

simonster: Generally I think the user needs to do this themselves. Exposing an API that's consistent with the API for arrays etc. seems more important than implementing this functionality here, and there are legitimate reasons you might not want a copy.

src/circularbuffer.jl
tbreloff: afaik, `copy!` expects that `length(v) == n`, which is not necessarily true.  I might initialize with a vector larger or smaller than the capacity.

simonster: Yes, you're right. But couldn't you use `append!(CircularBuffer(T, n), v)`?

tbreloff:
Closing in favor of #144

PR #112, opened 2015-08-04
fix Pair usage for ordereddict

rened:
Make `collect` return an array of `Pair`s, fixes unit tests.
~~Fixed `IntSet` deprectation warning for `status()` of `SortedDict`.~~(not necessary after rebase)

PR #113, opened 2015-08-09
For sorted containers: add more constructors; add new behavior for 'i…

StephenVavasis:
…n' function

Added more constructors for SortedDict and SortedMultiDict to mimic bulit-in Dict constructors; added code for fast lookup of in(k, keys(sc)) where sc is is a SortedDict or SortedMultiDict.

Did not address issue #112 because it appears that the changed to IntSet in 0.4 is going to be reverted soon.


StephenVavasis:
Also: added 'similar' function to mimic built-in 'similar' for Dict.

PR #125, opened 2015-09-02
More deque tests

kshyatt:
Testing the IO methods and `empty!`.

### DIFF
--- a/test/test_deque.jl
+++ b/test/test_deque.jl
@@ -11,6 +11,7 @@ q = Deque{Int}()
 @test q.blksize == DataStructures.DEFAULT_DEQUEUE_BLOCKSIZE
 @test_throws ErrorException front(q)
 @test_throws ErrorException back(q)
+@test length(sprint(dump,q)) >= 0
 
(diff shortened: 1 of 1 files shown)


StephenVavasis:
Just wondering about checking IO methods with this kind of code:

```julia
@test sprint(dump,q) == "Deque (length = 0, nblocks = 1)\nblock 1 [1:0] ==> \n"
```
It seems like this would be quite brittle since details of how ``show`` works for built-in types could change from one Julia build to the next.  On the other hand, I don't have a proposal for a better way to test a ``show`` method.

kshyatt:
I agree that it's brittle, and I also don't have any better suggestions. Does this (now that I've updated it to not fail!) look ok to you?

StephenVavasis:
Maybe just ``@test length(sprint(show,a)) > 0``?  At least this runs the code and confirms that it returns something?

kshyatt:
@StephenVavasis I've updated the PR per your comments. Is this ok?

StephenVavasis:
OK with me.  Thanks for taking care of this.

PR #136, opened 2015-10-13
Fix #135, define ==(x::Nil, y::Nil) and ==(x::Cons, y::Cons)

### LINKED ISSUE #135: list(x)==list(x) returns false
It looks like list `isequal` only yields the correct answer for empty lists.

```
julia> nil() == nil()
true

julia> list(2,3,4) == list(2,3,4)
false

julia> versioninfo()
Julia Version 0.5.0-dev+238
Commit 5bebf41 (2015-09-17 07:27 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas
  LIBM: libopenlibm
  LLVM: libLLVM-3.3
```

The reason it that comparison of values `==` falls back to comparison of hashes `===`. Yet list hashes are guaranteed to be the same only when the list is `nil()`.

```
julia> @which list(2,3,4) == list(2,3,4)
==(x, y) at operators.jl:11
```

Is this intended ? If not I can submit a PR to fix this.

lostanlen:
Bugfix to issue #135 : `list(x)==list(x)` returns false.
With this implementation, `nil(T) == nil(S)` is `true` for all types `T` ans `S`. This follows the same equality rule as in arrays : `T[] == S[]`.
When `x::Cons` and `y::Cons`, we compare `x.head` and `y.head` and call `==` recursively on the tails of `x` and `y`.
All other equality rules for lists can be implemented implicitly as a generic fallback to `Base.===`.
Tests included.

### DIFF
--- a/src/list.jl
+++ b/src/list.jl
@@ -16,6 +16,9 @@ nil() = nil(Any)
 head(x::Cons) = x.head
 tail(x::Cons) = x.tail
 
+Base.(:(==))(x::Nil, y::Nil) = true
+Base.(:(==))(x::Cons, y::Cons) = (x.head == y.head) && (x.tail == y.tail)
+
 function show{T}(io::IO, l::LinkedList{T})
     if isa(l,Nil)
         if is(T,Any)
--- a/test/test_list.jl
+++ b/test/test_list.jl
@@ -3,10 +3,14 @@ using Base.Test
 
 l0 = nil(Char)
 @test length(l0) == 0
+@test l0 == nil(Char)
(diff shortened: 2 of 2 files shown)


kmsquire:
LGTM.

PR #137, opened 2015-10-18
Fix depwarn about extending == without explicit import

yuyichao:


yuyichao:
Will merge tomorrow if there's no objection.

PR #138, opened 2015-10-26
Serialize & deserialize hashdicts in Julia 0.4

danielarndt:
The signature of `write` changed a bit in Julia 0.4, and using `SerState` (ie. `Base.Serializer.SerializationState`) directly doesn't work. `s.io` works in 0.4 but wouldn't be backwards compatible with Julia 0.3.

Using the `serialize` method provided in the Julia distribution will allow the base library to make the appropriate call.


danielarndt:
@kmsquire thoughts?

kmsquire:
LGTM.

Sorry, my Julia time has been almost non-existent recently.

Issue #139, opened 2015-10-29
Strange values in counter


DSLituiev:
I would like to get a sum of counts in the counter, analogously to pythonic `c= Counter([...]);sum(c)` and I am getting extraterrestrial values out of it counter:
```
using Distributions
z = Exponential(.5)
using(DataStructures)
samples = int(round(rand(z, int(100))));
a = counter(samples)    #
a.map
```
which outputs
```
Dict{Int64,Int64} with 3 entries:
  0 => 65
  2 => 5
  1 => 30
```

Now, however, `a.map.keys` returns something strange (it might be ok first initialization, but it gets  this crazy with re-assignment):
```
16-element Array{Int64,1}:
               0
 140550171844608
 140550171844624
 140550059842144
 140550121312448
               2
              17
          559105
              17
              17
      4360087936
      4524992568
      4524992608
      4359842408
      4360086048
               1
```
and in `a.map.vals`:
```
16-element Array{Int64,1}:
              65
      4360086448
      4360086488
      4360086088
      4360086528
               5
 140550121312640
 140550121312496
              18
          559105
              18
              18
 140550075695680
 140550075695760
 140550075695840
              30
```

DSLituiev:
this beheaviour is observed in versions 0.3.11 and 0.4.0

danielarndt:
By calling `a.map.keys` (or `a.map.vals`), you are accessing the internal representation of the `Dict`. There is no scoping in Julia, but typically something like this would be hidden from the user (In python, I believe this would typically be stored in `a.map._keys`. In other words, you're looking at how a hashmap looks on the inside.

You'll notice you can do the same with a regular Dict (put some values in it, then access `.keys` and `.values` on it)

You're probably looking for `keys(a.map)` or `values(a.map)`.

Hope that clears things up.

yuyichao:
As @danielarndt said, don't access non-documented internal members. There are on-going discussions about what should be part of a public api but these are clearly nothing you should access directly. Please use the wrapper functions that does the right things instead (`keys` and `values`).

DSLituiev:
is there a way to implement sum() method for counters like in python?
it is very useful i believe, as it allows to normalize the histograms counted with counters

yuyichao:
As @danielarndt already said, use `values(a.map)` instead of `a.map.vals`. `sum(values(a.map))` works just fine.

It does seem like counter doesn't implement `values` and you have to use the (documented) `map` field to access the values.

[...]

PR #140, opened 2015-10-31
implement values(::Accumulator)

yuyichao:
Ref #139


yuyichao:
Does anyone have any issue with this?

yuyichao:
Will merge tomorrow if no one objects.

danielarndt:
@yuyichao Seems to be pretty dead here. Do you know if this is still actively maintained?

yuyichao:
I don't think this has a maintainer (i.e. no one is specifically in charge of it) but it's certainly not dead.

Feel free to ping people on your PR (either just `ping` or use `git blame` to find the people that might be most qualified for reviewing).

PR #144, opened 2015-11-10
CircularBuffer

tbreloff:
This replaces #109.  Comments welcome.

### DIFF
--- a/src/circularbuffer.jl
+++ b/src/circularbuffer.jl
@@ -1,0 +1,61 @@
+
+"""
+New items are pushed to the back of the list, overwriting values in a circular fashion.
+"""
+type CircularBuffer{T} <: AbstractVector{T}
+    capacity::Int
+    first::Int
+    buffer::Vector{T}
+
+    CircularBuffer(capacity::Int) = new(capacity, 1, T[])
+end
+
+function _buffer_index(cb::CircularBuffer, i::Int)
+    n = length(cb)
+    if i < 1 || i > n
+        error("CircularBuffer out of range. cb=$cb i=$i")
+    end
+    idx = cb.first + i - 1
+    if idx > n
+        idx - n
+    else
+        idx
+    end
+end
+
+function Base.getindex(cb::CircularBuffer, i::Int)
+    cb.buffer[_buffer_index(cb, i)]
+end
+
+function Base.setindex!(cb::CircularBuffer, data, i::Int)
+    cb.buffer[_buffer_index(cb, i)] = data
+    cb
+end
+
+function Base.push!(cb::CircularBuffer, data)
+    # if full, increment and overwrite, otherwise push
+    if length(cb) == cb.capacity
+        cb.first = (cb.first == cb.capacity ? 1 : cb.first + 1)
+        cb[length(cb)] = data
+    else
+        push!(cb.buffer, data)
+    end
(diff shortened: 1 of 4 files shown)


src/circularbuffer.jl
```diff
@@ -0,0 +1,61 @@
...
+    first::Int
+    buffer::Vector{T}
+
+    CircularBuffer(capacity::Int) = new(capacity, 1, T[])
+end
+
+function _buffer_index(cb::CircularBuffer, i::Int)
+    n = length(cb)
+    if i < 1 || i > n
+        error("CircularBuffer out of range. cb=$cb i=$i")
```
simonster: This should probably be a BoundsError.

test/test_circularbuffer.jl
```diff
@@ -0,0 +1,34 @@
+using DataStructures
+using Base.Test
+
+
+cb = CircularBuffer{Int}(5)
+@test length(cb) == 0
+@test capacity(cb) == 5
+@test_throws ErrorException first(cb)
```
simonster: It looks like this is throwing an ArgumentError instead of an ErrorException on 0.4+. Maybe this has to be `@test_throws Union{ErrorException,ArgumentError} first(cb)`

test/test_circularbuffer.jl
```diff
@@ -0,0 +1,34 @@
...
+@test isempty(cb) == false
+@test isfull(cb) == true
+@test convert(Array, cb) == Int[4,5,6,7,8]
+@test cb[1] == 4
+@test cb[2] == 5
+@test cb[3] == 6
+@test cb[4] == 7
+@test cb[5] == 8
+@test_throws ErrorException cb[6]
+@test_throws BoundsError cb[3:6]
```
simonster: I don't think indexing with ranges/vectors will work on 0.3 unless you implement it yourself, which probably isn't worth it. So maybe encapsulate these in `if VERSION >= v"0.4"`. Or alternatively, I wonder whether people would be opposed to dropping 0.3 support now that 0.4 is out.

src/circularbuffer.jl
tbreloff: I agree, but what is the right way to support 0.3 with a BoundsError?  The signatures are different 0.3 vs 0.4.

simonster: The lowest common denominator is to `throw(BoundsError())`, which should work on 0.3 or 0.4. But maybe we can just drop 0.3 support here now that 0.4 is released? I just opened #145 to see how others feel...

simonster:
Sounds like we're okay with dropping 0.4 support, so if you don't want to bother with 0.3, you can remove the line from `.travis.yml` to test on 0.3, update `REQUIRE`, and make sure your tests pass, and then I'll merge.

StephenVavasis:
Could we hold off a bit on dropping 0.3 support?  The issue is that my section of the README.rst file (sorted containers) makes extensive reference to 0.3.  It could mislead users to leave all those references to 0.3 in the documentation if we stop supporting it.  So I need a couple of days to edit README.rst (unless someone else wants to do it).

danielarndt:
Just for book keeping, @StephenVavasis has made the README changes mentioned available at #146

hayd:
0.3 has been removed.

PR #147, opened 2015-11-25
Bottom-up heap construction

doppioandante:
Bubbling down backwards guarantees a a tighter bound (O(n)) when constructing the heap from a linear array.
I still have to look at the mutable heap but I suppose the same thing can be done there.


doppioandante:
Well I'm not really familiar with the mutable heap implementation, though the change seems to apply neatly even there, I propose to merge this first.

kmsquire:
Hi @doppioandante, sorry for the slow response here.  I wanted to give this a closer look before merging, but it's been hard making the time.  If you have time, can you 

1. point to a reference for the tighter bound, and/or
2. give an example of a situation where this implementation is better/faster?

PR #152, opened 2015-12-10
[RFC] deque(T) == deque(T) returns false

lostanlen:
After #135 (fixed by #136) about `LinkedList`s, I stumbled again on the equality problem again with `Deque`s.
```julia
a = deque(Int)
b = deque(Int)
a == b # returns false, should return true
push!(a, 2)
push!(b, 2)
a == b # returns false, should return true
```
The reason is that `==` falls back to `===` since it is not explicitly implemented (much like #135).
So this PR suggests a fix. 
Note that, with this implementation, empty deques are considered equal even if their eltypes are different. In other words: `deque(T) == deque(S)`. This is consistent with arrays, where `T[] == S[]`.
I have written two recursive implementations, one for each deque traversing direction: `pop!` and `shift!`.

Tests are not included. I will add some if you agree with the idea.

### DIFF
--- a/src/deque.jl
+++ b/src/deque.jl
@@ -90,6 +90,29 @@ function back(q::Deque)
 end
 
 
+# Equality
+
+function ==(x::Deque, y::Deque ; front::Bool = true)
+    xcopy = deepcopy(x)
+    ycopy = deepcopy(y)
+    front ? _equals_shift!(xcopy, ycopy) : _equals_pop!(xcopy, ycopy)
+end
+
+function _equals_pop!(x::Deque, y::Deque)
+    isempty(x) && return isempty(y)
+    isempty(y) && return false
+    isequal(pop!(x), pop!(y)) || return false
+    isequal_pop!(x, y)
+end
+
+function _equals_shift!(x::Deque, y::Deque)
+    isempty(x) && return isempty(y)
+    isempty(y) && return false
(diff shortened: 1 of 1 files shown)


yuyichao:
Is it necessary to always make a copy?

Also I think if you implement `==` you should also implement `hash` in a consistent manner.

lostanlen:
@yuyichao You're right, we could avoid making a copy e.g. by using iterators.
How about this solution ?
```julia
function ==(x::Deque, y::Deque)
    itx = start(x)
    ity = start(y)
    while ~done(x, itx) && ~done(y, ity)
        (valx, itx) = next(x, itx)
        (valy, ity) = next(y, ity)
        (valx == valy) || return false
    end
    return done(x, itx) && done(y, ity)
end
```

I must say I have never written a hash function.

yuyichao:
You could probably use `zip` instead of manually interating both of them

You can also check the hash function of other container types.

kmsquire:
@lostanlen, if you have time, please feel free to modify this to be the `zip` version without the hash function--we can add `hash` later.

hayd:
Should Deques be == if they have a different blksize ? Should they have a different hash? (I think so...)

lostanlen:
Mille mercis @hayd and reviewers :)

PR #158, opened 2016-01-08
fix bug in de/serialize for ≥ juila v0.4

StefanKarpinski:


StefanKarpinski:
Nevermind. Should have pulled master first.

kmsquire:
Is this fixed only on master and not fixed in the latest tagged release?  I don't think anything has changed since then.

StefanKarpinski:
No, it's fixed in 0.4.1

PR #161, opened 2016-01-11
RFC: Delete deprecation warnings

kmsquire:
* The latest of these was added 8 months ago, which is probably long
  enough for most packages to adapt
* The corresponding deprecations in Julia itself have been removed
* In the future, these should include the specific DataStructures/
  julia version when added and/or when to remove


hayd:
+1

PR #163, opened 2016-01-13
Add hash and == for Deque.

hayd:
fixes #152.

Also offers a hash.

Both of these are different if the blksize is different. Perhaps this is not desired?

### DIFF
--- a/src/deque.jl
+++ b/src/deque.jl
@@ -271,3 +271,20 @@ function shift!{T}(q::Deque{T})  # pop front
     q.len -= 1
     x
 end
+
+const _deque_hashseed = UInt === UInt64 ? 0x950aa17a3246be82 : 0x4f26f881
+function hash(x::Deque, h::UInt)
+    h += _deque_hashseed
+    for (i, x) in enumerate(x)
+        h += i * hash(x)
+    end
+    h
+end
+
+function ==(x::Deque, y::Deque)
+    length(x) != length(y) && return false
+    for (i, j) in zip(x, y)
+        i == j || return false
+    end
+    true
+end
--- a/test/test_deque.jl
+++ b/test/test_deque.jl
@@ -152,6 +152,23 @@ for k = 1 : m
     @test collect(q) == r
 end
 
+# hash and ==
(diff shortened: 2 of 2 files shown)


kmsquire:
@hayd, thanks!

I would suggest that the most desirable version of equality would be content equality and order, but not structural equality, so I would remove the dependence on `blksize`.

For example, Julia makes an effort to make `0` and `0.0` give the same hash, even though they have totally different types:

```julia
julia> hash(0)
0x77cfa1eef01bca90

julia> hash(0.0)
0x77cfa1eef01bca90
```

Otherwise, looks good!

hayd:
I've removed that part and pushed!

src/deque.jl
```diff
@@ -271,3 +271,20 @@ function shift!{T}(q::Deque{T})  # pop front
...
 end
+
+const _deque_hashseed = UInt === UInt64 ? 0x950aa17a3246be82 : 0x4f26f881
+function hash(x::Deque)
+    h = _deque_hashseed
+    for (i, x) in enumerate(x)
+        h += i * hash(x)
+    end
+    h
+end
```
kmsquire: Sorry, one more comment: this should be changed to the 2-argument version of `hash` (I just discovered this).

```julia
help?> hash
search: hash haskey skipchars blas_set_num_threads searchsortedlast with_bigfloat_precision

  hash(x[, h::UInt])

  Compute an integer hash code such that isequal(x,y) implies hash(x)==hash(y). The optional second argument h is a hash code to be mixed with the result.

  New types should implement the 2-argument form, typically by calling the 2-argument hash method recursively in order to mix hashes of the contents with each other (and with h). Typically, any type that implements hash should also
  implement its own == (hence isequal) to guarantee the property mentioned above.
```

See https://github.com/JuliaLang/julia/blob/master/base/abstractarray.jl#L1415 for a (slightly complicated) example.

hayd: Dang it, I knew that. My bad, will update.

hayd:
@kmsquire thanks for the review! Let's merge and perhaps use a cleverer hash in the future.

PR #164, opened 2016-01-14
RFC: change type to hide implementation

kmsquire:
This is a resubmission of #98, which I was too slow to merge before.  It's a minor change, and if there's reason not to do this, it can easily be reverted later.  But it's a small step in the right direction to cleaning up `Stack` and `Queue` (and `Deque`) types.

(Original PR from @andrewcooke; he still owns the commit.)

### DIFF
--- a/src/queue.jl
+++ b/src/queue.jl
@@ -1,7 +1,7 @@
 # FIFO queue
 
-type Queue{S}   # S is the type of internal deque
-    store::S
+type Queue{T}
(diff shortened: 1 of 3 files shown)


DanielArndt:
LGTM, I was always a little curious why the type was exposed like that.

hayd:
+1 a related change is adding eltype for many types, I have a branch but will rebase off this once merged.

PR #165, opened 2016-01-16
CircularBuffer

kmsquire:
Small patch to make tests in #144 pass.  Closes #144.

Cc: @tbreloff

### DIFF
--- a/src/circularbuffer.jl
+++ b/src/circularbuffer.jl
@@ -1,0 +1,61 @@
+
+"""
+New items are pushed to the back of the list, overwriting values in a circular fashion.
+"""
+type CircularBuffer{T} <: AbstractVector{T}
+    capacity::Int
+    first::Int
+    buffer::Vector{T}
+
+    CircularBuffer(capacity::Int) = new(capacity, 1, T[])
+end
+
+function _buffer_index(cb::CircularBuffer, i::Int)
+    n = length(cb)
+    if i < 1 || i > n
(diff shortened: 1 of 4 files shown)


src/circularbuffer.jl
```diff
@@ -0,0 +1,61 @@
...
+    first::Int
+    buffer::Vector{T}
+
+    CircularBuffer(capacity::Int) = new(capacity, 1, T[])
+end
+
+function _buffer_index(cb::CircularBuffer, i::Int)
+    n = length(cb)
+    if i < 1 || i > n
+        error("CircularBuffer out of range. cb=$cb i=$i")
```
simonster: This should still be a BoundsError.

kmsquire: Fixed, thanks!

kmsquire:
This has been around for a while, first as #109, then #144, so there's been quite a bit of time for commenting already.

Since I haven't commented myself before... `DataStructures` has kind of become a catch-all for various general data structures that people need/want, but the package itself doesn't have a roadmap or philosophy about what belongs here and what form the things that do go here should have.  

I'd like that to change, but with my uncertainty of my own availability, there's no guarantee that I will make this will happen in a reasonable time frame.  

So if anyone has a burning desire to gain fame and glory by whipping `DataStructures` into shape, please speak up... ;-)

kmsquire:
Oh, and I'll merge tomorrow if there are no further comments.

kmsquire:
Thanks Daniel!

PR #166, opened 2016-01-17
RFC: Use "throw" with a specific exception instead of "error()"

kmsquire:
Cleaning up `error` functions.

In cases where the correct exception seemed ambiguous (e.g., `pop!` on an empty `Deque`), I followed conventions in `Base`.

### DIFF
--- a/src/defaultdict.jl
+++ b/src/defaultdict.jl
@@ -19,7 +19,7 @@ immutable DefaultDictBase{K,V,F,D} <: Associative{K,V}
     d::D
 
     check_D(D,K,V) = (D <: Associative{K,V}) ||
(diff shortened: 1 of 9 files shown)


kmsquire:
Will merge tomorrow if there are no comments.

DanielArndt:
@kmsquire really quick rebase needed here (there's a small conflict with you adding a `Base.Test` import, and pr #172 removed the imports) which I can do if you wouldn't mind that I force push over your branch (Or if you have a few minutes quickly do it yourself).

kmsquire:
If you wouldn't mind doing that and merging, that would be great!

kmsquire:
Thanks!

PR #168, opened 2016-01-17
RFC: Add default 2-type argument constructors to sorted collections

### LINKED ISSUE #134: missing default constructor for SortedDict,SortedMultiDict, SortedSet
The SortedDict, SortedMultiDict, and SortedSet types are missing the default constructors, and I'm not sure how to write them.In other words, the following code gives an error about a missing constructor:

``` julia
   s = SortedDict{Int, Int}()
```

(There are other ways to construct an empty SortedDict, for instance `s=SortedDict(Dict{Int,Int}())`, but the above would be the most natural.) Any suggestions?  The relevant source code is in sortedDict.jl.  I am embarrassed that I still don't understand this!

Edit (3 hours later): For further clarification: I am able to create a sorteddict by specifying all three type parameters, e.g., 

```
   s = SortedDict{Int, Int, Base.ForwardOrdering}()
```

works, but ideally `SortedDict` will support the same empty-dictionary constructor as `Dict` (and hence choose the default Base.ForwardOrdering for the third type parameter if it is omitted).    But I have not been able to figure out how to do this.

kmsquire:
This allows one to write constructors without specifying an ordering, e.g.:

``` julia
SortedDict{Int, Int}()
SortedMultiDict{Int, Int}()
SortedSet{Int}()
```

Defaults to `Forward` ordering.  Fixes #134.

Cc: @StephenVavasis

### DIFF
--- a/doc/source/sorted_containers.rst
+++ b/doc/source/sorted_containers.rst
@@ -9,7 +9,7 @@ SortedDict, SortedMultiDict and SortedSet.
 *SortedDict* is similar to the built-in Julia type ``Dict``
 with the additional feature that the keys are stored in
 sorted order and can be efficiently iterated in this order.
-SortedDict is a subtype of Associative.  It is slower than ``Dict``
+SortedDict is a subtype of Associative.  It is generally slower than ``Dict``
 because looking up a key requires an O(log *n*) tree search rather than
 an expected O(1) hash-table lookup time as with Dict.
 SortedDict is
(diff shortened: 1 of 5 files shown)


kmsquire:
I've added a few more constructors for `SortedDict`s--basically, I tried things I thought were reasonable, and whenever it broke, I added a constructor that made it work again.

These new constructors still need tests, and similar constructors should probably be added to `SortedSet` and `SortedMultiDict`.

kmsquire:
@garborg @StephenVavasis, I had been working on a separate branch, and hadn't finished pushed back here.

This branch now contains my latest changes.  `SortedDict` constructors should be in good shape.  The new constructors are based on those for `Base.Dict`, and should be pretty complete.

`SortedMultiDict` and `SortedSet` have a few additional constructors, but are not quite as filled out.  I also probably won't have time to work on those again anytime soon.  

As long as CI passes, I think it can be merged as is, as it is an improvement over the current state.  The `SortedMultiDict` and `SortedSet` constructors can be updated in a later PR.

@StephenVavasis, can you please review, and suggest any changes, or merge if it seems okay?

kmsquire:
The changes here are awaiting CI (which already passed before I squashed a few of the commits together) and should be ready to merge.  I'd appreciate at least one set of eyes to go over this PR (preferrably @StephenVavasis, but it seems he might not be getting these messages?).  

Either way, unless there are comments, I'll probably merge in the next day or two.

StephenVavasis:
Sorry for the delay in responding... To the extent that I understand it, it looks good to me except for one point: the new code seems to have a default constructor for SortedSet, SortedDict and SortedMultiDict's with key type=Any.  This does not seem so useful, because there is no obvious way to define an ordering on objects of Any, and so any attempt to use such a structure will lead to an error down the road.  So maybe it is best to disallow this construction?

kmsquire:
That's a good point, thanks for the feedback. In practice, as long as `isless` is defined between all pairs of keys, having a key type of `Any` will still work. But the keys will all be boxed, and `isless` will be dispatched at runtime, making the resulting code less than ideal. 

So, we could disallow it, probably with an explanation. Or allow it with a warning.  Or allow it without a warning, but expect that people might complain about performance. 

Not at a computer, so I can't test right now, but I suggest mimicking whatever happens when an Any array is sorted.

StephenVavasis:
How about if we allow the constructors with explicit arguments to create `SortedDict{Any,D}` without a warning (this is how `sort` works).  The user will encounter an error if two keys are incomparable, and, as you observe, the user will encounter a performance hit due to run-time dispatch.  We can add a warning about this to the documentation.  And how about if `SortedDict()` is invoked (zero-argument constructor), then the resulting key-type is Void instead of Any.  This means that any push! or insert! statement will cause an error at compile time.

kmsquire:
@StephenVavasis, your first proposal on allowing `SortedDict{Any,D}` sounds fine to me.

> And how about if `SortedDict()` is invoked (zero-argument constructor), then the resulting key-type is `Void` instead of `Any`. This means that any `push!` or `insert!` statement will cause an error at compile time.

For `Arrays`, `[]` used to be defined this way (i.e., `[] == None[]`, which is now `Void[]`).  There were a number of confused users, and suggestions by some (including me) to change this to `[] == Any[]`, and this eventually happened.  The argument against this change was (presumably) similar to the argument here: `Any[]` arrays are a performance trap, so we shouldn't encourage them.  The argument for this change was that `None[]` arrays are virtually useless (they can hold as many `nothing`s as you want, and that's all), and the error message you get when trying to use them as anything but `None[]` wasn't useful to users.

Given this, 
1. if we really don't want `SortedDict()` to work as a constructor, we should make the error message explicit and explain to the user what to do, rather than let it fail because the type is `Void`.
2. typically in Julia, when not explicitly writing type parameters in constructors for collections, the type is either inferred from the input, or if there is no input, is set to `Any`.
3. the trend in Julia has not been to prevent users from writing poor performing code (hence, allowing `[] == Any[]`).  
4. there are also situations where ease of typing/readability is much more important than performance

For these reasons and for consistency with the rest of Julia, I'm disinclined to set the default key type to `Void`, and I'm inclined to allow `SortedDict() == SortedDict{Any,Any}()`, rather than force the user to add type information.  The performance hit should definitely be documented, though.

kmsquire:
@StephenVavasis, I'm going to merge this for now, with the behavior I described above.  If you (or anyone else) has further discussion, feel free to add it here or open a new Issue.  I don't think changes will be needed (to the `SortedDict` constructors), but if they are, they can be addressed in a future pull request.

PR #172, opened 2016-01-19
move imports to runtests

rawls238:
@kmsquire 

this cleans up the tests a bit so that most of the imports are done in `runtests` instead of duplicated across test files

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,3 +1,7 @@
+using Base.Test
+using DataStructures
+const IntSet = DataStructures.IntSet
+
 tests = ["intset",
          "deque",
          "sortedcontainers",
@@ -13,6 +17,10 @@ tests = ["intset",
          "trie",
          "list",
          "multidict"]
+
+if length(ARGS) > 0
+    tests = ARGS
+end
 
 for t in tests
     fp = joinpath(dirname(@__FILE__), "test_$t.jl")
--- a/test/test_accumulator.jl
+++ b/test/test_accumulator.jl
@@ -1,8 +1,5 @@
 # Test of accumulators
 
-using DataStructures
-using Base.Test
-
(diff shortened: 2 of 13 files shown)


kmsquire:
Actually, I would prefer to leave the imports at the top of each file.  Without that, it's not possible to run individual tests from the command line--you'd have to run them all, and that's slightly annoying during development.  

Also, it doesn't really hurt anything--if a file has already been imported, the second import is almost a no-op.

(But I really, really do appreciate trying to clean things up--thanks!)

rawls238:
@kmsquire one thing we could do to alleviate that is to do what the `QuantEcon` package does and optionally allow the tests to be run to be passed as command-line arguments. https://github.com/QuantEcon/QuantEcon.jl/blob/master/test/runtests.jl#L34

kmsquire:
Sure, that would be fine.  The main Julia repo does something similar.

DanielArndt:
I like the idea of being able to run tests in groups and individually (it can really cut down turnaround time if you're doing a little TDD or "test backed re-factoring" (TBR?), but some of the test files don't currently run independently.

kmsquire:
@DanielArndt, would you be okay with @rawls238 proposed solution of an extra command line parameter?

DanielArndt:
Certainly. I think another feature that could possibly extend the model QuantEcon uses is to put individual tests in functions, and add the ability to run tests individually (instead of entire files). 

We use a model similar to what I describe internally, and I've been meaning to look at where base Julia is headed for tests, and see if there was a benefit to extracting some of that out.

rawls238:
@DanielArndt that seems interesting to implement, but I think beyond the scope of this PR. Certainly something that would be nice to have. Let me know if you guys are good with merging this

kmsquire:
LGTM!  Feel free to merge.

PR #173, opened 2016-01-19
[WIP] move docs from README to sphinx docs Fixes #108

### LINKED ISSUE #108: Use Sphinx documentation
As the package grows and more data structures come in, it becomes cumbersome to use a single readme page to host all the documentation. 

I propose to convert the documentation to Sphinx (similar to that in the Distributions package), and host the documentation on [ReadTheDocs](https://readthedocs.org).

If there's no objection, I would make a PR later this week.

rawls238:
@lindahua @kmsquire 

Moves docs from the README to sphinx docs, currently the sorted container section is massive and we could probably break it up into subsections but this should be a good start in better to maintain and easier to read docs


kmsquire:
Nice!  

Looks like you got a `.DS_Store` caught in the mix.

I've requested commit access for you.  When that goes through, can you push this to a branch in the main repo, so we can look at it at RTD before merging it?

You've likely already seen this, but just in case, by convention, branches here are prefixed with the user's initials or something similar (`guy` is fine).

Cheers!

rawls238:
sure, I'll move this over tomorrow

hayd:
The direction should really be towards generating documentation via the julia documentation system, and markdown, granted this isn't quite here yet... But the internal documentation is, it would be great to have that for all DataStructures methods.

Otherwise documentation will be forever duplicated (inline docstrings and docs).

PR #175, opened 2016-02-04
sphinx docs

rawls238:
@kmsquire


rawls238:
sorry this was so delayed - you should be able to just pull this down and run `make html` in the doc folder to auto-generate the documentation

DanielArndt:
This would close #108 -- I'm excited to check this out when I get a minute.

rawls238:
unless anyone has any objections with this, I'm going to try to put this up onto `readthedocs` later today.

DanielArndt:
Finally got around to checking this out. Thanks for doing this. I haven't spent a tonne of time with Sphinx, but are themes added after the build (ie. if you were to upload to readthedocs)? The documentation generated looks very basic, but maybe that's just the state of things currently.

rawls238:
going to try to merge this in and publish this. I'm going to revert the README changes for now and will add them back once the docs are in place.

rawls238:
@DanielArndt it was using the wrong theme, I switched it over to match those of other Julia projects I've seen. Check out the docs here: http://datastructuresjl.readthedocs.org/en/latest/

DanielArndt:
Looks great!

One thing I noticed is that when viewing the documentation, in the left-side menu it appears that "Overview of Sorted Containers" is on the same level as everything below it (ie. "Tokens for Sorted Containers", which is actually a sub-section).

Not sure if this is an issue with the theme, or the way it is contructed.

kmsquire:
Looks good--thanks Guy!

PR #176, opened 2016-02-07
For DisjointSets, make union! return the root of the newly merged set

weijianzhang:
For `DisjointSets`, it is useful if `union!` can return the root of the newly merged set. 
For example, it is used in implementing elimination tree for sparse factorization.
(See Algorithm 5.1 of [2])

The implementation of `_union!` is based on [1].  
After this merge, we have 

```julia
julia> a = IntDisjointSets(10)
DataStructures.IntDisjointSets([1,2,3,4,5,6,7,8,9,10],[0,0,0,0,0,0,0,0,0,0],10)

julia> union!(a, 3, 5)
3

julia> a
DataStructures.IntDisjointSets([1,2,3,4,3,6,7,8,9,10],[0,0,1,0,0,0,0,0,0,0],9)
```

References: 
[1] Data Structures and Network Algorithms, Robert Tarjan (1983) Chapter 2 
Disjoint Sets [[url]](http://epubs.siam.org/doi/abs/10.1137/1.9781611970265.ch2)
[2] The Role of Elimination Trees in Sparse Factorization, Joseph W.H. Liu [[url]](http://epubs.siam.org/doi/abs/10.1137/0611010)

### DIFF
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -52,27 +52,31 @@ find_root(s::IntDisjointSets, x::Integer) = find_root_impl!(s.parents, x)
 in_same_set(s::IntDisjointSets, x::Integer, y::Integer) = find_root(s, x) == find_root(s, y)
 
 # merge the subset containing x and that containing y into one
-#
+# and return the root of the new set.
 function union!(s::IntDisjointSets, x::Integer, y::Integer)
     parents = s.parents
     xroot = find_root_impl!(parents, x)
     yroot = find_root_impl!(parents, y)
+    xroot != yroot ?  root_union!(s, xroot, yroot) : xroot
+end
 
-    if xroot != yroot
-        rks = s.ranks
-        @inbounds xrank = rks[xroot]
-        @inbounds yrank = rks[yroot]
+# form a new set that is the union of the two sets whose root elements are
+# x and y and return the root of the new set
+# assume x ≠ y (unsafe)
+function root_union!(s::IntDisjointSets, x::Integer, y::Integer)
+    parents = s.parents
+    rks = s.ranks
+    @inbounds xrank = rks[x]
+    @inbounds yrank = rks[y]
 
-        if xrank < yrank
-            @inbounds parents[xroot] = yroot
-        else
-            @inbounds parents[yroot] = xroot
-            if xrank == yrank
-                rks[xroot] += 1
-            end
-        end
-        @inbounds s.ngroups -= 1
+    if xrank < yrank
+        x, y = y, x
+    elseif xrank == yrank
+        rks[x] += 1
     end
+    @inbounds parents[y] = x
+    @inbounds s.ngroups -= 1
+    x
 end
 
 # make a new subset with an automatically chosen new element x
@@ -119,8 +123,10 @@ find_root{T}(s::DisjointSets{T}, x::T) = find_root(s.internal, s.intmap[x])
 
 in_same_set{T}(s::DisjointSets{T}, x::T, y::T) = in_same_set(s.internal, s.intmap[x], s.intmap[y])
 
-function union!{T}(s::DisjointSets{T}, x::T, y::T)
-    union!(s.internal, s.intmap[x], s.intmap[y])
+for f in (:union!, :root_union!)
+    @eval begin
+        ($f){T}(s::DisjointSets{T}, x::T, y::T) = ($f)(s.internal, s.intmap[x], s.intmap[y])
+    end
 end
 
 function push!{T}(s::DisjointSets{T}, x::T)
--- a/doc/source/disjoint_sets.rst
+++ b/doc/source/disjoint_sets.rst
@@ -10,7 +10,9 @@ A widely used data structure for this purpose is the *Disjoint set forest*.
 Usage::
 
   a = IntDisjointSets(10)      # creates a forest comprised of 10 singletons
-  union!(a, 3, 5)             # merges the sets that contain 3 and 5 into one
+  union!(a, 3, 5)             # merges the sets that contain 3 and 5 into one and returns the root of the new set
+  root_union!(a, x, y)     # merges the sets that have root x and y into one and returns the root of the new set
+  find_root(a, 3)              # finds the root element of the subset that contains 3
   in_same_set(a, x, y)        # determines whether x and y are in the same set
   elem = push!(a)             # adds a single element in a new set; returns the new element
                               # (this operation is often called MakeSet)
@@ -25,3 +27,4 @@ One may also use other element types::
 
 
 Note that the internal implementation of ``IntDisjointSets`` is based on vectors, and is very efficient. ``DisjointSets{T}`` is a wrapper of ``IntDisjointSets``, which uses a dictionary to map input elements to an internal index.
+Note for ``DisjointSets``, ``union!``, ``root_union!`` and ``find_root`` return the index of the root. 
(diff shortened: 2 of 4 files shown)


weijianzhang:
cc @jiahao @kmsquire

src/disjoint_set.jl
```diff
@@ -57,22 +57,28 @@ function union!(s::IntDisjointSets, x::Integer, y::Integer)
     parents = s.parents
```
kmsquire: Not used below.

src/disjoint_set.jl
```diff
@@ -57,22 +57,28 @@ function union!(s::IntDisjointSets, x::Integer, y::Integer)
...
-            @inbounds parents[xroot] = yroot
-        else
-            @inbounds parents[yroot] = xroot
-            if xrank == yrank
-                rks[xroot] += 1
-            end
-        end
-        @inbounds s.ngroups -= 1
+        _union!(s, xroot, yroot)
+    end
```
kmsquire: If `xroot == yroot`, nothing will be returned.  For type-stability, it should return `xroot`.

src/disjoint_set.jl
```diff
@@ -57,22 +57,28 @@ function union!(s::IntDisjointSets, x::Integer, y::Integer)
...
+
+    if xrank < yrank
+        x, y = y, x
+    elseif xrank == yrank
+        rks[x] += 1
     end
+    @inbounds parents[y] = x
+    @inbounds s.ngroups -= 1
+    x
 end
```
kmsquire: Clean implementation.  Not sure if it needs to be a separate function, since it's only used from `union!`, but should be fine.

kmsquire:
I think that makes sense and is a reasonable change.  This should be merged after addressing the comments inline.

kmsquire:
I think that makes sense and is a reasonable change.  This can be merged after addressing the inline comments.

src/disjoint_set.jl
weijianzhang: Thanks for the quick reply. I think `parents` is used in `find_root_impl!` below.

src/disjoint_set.jl
weijianzhang: This is a good point.  I will update it.

src/disjoint_set.jl
weijianzhang: I am considering a use case when a user knows the roots of two sets, say `x` and `y`, and wants to merge them. She could call `DataStructures._union!(a, x, y)`, which is faster than `union!(a, x ,y)`.

src/disjoint_set.jl
kmsquire: Whoops, sorry, I don't know how I missed that.

src/disjoint_set.jl
kmsquire: In Julia, it's probably not a good idea to call a function which isn't part of the official interface, especially one with an underscore as the first character.  How about calling it `root_union!` (or some other name), documenting and exporting it?

Of course, then it should also be implemented for `DisjointSets`.

kmsquire:
A couple of more (quick) requests: Please

1. add a couple of quick tests that verify this functionality
2. add an example to the documentation.
3. squash your commits when ready to merge

src/disjoint_set.jl
weijianzhang: Nice idea! I like it.

weijianzhang:
@kmsquire all the requests are addressed. Please let me know if you have any other comments.

README.rst
```diff
@@ -141,10 +143,13 @@ One may also use other element types::
...
   union!(a, "a", "b")
   in_same_set(a, "c", "d")
+  root_union!(a, "a", "d")
   push!(a, "f")
 
 
-Note that the internal implementation of ``IntDisjointSets`` is based on vectors, and is very efficient. ``DisjointSets{T}`` is a wrapper of ``IntDisjointSets``, which uses a dictionary to map input elements to an internal index.
+Note that the internal implementation of ``IntDisjointSets`` is based on vectors, and is very efficient. ``DisjointSets{T}`` is a wrapper of ``IntDisjointSets``, which uses a dictionary to map input elements to an internal index. 
+As a result, ``union!(a, "a", "b")`` and ``root_union!(a, "a", "d")`` returns the indices of the
+root elements rather than the root elements themselves.
```
kmsquire: I'm wondering...

1. Is there a way to go from index to element?
2. If someone is manipulating non-integer `DisjointSets`, is returning the resulting index is useful?  Can they use that index in further processing?  How?

weijianzhang: > Is there a way to go from index to element?

Since ``DisjointSets`` use a dictionary to wrap ``IntDisjointSets``, "to go from index to element" means get a key by value in a dictionary. This is doable but not very efficient, perhaps something like
```julia
for (key, value) in intmap
     if value == index
          return key
     end
end
```
Perhaps we could add a new field to ``DisjointSets``, but this is a major change and shoud be considered in a different pull request. By the way, currently ``find_root`` only  returns the index of the root rather than the root itself.

> If someone is manipulating non-integer DisjointSets, is returning the resulting index is useful? Can they use that index in further processing? How?

Yes, the index can be used in further processing. Just use the simple example from the documentation. 
```julia
> list = ["a", "b", "c", "d"]
> a = DisjointSets{AbstractString}(list)
> indx = union!(a, "a", "b")
> list[indx]
"a"
```

kmsquire:
@weijianzhang, sorry, I was about to merge, but it looks like something else snuck in that caused a merge conflict (probably in the exports).  Can you rebase on the current master?  After that, I thing this is mergable.

weijianzhang:
@kmsquire  sure, no problem. I will fix the conflicts. Thanks.

kmsquire:
(Thanks... and can you squash those please?)

kmsquire:
(Thanks... and can you squash those please?)

weijianzhang:
Hi @kmsquire, sorry but I don't know how. I can only squash consecutive commits but there are a few commits between my last commit and this commit. I've made a few attempts but was unsuccessful. Could you let me know how I can do this?

kmsquire:
Hi @weijianzhang, if you run 

```
git fetch
git rebase origin/master
```

it should move your commits to the most recent position.  From there, you should be able to squash and force push to this branch.

If that doesn't work for some reason, just post back here and we'll figure something out.

DanielArndt:
@weijianzhang Looks like you merged the current master, instead of rebasing. Rebasing re-writes history as if your changes came after. Here's one way to fix:

```
git checkout weijianzhang/master
git rebase origin/master
```
You will need to fix the README conflicts here (simple, remove all the stuff on the bottom). I wouldn't bother adding the doc changes, since we can retrieve those from the merge you already did. When you're done

```
git rebase --continue
```
Take a look at what you're missing
```
git diff ..weijianzhang/master
```

Apply those changes
```
git diff ..wz/master | git apply
```

Add your missing changes:
```
git add doc/source/disjoint_sets.rst
```

Amend them (add them) to your last commit (which should now, be your first commit)
```
git commit --amend
```

If everything looks right now, you should be able to push this to your remote.

weijianzhang:
@kmsquire @DanielArndt Thanks a lot for your help. I think it is fixed now.

DanielArndt:
Great! The history looks a lot cleaner now.

rawls238:
@kmsquire should we bump the version to v0.4.3 now that these recent changes have been merged in?

kmsquire:
Yep. Can you?

rawls238:
Sure, will do so tonight

PR #177, opened 2016-02-07
Clean up Deque iterators, collect

kmsquire:
* Added reverse `Deque` iterator
* Removed extraneous state from `Deque` iterator


rawls238:
I built on this in #171

rawls238:
these changes were incorporated into https://github.com/JuliaLang/DataStructures.jl/pull/171, thanks again @kmsquire

PR #179, opened 2016-02-08
replace README docs with link to readthedocs

rawls238:
@kmsquire @DanielArndt 

Let me know if you guys think any changes should be made to the docs before merging this in: 

http://datastructuresjl.readthedocs.org/en/latest

@kmsquire I'm not sure if the `readthedocs` github webhooks are properly set up so that the docs can be updated (I don't have the github permissions to look in settings). Could you confirm that they're set / if they're not set them?


DanielArndt:
Looks great!

One thing I noticed is that when viewing the documentation, in the left-side menu it appears that "Overview of Sorted Containers" is on the same level as everything below it (ie. "Tokens for Sorted Containers", which is actually a sub-section).

Not sure if this is an issue with the theme, or the way it is contructed.

kmsquire:
@rawls238 I added you as a maintainer on DataStructures.  

@DanielArndt, @StephenVavasis, if you have or create an account on ReadTheDocs, I can add you as well.

kmsquire:
Actually, ignore that last comment: that requires someone who's an owner on JuliaLang, and I'm not, so I don't have access either.

kmsquire:
@StefanKarpinski @ViralBShah can one of you modify the permissions on RTD to allow myself and @rawls238 to maintain DataStructures.jl there?

kmsquire:
Okay, nevermind. I do actually have access.  Sorry for the noise.  ( @rawls238, you should as well.)

DanielArndt:
I created an account and linked it to my github. Is there anything else I need to do?

rawls238:
@danielarndt that's valid. I think that has to do with how I set up the index file, I'll try to change that around before merging this in

rawls238:
@DanielArndt fixed the issue with sorted containers

hayd:
As I commented in the closed #173 this feels like the wrong direction. Nothing in DataStructures has any ? help (which will have to be markdown at some point).

I guess I should **do** rather than complain...

DanielArndt:
@hayd Are you suggesting the best thing to do would be to move all this information to the function definitions, inline as markdown, and wait for the Julia documentation system to be able to generate the documentation?

Sorry if I'm being a little dense, I'm not entirely familiar with the state of the documentation system in the language. I know that base Julia has a lot of documentation in `rst`/sphinx format. Is it moving away from that entirely in favour of documentation generated from inline comments? As far as I remember, I had to change the documentation in order to generate the inline comments in base (which seems backwards).

hayd:
yes, doc generation will be ready soon (next few weeks hopefully). yes, base is moving away from rst entirely. Moving to markdown would make the transition (to inline docstrings and generated online docs) for DataStructures much easier.

rawls238:
sweet, we could try running the doc files through [pandoc](http://pandoc.org/) to convert them to markdown.

StefanKarpinski:
I'm not sure if it's helpful or not, but https://github.com/JuliaLang/julia/tree/sk/markdown implements a Sphinx backend that generates Markdown that's a speculative version of JuliaMarkdown.

kmsquire:
Just pinging @StephenVavasis, who previously explicitly converted the `README` file to *rst* (by hand, I believe), because he was concerned about the expressiveness of markdown.

PR #180, opened 2016-02-08
Add indexing to OrderedSets

ahwillia:
Discussed here: https://github.com/JuliaLang/DataStructures.jl/issues/178

### DIFF
--- a/src/orderedset.jl
+++ b/src/orderedset.jl
@@ -29,6 +29,9 @@ pop!(s::OrderedSet, x) = (pop!(s.dict, x); x)
 pop!(s::OrderedSet, x, deflt) = pop!(s.dict, x, deflt) == deflt ? deflt : x
 delete!(s::OrderedSet, x) = (delete!(s.dict, x); s)
 
+getindex(x::OrderedSet,i::Int) = x.dict.keys[i]
+endof(x::OrderedSet) = endof(x.dict.keys)
+
 union!(s::OrderedSet, xs) = (for x in xs; push!(s,x); end; s)
 setdiff!(s::OrderedSet, xs) = (for x in xs; delete!(s,x); end; s)
 setdiff!(s::Set, xs::OrderedSet) = (for x in xs; delete!(s,x); end; s)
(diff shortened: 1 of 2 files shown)


ahwillia:
I was using an outdated version, so the code is tweaked slightly. Just by providing indexing we can rely on the `findfirst` and `findin` functions provided in Base to find element indices...

kmsquire:
LGTM.  @DanielArndt @rawls238, any thoughts?  If you have nothing against this, I think it can be merged.

rawls238:
this lgtm as well

timholy:
As discovered by @StephenVavasis, indexing does not currently play well with deletion:
```julia
julia> function testindexing()
           s = OrderedSet([i for i = 1:12])
           delete!(s, 7)
           prerh = s[7]
           OrderedCollections.rehash!(s.dict)     # see #392 for the OrderedCollections name
           postrh = s[7]
           return (prerh, postrh)
       end

julia> testindexing()
(7, 8)
```
(I did this inside a function because simply `show`ing at the REPL rehashes the dictionary, so it's easy for this problem to be disguised.)

Our options appear to be to check the `dirty` flag before every indexing operation, or just drop indexing altogether. It's worth noting that one can naturally `rehash!` once when you first begin iteration (and hence is efficient), but indexing requires a check each time. It is just a `Bool` check, so perhaps not a big deal, but since it's a `mutable` struct it's not entirely free.

So, worth asking, how useful is this really? If we do drop it, I can add a deprecation that also has the proper behavior.

yakir12:
Sorry for butting in, but if we shouldn't index but use iteration instead, why not add a function like this:
```julia
function getindex(s, i)
    i < 1 && throw(BoundsError(s, i))
    for (j, sj) in enumerate(s)
        j == i && return sj
    end
    throw(BoundsError(s, i))
end
```

bzinberg:
I've found myself often wanting an ordered dict object that is ergonomic to index into either by dict key or by index in the key sequence.  And, often I don't need the ability to delete from this data structure.  For this case I've created [`AppendOnlyOrderedDict`](https://gist.github.com/bzinberg/fb3981dd2251fb25cdbcd4f90b8b9a72) which is basically a dict and a vector (which records the key order) glued together plus some syntactic sugar for the second kind of indexing.
```julia
>>> d = AppendOnlyOrderedDict(:a => 2, :b => 4, :c => 6);
>>> d[:b]
4
>>> d.seq[2]
:b => 4
>>> d.seqvals[2]
4
>>> d.seqvals[2:end]
2-element Array{Int64,1}:
 4
 6
```

PR #182, opened 2016-02-15
inline documentation

abhijithch:
Basic inline documentation of the exported functions.

### DIFF
--- a/src/accumulator.jl
+++ b/src/accumulator.jl
@@ -12,6 +12,11 @@ counter(T::Type) = Accumulator(T,Int)
 Accumulator{T,V<:Number}(dct::Dict{T,V}) = Accumulator{T,V}(copy(dct))
 counter{T}(dct::Dict{T,Int}) = Accumulator{T,Int}(copy(dct))
 
+"""
+    counter{T}(seq::AbstractArray)
+
+Returns an `Accumulator` object containing the elements from `seq`.
+"""
 function counter{T}(seq::AbstractArray{T})
     ct = counter(T)
     for x in seq
--- a/src/deque.jl
+++ b/src/deque.jl
@@ -1,5 +1,4 @@
 # Block-based deque
-
 #######################################
 #
 #  DequeBlock
@@ -74,18 +73,33 @@ type Deque{T}
     Deque() = Deque{T}(DEFAULT_DEQUEUE_BLOCKSIZE)
 end
 
+"""
+    deque(T)
+
+Create a deque of type `T`.
+"""
 deque{T}(::Type{T}) = Deque{T}()
 
 isempty(q::Deque) = q.len == 0
 length(q::Deque) = q.len
 num_blocks(q::Deque) = q.nblocks
 
+"""
+    front(q::Deque)
+
+Returns the first element of the deque `q`.
+"""
 function front(q::Deque)
     isempty(q) && throw(ArgumentError("Deque must be non-empty"))
     blk = q.head
     blk.data[blk.front]
 end
 
+"""
+    back(q::Deque)
+
+Returns the last element of the deque `q`.
+"""
 function back(q::Deque)
     isempty(q) && throw(ArgumentError("Deque must be non-empty"))
     blk = q.rear
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -49,6 +49,11 @@ end
 
 find_root(s::IntDisjointSets, x::Integer) = find_root_impl!(s.parents, x)
 
+"""
+    in_same_set(s::IntDisjointSets, x::Integer, y::Integer)
+
+Returns `true` if `x` and `y` belong to the same subset in `s` and `false` otherwise.
+"""
 in_same_set(s::IntDisjointSets, x::Integer, y::Integer) = find_root(s, x) == find_root(s, y)
 
 # merge the subset containing x and that containing y into one
@@ -119,6 +124,11 @@ end
 length(s::DisjointSets) = length(s.internal)
 num_groups(s::DisjointSets) = num_groups(s.internal)
 
+"""
+    find_root{T}(s::DisjointSets{T}, x::T)
+
+Finds the root element of the subset in `s` which has the element `x` as a member.
+"""
 find_root{T}(s::DisjointSets{T}, x::T) = find_root(s.internal, s.intmap[x])
 
 in_same_set{T}(s::DisjointSets{T}, x::T, y::T) = in_same_set(s.internal, s.intmap[x], s.intmap[y])
--- a/src/heaps/binary_heap.jl
+++ b/src/heaps/binary_heap.jl
@@ -140,6 +140,12 @@ function push!{T}(h::BinaryHeap{T}, v::T)
     _heap_bubble_up!(h.comparer, valtree, length(valtree))
 end
 
+
+"""
+    top(h::BinaryHeap)
+
+Returns the element at the top of the heap `h`.
+"""
 top(h::BinaryHeap) = h.valtree[1]
 
 pop!{T}(h::BinaryHeap{T}) = _binary_heap_pop!(h.comparer, h.valtree)
--- a/src/heaps/mutable_binary_heap.jl
+++ b/src/heaps/mutable_binary_heap.jl
@@ -212,6 +212,11 @@ top(h::MutableBinaryHeap) = h.nodes[1].value
 
 pop!{T}(h::MutableBinaryHeap{T}) = _binary_heap_pop!(h.comparer, h.nodes, h.node_map)
 
+"""
+    update!{T}(h::MutableBinaryHeap{T}, i::Int, v::T)
+
+Replace the element at index `i` in heap `h` with `v`.
+"""
 function update!{T}(h::MutableBinaryHeap{T}, i::Int, v::T)
     nodes = h.nodes
     nodemap = h.node_map
--- a/src/ordereddict.jl
+++ b/src/ordereddict.jl
@@ -8,7 +8,12 @@ import Base: haskey, get, get!, getkey, delete!, push!, pop!, empty!,
              hash, eltype, KeyIterator, ValueIterator, convert, copy,
              merge
 
+"""
+    OrderedDict
 
+`OrderedDict`s are  simply dictionaries  whose entries  have a  particular order.  The order
+refers to insertion order, which allows deterministic iteration over the dictionary or set.
+"""
 type OrderedDict{K,V} <: Associative{K,V}
     slots::Array{Int32,1}
     keys::Array{K,1}
--- a/src/queue.jl
+++ b/src/queue.jl
@@ -7,8 +7,7 @@ end
 """
     Queue(T[, blksize::Integer=1024])
 
-This is a  constructor to create an object of  a `Queue`.`T` specifies
-the data type of the member elements of the `Queue`.
+Create a `Queue` objet containing elements of type `T`.
 """
 Queue{T}(ty::Type{T}) = Queue(Deque{T}())
 Queue{T}(ty::Type{T}, blksize::Integer) = Queue(Deque{T}(blksize))
@@ -32,7 +31,7 @@ end
 """
     dequeue!(s::Queue)
 
-Removes an element from the front of the queue `s`.
+Removes an element from the front of the queue `s` and removes it.
 """
 dequeue!(s::Queue) = shift!(s.store)
 
--- a/src/queue.jl
+++ b/src/queue.jl
@@ -4,6 +4,12 @@ type Queue{T}
     store::Deque{T}
 end
 
+"""
+    Queue(T[, blksize::Integer=1024])
+
+This is a  constructor to create an object of  a `Queue`.`T` specifies
+the data type of the member elements of the `Queue`.
+"""
 Queue{T}(ty::Type{T}) = Queue(Deque{T}())
 Queue{T}(ty::Type{T}, blksize::Integer) = Queue(Deque{T}(blksize))
 
@@ -13,11 +19,21 @@ length(s::Queue) = length(s.store)
 front(s::Queue) = front(s.store)
 back(s::Queue) = back(s.store)
 
+"""
+    enqueue!(s::Queue, x)
+
+Inserts the value `x` to the end of the queue `s`.
+"""
 function enqueue!(s::Queue, x)
     push!(s.store, x)
     s
 end
 
+"""
+    dequeue!(s::Queue)
+
+Removes an element from the front of the queue `s`.
+"""
 dequeue!(s::Queue) = shift!(s.store)
 
 # Iterators
--- a/src/tokens2.jl
+++ b/src/tokens2.jl
@@ -66,7 +66,15 @@ end
          ii[2].address == 1?                       2 :
          ii[2].address == 2?                       3 : 1
 
+"""
+    compare(m::SAContainer, s::IntSemiToken, t::IntSemiToken)
 
+Here, `s`  and `t`  are semitokens  for the  same container  `m`; this
+function determines the  relative positions of the  data items indexed
+by `(m,s)` and  `(m,t)` in the sorted order. The  return value is `-1`
+if `(m,s)` precedes `(m,t)`, `0` if they are equal, and `1` if `(m,s)`
+succeeds `(m,t).`
+"""
 @inline compare(m::SAContainer, s::IntSemiToken, t::IntSemiToken) =
       compareInd(m.bt, s.address, t.address)
 
--- a/src/tokens2.jl
+++ b/src/tokens2.jl
@@ -69,11 +69,10 @@ end
 """
     compare(m::SAContainer, s::IntSemiToken, t::IntSemiToken)
 
-Here, `s`  and `t`  are semitokens  for the  same container  `m`; this
-function determines the  relative positions of the  data items indexed
+Determines the  relative positions of the  data items indexed
 by `(m,s)` and  `(m,t)` in the sorted order. The  return value is `-1`
 if `(m,s)` precedes `(m,t)`, `0` if they are equal, and `1` if `(m,s)`
-succeeds `(m,t).`
+succeeds `(m,t)`. `s`  and `t`  are semitokens  for the  same container `m`.
 """
 @inline compare(m::SAContainer, s::IntSemiToken, t::IntSemiToken) =
       compareInd(m.bt, s.address, t.address)


src/accumulator.jl
```diff
@@ -12,6 +12,16 @@ counter(T::Type) = Accumulator(T,Int)
 Accumulator{T,V<:Number}(dct::Dict{T,V}) = Accumulator{T,V}(copy(dct))
 counter{T}(dct::Dict{T,Int}) = Accumulator{T,Int}(copy(dct))
 
+"""
+```
```
nalimilan: I haven't looked at the rest, but the ` are not needed. OTOH, add 4 spaces before code lines (such as signatures).

kmsquire:
@abhijithch, thanks for doing this!  Not sure I'll have the chance to look soon, but hopefully you'll get a review and merge soon. @hayd, can you take a look?

src/accumulator.jl
hayd: The convention is four spaces for the signature code block. Other code blocks can use `s.

hayd:
this looks good to me!  :+1:

src/accumulator.jl
DanielArndt: Might be useful to point at an example, so here is one I was just digging through:

https://github.com/JuliaLang/julia/blob/master/base/sharedarray.jl#L38-L51

src/deque.jl
```diff
@@ -74,18 +73,45 @@ type Deque{T}
...
 
+"""
+```
+front(q::Deque)
+```
+
+### Args:
+* `q`: A deque type datastructure.
+
+Returns the first element of the deque `s`.
```
DanielArndt: Should this be "Returns the first element of the deque **`q`**."?

src/deque.jl
```diff
@@ -74,18 +73,45 @@ type Deque{T}
...
 end
 
+"""
+```
+back(q::Deque)
+```
+### Args:
+* `q`: A deque type datastructure.
+
+Returns the last element of the deque `s`.
```
DanielArndt: Should this be "Returns the last element of the deque **`q`**."?

src/disjoint_set.jl
```diff
@@ -119,6 +129,15 @@ end
 length(s::DisjointSets) = length(s.internal)
 num_groups(s::DisjointSets) = num_groups(s.internal)
 
+"""
+    find_root
```
DanielArndt: Should this have the full function signature like the others?

src/hashdict.jl
```diff
@@ -254,6 +254,19 @@ function _compact_order{K,V}(h::HashDict{K,V,Ordered})
...
+sizehint!(d, n)
+```
+
+### Args:
+* `d`: Object of the datastructure.
+* `n`: The size to reserve in `d`.
+
+Reserves capacity for `n` elements within `d`.
+"""
+function sizehint! end
```
DanielArndt: Oops. Should this be here?

src/hashdict.jl
```diff
@@ -254,6 +254,19 @@ function _compact_order{K,V}(h::HashDict{K,V,Ordered})
     nothing
 end
 
+"""
+```
+sizehint!(d, n)
```
DanielArndt: Is it necessary to re-define doc strings for functions like this, where there is already a definition in base Julia? Also, the argument names don't match up.

src/heaps.jl
```diff
@@ -63,7 +63,14 @@ include("heaps/binary_heap.jl")
 include("heaps/mutable_binary_heap.jl")
 
 # generic functions
+"""
+    extract_heap!(h)
 
+### Args:
+* `h`: `h` and abstract heap.
```
DanielArndt: Should this read "`h`: an abstract heap."?

src/heaps/binary_heap.jl
```diff
@@ -140,6 +140,14 @@ function push!{T}(h::BinaryHeap{T}, v::T)
     _heap_bubble_up!(h.comparer, valtree, length(valtree))
 end
 
+
+"""
+```
+top(h::BinaryHeap)
+```
+
+Returns the top the heap.
```
DanielArndt: "Returns the top **of** the heap.

src/heaps/mutable_binary_heap.jl
```diff
@@ -212,6 +212,18 @@ top(h::MutableBinaryHeap) = h.nodes[1].value
...
+```
+update!{T}(h::MutableBinaryHeap{T}, i::Int, v::T)
+```
+
+### Args:
+* `h`: Heap derived from `AbstractMutableHeap`.
+* `i`: The handle to the element which needs to be updated.
+* `v`: The updated value.
+
+Heaps of type `AbstractMutableHeap` allows to change its values after being pushed through this function `update!`.
```
DanielArndt: Hmm, I'm not so sure about this description but I don't have any experience with it, so re-writing it is proving difficult. But heres an attempt derived from the old documentation:

"Updates the value of an element in heap `h` (referred to by the handle i) with the value `v`"

DanielArndt:
Hopefully some constructive comments there. Thanks for the PR!

abhijithch:
Thanks @DanielArndt for the comments, will update the PR.

abhijithch:
@nalimilan Is there any standard best practices for documentation available anywhere? Its just that I saw the tick in another package and assumed that would be the right way, anyways will change it to four spaces. Thanks.

src/accumulator.jl
```diff
@@ -12,6 +12,14 @@ counter(T::Type) = Accumulator(T,Int)
 Accumulator{T,V<:Number}(dct::Dict{T,V}) = Accumulator{T,V}(copy(dct))
 counter{T}(dct::Dict{T,Int}) = Accumulator{T,Int}(copy(dct))
 
+"""
+    counter{T}(seq::AbstractArray{T})
+
+### Args:
```
nalimilan: As I noted on the StatsBase PR, I think in most cases the argument list isn't useful. Here, `seq` is mentioned below anyway, and the signature gives the expected type.

src/accumulator.jl
```diff
@@ -12,6 +12,14 @@ counter(T::Type) = Accumulator(T,Int)
 Accumulator{T,V<:Number}(dct::Dict{T,V}) = Accumulator{T,V}(copy(dct))
 counter{T}(dct::Dict{T,Int}) = Accumulator{T,Int}(copy(dct))
 
+"""
+    counter{T}(seq::AbstractArray{T})
```
nalimilan: `T` doesn't bring anything here

src/deque.jl
```diff
@@ -74,18 +73,40 @@ type Deque{T}
     Deque() = Deque{T}(DEFAULT_DEQUEUE_BLOCKSIZE)
 end
 
+"""
+    deque(::Type{T})
+
+Create a deque of type `T`
```
nalimilan: The signature could be written in a nicer way as `deque(T)` (even if under the hood we need a more complex one).

src/disjoint_set.jl
```diff
@@ -49,6 +49,16 @@ end
...
 
+"""
+    in_same_set(s::IntDisjointSets, x::Integer, y::Integer)
+
+### Args:
+* `s`: An object of type `IntDisjointSets`. 
+* `x`: First of two elements in `s`.
+* `y`: Second element to compare with `x`.
+
+The function returns `true` if `x` and `y` belong to the same subset in `s` and `false otherwise.`
```
nalimilan: The ending ` is at the wrong position.

src/deque.jl
```diff
@@ -74,18 +73,40 @@ type Deque{T}
...
 
 isempty(q::Deque) = q.len == 0
 length(q::Deque) = q.len
 num_blocks(q::Deque) = q.nblocks
 
+"""
+    front(q::Deque)
+
+### Args:
+* `q`: A deque type datastructure.
```
nalimilan: Argument list is clearly redundant here too.

src/disjoint_set.jl
nalimilan: Ah, and expressions like "the function" are generally not useful: better get straight to the point.

src/disjoint_set.jl
```diff
@@ -119,6 +129,15 @@ end
...
 num_groups(s::DisjointSets) = num_groups(s.internal)
 
+"""
+    find_root{T}(s::DisjointSets{T}, x::T)
+
+### Args:
+* `s`: DisjointSets which has the element `x` as a member of one of its subsets.
+* `x`: The element which needs to identified as belonging to which subset.
+
+`find_root()` finds the root element of the subset that contains `x`.
```
nalimilan: Do not repeat the name of the function.

src/ordereddict.jl
```diff
@@ -8,6 +8,11 @@ import Base: haskey, get, get!, getkey, delete!, push!, pop!, empty!,
              hash, eltype, KeyIterator, ValueIterator, convert, copy,
              merge
 
+"""
+    OrderedDict
+
+OrderedDicts are simply dictionaries whose entries have a particular order. The order refers to insertion order, which allows deterministic iteration over the dictionary or set.
+"""
```
nalimilan: Remove the empty line here.

src/ordereddict.jl
```diff
@@ -8,6 +8,11 @@ import Base: haskey, get, get!, getkey, delete!, push!, pop!, empty!,
              hash, eltype, KeyIterator, ValueIterator, convert, copy,
              merge
 
+"""
+    OrderedDict
+
+OrderedDicts are simply dictionaries whose entries have a particular order. The order refers to insertion order, which allows deterministic iteration over the dictionary or set.
```
nalimilan: Too long line.

nalimilan: And ` around `OrderedDict`s.

src/queue.jl
```diff
@@ -4,6 +4,15 @@ type Queue{T}
     store::Deque{T}
 end
 
+"""
+    Queue{T}(ty::Type{T}, blksize::Integer)
```
nalimilan: Here again I think you can write `ty::Type{T}` as `T`, and get rid of the type parameter.

src/deque.jl
nalimilan: Ah, and missing ending period.

nalimilan: Should also mention that `blksize` can be passed to set blocksize (see constructor above). Anyway, this needs to be symmetric with the `Queue` docstring.

src/heaps/binary_heap.jl
```diff
@@ -140,6 +140,12 @@ function push!{T}(h::BinaryHeap{T}, v::T)
     _heap_bubble_up!(h.comparer, valtree, length(valtree))
 end
 
+
+"""
+    top(h::BinaryHeap)
+
+Returns the top of the heap.
```
nalimilan: Maybe "element at the top of the heap"?

src/heaps.jl
```diff
@@ -63,7 +63,14 @@ include("heaps/binary_heap.jl")
 include("heaps/mutable_binary_heap.jl")
 
 # generic functions
+"""
+    extract_heap!(h)
```
nalimilan: Missing type for `h`.

src/heaps.jl
```diff
@@ -63,7 +63,14 @@ include("heaps/binary_heap.jl")
...
 
+### Args:
+* `h`: an abstract heap.
+
+This function returns an array which has the elements from the heap extracted.
```
nalimilan: No need for "this function". Also, this string doesn't explain much: it merely repeats "extracted". IIUC, what's important here is that `r` is emptied.

src/heaps/mutable_binary_heap.jl
```diff
@@ -212,6 +212,16 @@ top(h::MutableBinaryHeap) = h.nodes[1].value
...
 
+"""
+    update!{T}(h::MutableBinaryHeap{T}, i::Int, v::T)
+
+### Args:
+* `h`: Heap derived from `AbstractMutableHeap`.
+* `i`: The handle to the element which needs to be updated.
+* `v`: The updated value.
+
+Updates the value of an element in heap `h` (referred to by the handle `i`) with the value `v`.
```
nalimilan: Shouldn't "handle" be simply "index"? If so, I'd simply say "Replace the element at index `i` in heap `h` with `v`.

src/queue.jl
```diff
@@ -13,11 +22,28 @@ length(s::Queue) = length(s.store)
...
 back(s::Queue) = back(s.store)
 
+"""
+    enqueue!(s::Queue, x)
+
+### Args:
+* `s`: A queue into which to insert a value. 
+* `x`: The value to be inserted.
+
+`enqueue!` inserts the value `x` into the queue `s`.
```
nalimilan: Do not repeat name of function.

nalimilan: Also, "insert" could be more explicit about whether it appends to the end of the queue, or to its start.

src/queue.jl
```diff
@@ -13,11 +22,28 @@ length(s::Queue) = length(s.store)
...
 function enqueue!(s::Queue, x)
     push!(s.store, x)
     s
 end
 
+"""
+    dequeue!(s::Queue)
+
+### Args:
+* `s`: The queue from which to remove the least significant entry.
```
nalimilan: "Least significant" sounds weird to me (but anyway remove the argument lists ;-).

src/tokens2.jl
```diff
@@ -66,7 +66,16 @@ end
...
 
+"""
+    compare(m::SAContainer, s::IntSemiToken, t::IntSemiToken)
 
+### Args:
+* `m`: SortedDict or SortedMultiDict.
+* `s`: Semitoken
+* `t`: Semitoken
+
+This function determines the relative positions of the data items indexed by (m,s) and (m,t) in the sorted order.
```
nalimilan: I'm not familiar with this at all, but I must say I don't understand from this description what the function does. Maybe detail a bit more? Or give an example?

src/accumulator.jl
```diff
@@ -12,6 +12,14 @@ counter(T::Type) = Accumulator(T,Int)
...
+
+### Args:
+* `seq`: An AbstractArray
+
+`counter()` function returns an `Accumulator` object with the elements from `seq` assembled into.
```
nalimilan: Do not repeat function name. Also "assembled into" sounds weird to me: maybe just "containing the elements..."?

nalimilan:
@abhijithch Unfortunately, I don't know whether there's a style guide yet as regards the docs. I'm mostly taking example on what's in Julia Base.

src/tokens2.jl
DanielArndt: I was struggling a bit with this one too

StephenVavasis: Daniel and Milan,

I'm not completely clear on what you are doing in this thread, but the 'compare' function is documented in the DataStructures.jl documentation:

compare(sc,st1,st2)
    Here, st1 and st2 are semitokens for the same container sc; this function determines the relative positions of the data items indexed by (sc,st1) and (sc,st2) in the sorted order. The return value is -1 if (sc,st1) precedes (sc,st2), 0 if they are equal, and 1 if (sc,st1) succeeds (sc,st2). This function compares the tokens by determining their relative position within the tree without dereferencing them. For SortedDict it is mostly equivalent to comparing deref_key((sc,st1)) to deref_key((sc,st2)) using the ordering of the SortedDict except in the case that either (sc,st1) or (sc,st2) is the before-start or past-end token, in which case the deref operation will fail. Which one is more efficient depends on the time-complexity of comparing two keys. Similarly, for SortedSet it is mostly equivalent to comparing deref((sc,st1)) to deref((sc,st2)). For SortedMultiDict, this function is not equivalent to a key comparison since two items in a SortedMultiDict with the same key are not necessarily the same item. Time: O(log n)

nalimilan: I think we were just noting that the docstring isn't enough to understand what the function does. If there's a detailed documentation about this somewhere, why not move it to the docstring?

nalimilan:
I've just made a PR to try defining guidelines: https://github.com/JuliaLang/julia/pull/15136

src/ordereddict.jl
```diff
@@ -8,7 +8,11 @@ import Base: haskey, get, get!, getkey, delete!, push!, pop!, empty!,
              hash, eltype, KeyIterator, ValueIterator, convert, copy,
              merge
 
+"""
+    OrderedDict
 
+'OrderedDict's are simply dictionaries whose entries have a particular order. The order refers to insertion order, which allows deterministic iteration over the dictionary or set.
```
DanielArndt: I think the intention here was to have ticks (\`) instead of apostrophes (') so that `OrderedDict` is rendered as code/monospace

src/tokens2.jl
```diff
@@ -66,7 +66,11 @@ end
          ii[2].address == 1?                       2 :
          ii[2].address == 2?                       3 : 1
 
+"""
+    compare(m::SAContainer, s::IntSemiToken, t::IntSemiToken)
 
+Here, `s` and `t` are semitokens for the same container `m`; this function determines the relative positions of the data items indexed by `(m,s)` and `(m,t)` in the sorted order. The return value is `-1` if `(m,s)` precedes `(m,t)`, `0` if they are equal, and `1` if `(m,s)` succeeds `(m,t)`.
```
DanielArndt: I think you can drop the first part:

~~Here, `s` and `t` are semitokens for the same container `m`; this function~~

DanielArndt: This should probably be broken into lines of 80 chars max.

src/ordereddict.jl
DanielArndt: This should probably be broken into lines of 80 chars max

DanielArndt:
Just a few more comments. This is looking really good!

src/queue.jl
```diff
@@ -4,6 +4,11 @@ type Queue{T}
     store::Deque{T}
 end
 
+"""
+    Queue(ty{T}[, blksize::Integer])
+
+This is a constructor to create an object of a `Queue`.The block size `blksize` by default is `1024`.
```
nalimilan: Instead of "The block size `blksize` by default is `1024`.", just write `blksize::Integer=1024` in the signature.

src/queue.jl
```diff
@@ -4,6 +4,11 @@ type Queue{T}
     store::Deque{T}
 end
 
+"""
+    Queue(ty{T}[, blksize::Integer])
```
nalimilan: Not sure what `ty{T}` is here. I would just write it as `T`, and mention the meaning of this argument in the description.

src/queue.jl
```diff
@@ -13,11 +18,21 @@ length(s::Queue) = length(s.store)
 front(s::Queue) = front(s.store)
 back(s::Queue) = back(s.store)
 
+"""
+    enqueue!(s::Queue, x)
+
+Inserts the value `x` to the front of the queue `s`.
```
nalimilan: Are you sure this isn't the *end*/*back* of the queue?

DanielArndt: good catch!

src/queue.jl
```diff
@@ -13,11 +19,21 @@ length(s::Queue) = length(s.store)
...
+"""
 function enqueue!(s::Queue, x)
     push!(s.store, x)
     s
 end
 
+"""
+    dequeue!(s::Queue)
+
+Removes an element from the back of the queue `s`.
```
DanielArndt: Similarly to how `enqueue!` was flipped, so is this.

`dequeue!` should (and does) remove an item from the front of the queue

rawls238:
@nalimilan @DanielArndt @abhijithch should we merge this in?

kmsquire:
It would be good if the commits were squashed, although that's probably less necessary with documentation commits (unless they break something).

DanielArndt:
Sorry, this hadn't fallen off my radar, I've just been super overwhelmed lately. I wanted to give this the solid review it deserves instead of trickling in comments.

I think this can be merged as is. At some point I think it would be worth taking another stab at the [compare](https://github.com/abhijithch/DataStructures.jl/blob/inlinedoc/src/tokens2.jl#L70-L76) documentation. It doesn't seem to follow the same tone as the other function docs, but I've been unsuccessful at rewording it so far.

Thanks a lot @abhijithch -- and sorry I slowly trickled in comments. I hope that wasn't too frustrating.

src/queue.jl
```diff
@@ -4,6 +4,12 @@ type Queue{T}
     store::Deque{T}
 end
 
+"""
+    Queue(T[, blksize::Integer=1024])
+
+This is a  constructor to create an object of  a `Queue`.`T` specifies
```
nalimilan: Could simply be "Create a `Queue` objet containing elements of type `T`."

src/queue.jl
```diff
@@ -13,11 +19,21 @@ length(s::Queue) = length(s.store)
...
+"""
 function enqueue!(s::Queue, x)
     push!(s.store, x)
     s
 end
 
+"""
+    dequeue!(s::Queue)
+
+Removes an element from the front of the queue `s`.
```
nalimilan: "and returns it"

src/tokens2.jl
```diff
@@ -66,7 +66,15 @@ end
          ii[2].address == 1?                       2 :
          ii[2].address == 2?                       3 : 1
 
-
+"""
+    compare(m::SAContainer, s::IntSemiToken, t::IntSemiToken)
+
+Here, `s`  and `t`  are semitokens  for the  same container  `m`; this
```
nalimilan: `Here, `s`  and `t`  are semitokens  for the  same container  `m`;` should come at the end (and "here" isn't useful).

nalimilan:
Thanks, this looks good to me after the handful of small points I commented on are fixed. Squashing is a good idea too.

abhijithch:
Will update based on @nalimilan 's comments. Thanks a lot for your guidance.

PR #186, opened 2016-03-24
Fix a bug and add a few tests

kshyatt:
Pretty sure that `pop!` wants a `ClassifiedCollections` there.


rawls238:
lgtm, nice catch

PR #190, opened 2016-03-31
fix deprecated syntax

KristofferC:
Else this package fails on 0.5.


KristofferC:
Looks like there is a few more problems, probably methods getting overwritten. This at least causes it parsing + precompilation to work.

DanielArndt:
Is this actually deprecated syntax? Did they change the rules around whitespace? Either way, it's cleaner.

And yes, the tests are currently broken. Need to find some time to look at that.

KristofferC:
Yes, it has been deprecated since 0.4. I noticed it now because on master they turned the deprecation into an error: https://github.com/JuliaLang/julia/commit/dacbdd453ca043d57b2812664adcd50630e205e6

mlubin:
Could we have a tagged release with this fix? We're trying to make JuMP run cleanly on 0.5. Thanks!

CC @joehuchette

PR #192, opened 2016-03-31
Add the codecov badge

kshyatt:
We are submitting coverage stats to Codecov, and many people like its interface better.


DanielArndt:
I'm not sure I love having both badges, although I understand it's a minor preference thing. It just looks cluttered already. 

I'm in favour of removing coveralls for Codecov, if that's the consensus? I see the main Julia repo shows both though, so if I'm the only one this bothers I'll live.

PR #197, opened 2016-04-08
WIP: Athul/redblacktree

athulappadan:
This is a simple implementation of red black tree. Looking forward for comments.

### DIFF
--- a/src/redBlackTree.jl
+++ b/src/redBlackTree.jl
@@ -1,0 +1,111 @@
+import Base: haskey, getindex, setindex!
+
+export Tree, EmptyTree, RBNode, RedBlackTree
+
+abstract Tree{K,V}
+
+type EmptyTree{K,V} <: Tree{K,V}
+end
+
+type RBNode{K,V} <: Tree{K,V}
+    key :: K
+    data :: V
+    left :: Tree{K,V}
+    right :: Tree{K,V}
+    red :: Bool
+end
+
+type RedBlackTree{K,V}
+    root :: Tree{K,V}
+
+    RedBlackTree() = new(EmptyTree{K,V}())
+end
+
+isRed(t::EmptyTree) = false
+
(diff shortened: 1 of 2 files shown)


src/redBlackTree.jl
```diff
@@ -0,0 +1,111 @@
...
+abstract Tree{K,V}
+
+type EmptyTree{K,V} <: Tree{K,V}
+end
+
+type RBNode{K,V} <: Tree{K,V}
+    key :: K
+    data :: V
+    left :: Tree{K,V}
+    right :: Tree{K,V}
```
kmsquire: This is probably the most straightforward implementation.  Performance-wise, however, in Julia, it's a [bad idea to have abstract members of types](http://docs.julialang.org/en/release-0.4/manual/performance-tips/#avoid-fields-with-abstract-type).

A workaround: https://github.com/JuliaLang/julia/issues/15791#issuecomment-207060497

(You'll probably have to read the whole thread for context.)

kmsquire:
Seems like a reasonable start.  It would be good to

1. add some tests, to show that it works.  Probably also need to be able to display the tree somehow.
2. look at the tests for AVL trees, and compare the performance.

athulappadan:
Thanks for the suggestions. Will commit the changes soon.

athulappadan:
I have been going through the docs and different packages to understand how to efficiently define types, but still haven't got a grip on it. Main doubts are regarding abstract types, defining types recursively, parametric types, performance etc. Can anybody suggest some materials, papers to get some clarity on the julia type system..?

timholy:
There are several good books and other learning resources, including those mentioned [here](http://julialang.org/learning/).

kmsquire:
@athulappadan, I'm going to close this, but feel free to ping back here if you work on this some more, and someone can reopen it.  Or, you can just open another pull request.

PR #198, opened 2016-04-10
most_common method for counters

ymer:
Return a list of the most common elements in a counter and their counts, ordered from the most common to the least.

most_common(counter([1, 6, 2, 2, 5, 5, 5]))

4-element Array{Pair{Int64,Int64},1}:
5=>3
2=>2
6=>1
1=>1


doc/source/accumulators.rst
```diff
@@ -34,11 +34,13 @@ Usage of an accumulator/counter::
...
+  push!(a, x)      # add the value/count for x by 1
+  push!(a, x, v)   # add the value/count for x by v
+  push!(a, a2)     # add all counts from a2 to a1
 
   pop!(a, x)       # remove a key x from a, and returns its current value
 
   merge(a, a2)     # return a new accumulator/counter that combines the
                    # values/counts in both a and a2
+
+most_common(a)     # Return a list of the most common elements and their counts from the most common to the least.
```
rawls238: I think fixing the indentation here should get rid of the weirdness in the rich diff

src/accumulator.jl
```diff
@@ -41,6 +41,8 @@ values(ct::Accumulator) = values(ct.map)
 
 sum(ct::Accumulator) = sum(values(ct.map))
 
+most_common(ct::Accumulator) = sort(collect(ct.map), by=p->p[2], rev=true)
```
rawls238: this is fine and seems like it works but I wonder if we could do this better especially for larger objects. Also this isn't really picking the "most common" elements but more just sorting the elements in order of occurrence. 

I wonder if it would be more useful and potentially faster if instead we had both this function definition plus a function definition that took in a number of elements to return so that you could get the k most common elements. I feel like this would be more useful and we could use order statistics to implement a solution with `O(n)` runtime instead of an `O(nlogn)` solution. Any thoughts on this?

rawls238: http://stackoverflow.com/questions/5380568/algorithm-to-find-k-smallest-numbers-in-array-of-n-items looks like it may have some implementations that are `O(n)` but I haven't looked at them too carefully to verify correctness.

kmsquire:
For parity with Julia Base, it might be nice to create a version of `select` which does this.  See http://docs.julialang.org/en/release-0.4/stdlib/sort/#Base.select! (and http://docs.julialang.org/en/release-0.4/stdlib/sort/#Base.select).  Or use `select` to define `most_common`, if that makes sense.

There are related functions, `nsmallest` and `nlargest`, which are implemented for heaps in this package as well.

ymer:
Could the heap nlargest be extended to take a "by" keyword?

kmsquire:
Well, it could.  This might not be a good idea if performance is an issue (see http://docs.julialang.org/en/release-0.4/manual/performance-tips/#declare-types-of-keyword-arguments, although there isn't much explanation as to why keywords can be slow).

I'm curious what your use case would be, since `nlargest` would already choose the largest values according to the value, and that's what you're proposing here.

Another idea, which I like a little better, would be to implement a separate `HeapCounter` type, which created a counter based on heaps instead of dictionaries.

ymer:
I think that the users would often want all the values in order, rather than the n most common. And most_common could be somewhat more suitable for this than nlargest.

Here is another suggestion:

`most_common(ct) = sort(collect(ct.map), by=p->p[2], rev=true)`

`most_common(ct, n) = select(collect(ct.map), 1:n, by=p->p[2], rev=true) `

oxinabox:
It may also be worth special casing `most_common(ct, 1)` which I think is another common use-case.
`select` is really slow compared to other methods for finding the single most common. 

Also good idea. I didn't even know this is missing. This is one of the main uses of the Counter type.

kmsquire:
Bump.

kmsquire:
This seems to have been satisfied by #375 (using `nlargest` and `nsmallest`).

PR #202, opened 2016-05-14
RFC: Fix getfield, String deprecations for v0.5

kmsquire:
In Julia v0.4 and earlier, `a.(x)` is shorthand for `getfield(a, x)`.
In Julia v0.5, this is deprecated, so use the function call.

Also fixes (or will fix)  `ASCIIString`, `UTF8String` -> `String` deprecations.


src/delegate.jl
```diff
@@ -9,7 +9,7 @@ macro delegate(source, targets)
         funcname = esc(funcnames[i])
         fdefs[i] = quote
                      ($funcname)(a::($typename), args...) =
-                       ($funcname)(a.($fieldname), args...)
```
yuyichao: See https://github.com/JuliaGraphics/Colors.jl/pull/247/files , I think this is not necessary (you can remove the `esc(Expr(:quote))` altogether)

test/test_accumulator.jl
```diff
@@ -43,14 +43,14 @@ push!(ct, ct2)
 @test ct["c"] == 2
 
 ct3 = counter(Dict([("a",10), ("b",20)]))
-@test isa(ct3, Accumulator{ASCIIString,Int})
```
yuyichao: I believe you want `Compat.ASCIIString`

src/DataStructures.jl
```diff
@@ -2,6 +2,8 @@ __precompile__()
 
 module DataStructures
 
+    import Compat: String
```
yuyichao: This needs to be added to the test instead.

test/test_accumulator.jl
kmsquire: I can do that, but is there a good reason we should prolong the use of `ASCIIString`?

REQUIRE
```diff
@@ -1 +1,2 @@
 julia 0.4
+Compat 0.7.16
```
yuyichao: Can be in `test/REQUIRE` instead?

test/test_accumulator.jl
yuyichao: The test will fail on 0.4.

src/DataStructures.jl
kmsquire: Err... right.

test/test_list.jl
```diff
@@ -44,7 +44,7 @@ l5 = map((x) -> x*2, l4)
 @test collect(l5) == [2; 4; 6]
 
 l5b = map((x) -> "$x", l5)
-@test typeof(l5b) == Cons{ASCIIString}
+@test typeof(l5b) == Cons{String}
```
yuyichao: `Compat.ASCIIString` and `isa`

kmsquire:
Thanks for the review, @yuyichao!

Obviously, I didn't do the `String` stuff right, because the tests fail.

That said, I don't understand why the tests should continue to use `ASCIIString`.  Isn't the goal to remove that entirely in lieu of `String` in v0.5?

yuyichao:
> Isn't the goal to remove that entirely in lieu of `String` in v0.5?

Yes, but that won't magically make `ASCIIString` (and especially containers that are deducted to have such a type parameter) disappear on 0.4.

Most of the tests are testing the type of the container for constructors using automatic type deduction so the result will still have `ASCIIString` or `UTF8String` in them. In another word, many test are effectively testing,

```
@test isa(["1"], Vector{ASCIIString})
```

And no magic on 0.5 can make this pass if you replace `ASCIIString` with `ByteString` on 0.4.

yuyichao:
I went ahead and fixes the comment I added above and replaces most of the typeof in the test with `isa`. I hope you are fine with this.

kmsquire:
:+1:

Thanks!  I had a burst of time available this morning, but I've used that up (probably along with tkelman's patience!).

If you have the time and willingness, please feel free to fix the first commit as well. I'm not sure when I'll get to it.

kmsquire:
Actually, since the tests pass fine without deprecations, that can be done separately.

I'll leave this open for other comments.  If there aren't any, this should be mergeable.

yuyichao:
> fix the first commit

Already done.

yuyichao:
And I want to have a version of this now so that I can do https://github.com/ihnorton/Clang.jl/pull/164 without depwarn from here =p

PR #208, opened 2016-07-26
Pull request/15e18a0b

phaverty:
While working on using OrderedDict within NamedArrays, I found that the creation of new OrderedDicts was one of the slower parts of NamedArray subsetting. This PR creates an inner constructor for making a new, empty OrderedDict of a specified size. As this new OrderedDict is known to be empty and of the correct size, it can be filled in a simplified fashion. This is good for a 1.6X speedup for creating an OrderedDict with 1e3 key/value pairs.

The inner constructor for an iterable of Pairs could be modified to use the same trick.


tkelman:
I wonder why Travis CI didn't run here. Pardon the close-reopen noise.

tkelman:
Should the new constructor have a unit test if it's intended to be used more broadly?

phaverty:
Certainly. I have now added the tests. The new constructor is really for
internal use only as it creates an OrderedDict that is not in a usable
state. Perhaps the two lines of this constructor should just be moved into
the inner constructors for multiple key/value pairs or Pairs.

My git/Pkg.* skills failed me and my full set of changes ended up as pull
request #210, which supersedes #207 and #208. Sorry for being untidy.

Pete

____________________
Peter M. Haverty, Ph.D.
Genentech, Inc.
phaverty@gene.com

tkelman:
Probably best to close the obsolete ones.

PR #210, opened 2016-07-26
While working on using OrderedDict within NamedArrays, I found that t…

phaverty:
…he creation of new OrderedDicts was one of the slower parts of NamedArray subsetting. This PR creates an inner constructor for making a new, empty OrderedDict of a specified size. As this new OrderedDict is known to be empty and of the correct size, it can be filled in a simplified fashion. This is good for a 1.6X speedup for creating an OrderedDict with 1e3 key/value pairs. Also, added tests for new constructor of empty-ish OrderedDict of a specified size.


phaverty:
Supersedes #207 and #208

PR #211, opened 2016-07-26
Filling a new OrderedDict quickly

phaverty:
While working on using OrderedDict within NamedArrays, I found that the creation of new OrderedDicts was one of the slower parts of NamedArray subsetting. This PR creates an inner constructor for making a new, empty OrderedDict of a specified size. As this new OrderedDict is known to be empty and of the correct size, it can be filled in a simplified fashion. This is good for a 1.6X speedup for creating an OrderedDict with 1e3 key/value pairs.

I have added tests for the new constructor.


phaverty:
Closed and re-opened to get CI to run. Sorry for the noise. This version will be the keeper.

phaverty:
CI failure is on nightly only and is due to absence of Base.complement in 0.5.

phaverty:
CI passes on both 0.5 and 0.4 now after the "complement" fix.

src/ordered_dict.jl
```diff
@@ -24,19 +24,34 @@ type OrderedDict{K,V} <: Associative{K,V}
     function OrderedDict()
         new(zeros(Int32,16), Array(K,0), Array(V,0), 0, false)
     end
+    function OrderedDict(n::Integer)
+        slotsz = max(16, (n*3)>>1 )
+        new(zeros(Int32,slotsz), Array(K,n), Array(V,n), 0, false)
+    end
     function OrderedDict(kv)
-        h = OrderedDict{K,V}()
+        h = OrderedDict{K,V}(length(kv))
```
Keno: This introduces an assumption that `kv` supports `length`.

phaverty: Thanks for having a look. I assumed that there had to be a 'length' method
for 'kv' because it was used in a for loop

for (k,v) in kv

, but I guess 'for' only requires the three functions necessary for an
iterator.  Do I want/need to do some kind of Traits thing to dispatch on
whether or not 'kv' has 'length' and then fall back to the old version if
it does not?

While I've got you on the line, do you have any thoughts about having an
inner constructor that makes an empty OrderedDict that can only be used in
the context of these filling functions? Perhaps it would be better to
copy/paste

slotsz = max(16, (n*3)>>1 )
h = new(zeros(Int32,slotsz), Array(K,n), Array(V,n), 0, false)

into the two relevant places.


Pete

____________________
Peter M. Haverty, Ph.D.
Genentech, Inc.
phaverty@gene.com

KristofferC: There already is `Base.HasLength()`: https://github.com/JuliaLang/julia/blob/bcc2121dc31647fc0999d09c10df91d35f216838/base/generator.jl#L44

phaverty: HasLength is a great idea, thanks. However, I can't see a way to use that
for 0.4.  Are infinite iterators really that common? Perhaps it is better
to fail for rare iterators than to require 0.5?

Pete

____________________
Peter M. Haverty, Ph.D.
Genentech, Inc.
phaverty@gene.com

Keno: Can't we just retain the old version for 0.4?

phaverty:
Now that 0.5 is out, is there a path forward for this PR? It does use one item that is specific to 0.5.

PR #215, opened 2016-09-09
Add a Deque based on a circular buffer

timholy:
The main attraction of `CircularDeque` is its performance:
```julia
julia> using DataStructures, BenchmarkTools

julia> D = Deque{Int}()
Deque [Int64[]]

julia> C = CircularDeque{Int}(5)
CircularDeque{Int64}([])

julia> function foo!(D)
           for i = 1:5
               push!(D, i)
           end
           v = back(D)
           for i = 1:5
               shift!(D)
           end
           v
       end
foo! (generic function with 1 method)

julia> foo!(D)
5

julia> foo!(C)
5

julia> @benchmark foo!($D)
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     870
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  0.00 bytes
  allocs estimate:  0
  minimum time:     153.00 ns (0.00% GC)
  median time:      155.00 ns (0.00% GC)
  mean time:        156.19 ns (0.00% GC)
  maximum time:     265.00 ns (0.00% GC)

julia> @benchmark foo!($C)
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     995
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  0.00 bytes
  allocs estimate:  0
  minimum time:     30.00 ns (0.00% GC)
  median time:      33.00 ns (0.00% GC)
  mean time:        33.22 ns (0.00% GC)
  maximum time:     56.00 ns (0.00% GC)
```

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.4
+Compat 0.8.5
--- a/test/test_circular_buffer.jl
+++ b/test/test_circular_buffer.jl
@@ -1,6 +1,10 @@
 using DataStructures
-using Base.Test
-
+if VERSION >= v"0.5.0-dev+7720"
+    using Base.Test
+else
+    using BaseTestNext
+    const Test = BaseTestNext
+end
 
 cb = CircularBuffer{Int}(5)
 @test length(cb) == 0
--- a/doc/source/circ_deque.rst
+++ b/doc/source/circ_deque.rst
@@ -1,0 +1,24 @@
+.. _ref-circ-deque:
+
+-------------
+CircularDeque
+-------------
+
+The ``CircularDeque`` type implements a double-ended queue using a circular buffer of fixed ``capacity``. This data structure supports constant-time insertion/removal of elements at both ends of a sequence.
+
+Usage::
+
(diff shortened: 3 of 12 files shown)


REQUIRE
```diff
@@ -1 +1,3 @@
 julia 0.4
+Compat 0.8.5
+BaseTestNext
```
tkelman: only a test dependency, isn't it?

useful addition though

timholy:
With a couple more performance optimizations, I got it down to a median of 20ns, which is almost 8 times faster than the existing `Deque`.

test/test_circular_buffer.jl
```diff
@@ -1,6 +1,10 @@
 using DataStructures
-using Base.Test
-
+if VERSION >= v"0.5.0-dev+7720"
```
rawls238: Isn't it sufficient to just do this in the runtests.jl file?

timholy: It seems to have been set up so that you can also run at least some of the test files independently. For example, if I'm hacking on one particular structure, it's nice to be able to quickly run the tests for just that structure.

PR #221, opened 2016-10-29
Faster creation of OrderedDict, revisited

phaverty:
This PR enables much faster creation of OrderedDict objects in julia 0.5 and is backwards compatible with 0.4.

### DIFF
--- a/src/ordered_dict.jl
+++ b/src/ordered_dict.jl
@@ -2,14 +2,18 @@
 # OrderedDict
 
 import Base: haskey, get, get!, getkey, delete!, push!, pop!, empty!,
-             setindex!, getindex, length, isempty, start,
-             next, done, keys, values, setdiff, setdiff!,
-             union, union!, intersect, filter, filter!,
-             hash, eltype, KeyIterator, ValueIterator, convert, copy,
-             merge, HasLength, HasShape, SizeUnknown, iteratorsize
+setindex!, getindex, length, isempty, start,
+next, done, keys, values, setdiff, setdiff!,
+union, union!, intersect, filter, filter!,
+hash, eltype, KeyIterator, ValueIterator, convert, copy,
+merge
+
+if VERSION >= v"0.5.0"
+    import Base: HasLength, HasShape, SizeUnknown, iteratorsize
+end
 
 """
-    OrderedDict
+OrderedDict
 
 `OrderedDict`s are  simply dictionaries  whose entries  have a  particular order.  The order
 refers to insertion order, which allows deterministic iteration over the dictionary or set.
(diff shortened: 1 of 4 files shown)


src/ordered_dict.jl
```diff
@@ -21,25 +25,34 @@ type OrderedDict{K,V} <: Associative{K,V}
...
     dirty::Bool
 
-    function OrderedDict()
-        new(zeros(Int32,16), Array(K,0), Array(V,0), 0, false)
+    OrderedDict() = new(zeros(Int32,16), Array(K,0), Array(V,0), 0, false)
+
+    @static if VERSION >= v"0.5.0"
+        function OrderedDict(kv, isz::Union{HasLength, HasShape})
+            n = length(kv)
+            slotsz = max(16, (n*3)>>1 )
```
kmsquire: This PR could be a lot smaller if you did something like the following:

```julia
    @static if VERSION >= v"0.5.0"
        get_n_slotsz(kv) = get_slot_size(kv, iteratorsize(kv))
        get_n_slotsz(kv, isz::SizeUnknown) = (0,16)
        function get_n_slotsz(kv, isz::Union{HasLength, HasShape})
            n = length(kv)
            slotsz = max(16, (n*3)>>1)
            return n, slotsz
        end
    else
        get_n_slotsz(kv) = (0,16)
    end
```

src/ordered_dict.jl
```diff
@@ -21,25 +25,34 @@ type OrderedDict{K,V} <: Associative{K,V}
...
+            h = new(zeros(Int32,slotsz), Array(K,n), Array(V,n), 0, false)
+            i = 0
+            for (k,v) in kv
+                i = i + 1
+                index = hashindex(k,slotsz)
+                h.keys[i] = k
+                h.vals[i] = v
+                h.slots[index] = i
```
kmsquire: This won't work, as it doesn't handle index collisions.  Use `ht_keyindex2`.

src/ordered_dict.jl
```diff
@@ -84,9 +83,44 @@ function OrderedDict(kv)
...
+    OrderedDict{K,V}(kv::Associative{K,V})          = OrderedDict{K,V}(kv, iteratorsize(kv))
+
+    OrderedDict{K,V}(ps::Pair{K,V}...)          = OrderedDict{K,V}(ps, iteratorsize(ps))
+    OrderedDict{K}(ps::Pair{K}...,)             = OrderedDict{K,Any}(ps, iteratorsize(ps))
+    OrderedDict{V}(ps::Pair{TypeVar(:K),V}...,) = OrderedDict{Any,V}(ps, iteratorsize(ps))
+    OrderedDict(ps::Pair...)                    = OrderedDict{Any,Any}(ps, iteratorsize(ps))
+
+    dict_with_eltype{K,V}(kv, ::Type{Tuple{K,V}}) = OrderedDict{K,V}(kv, iteratorsize(kv))
+    dict_with_eltype{K,V}(kv, ::Type{Pair{K,V}}) = OrderedDict{K,V}(kv, iteratorsize(kv))
+    dict_with_eltype(kv, t) = OrderedDict{Any,Any}(kv, iteratorsize(kv))
```
kmsquire: This block can be removed if you define `get_n_slotsz` above.  (There is probably a better name for that function...)

phaverty:
Thanks for the thoughtful review @kmsquire .  I think I've implemented all of your suggestions.

src/ordered_dict.jl
```diff
@@ -21,25 +25,28 @@ type OrderedDict{K,V} <: Associative{K,V}
...
+        (n, slotsz) = get_n_slotsz(kv)
+        h = new(zeros(Int32,slotsz), Array(K,n), Array(V,n), 0, false)
+        if n > 0
+            i = 0
+            for (k,v) in kv
+                i = i + 1
+                index = ht_keyindex2(h, k)
+                @inbounds h.keys[i] = k
+                @inbounds h.vals[i] = v
+                @inbounds h.slots[-index] = i
```
kmsquire: Closer, but there needs to be a little more defensive programming here.  Since `kv` is an arbitrary set of key-value pairs, it's possible that a key is repeated, so you'll need to check the return value of `index` and respond accordingly.  You should double check against `setindex!` to make sure there aren't any other corner cases.

phaverty:
This new version handles duplicated keys in input to the constructor. I spent some time with setindex! and friends and I think this covers it. Any necessary rehashing should happen in ht_keyindex2 and we don't have to convert the incoming keys and values to the right type as we've created the OrderedDict to match the incoming K and V types.

phaverty:
I think this should do it. I spent some time with setindex! and friends. We don't have to worry about converting the types of the incoming keys and values as we have created the new OrderedDict to match those K and V types. ht_keyindex2 should handle any necessary rehashing. This commit handles incoming duplicated keys.  I think that's it. Please let me know if you can think of anything else.

src/ordered_dict.jl
```diff
@@ -21,25 +25,35 @@ type OrderedDict{K,V} <: Associative{K,V}
...
+                end
+            end
+        else # Unknown length
+            for (k,v) in kv
+                h[k] = v
+            end
         end
         return h
     end
+
```
kmsquire: Minor nitpick: spaces at end of line.

phaverty:
I've pushed a commit that cleans up the extra spaces.

Pete

____________________
Peter M. Haverty, Ph.D.
Genentech, Inc.
phaverty@gene.com

kmsquire:
Overall, I think this is good.  

I'm slightly concerned that `OrderedDicts` (and this PR) might be susceptible to https://github.com/JuliaLang/julia/issues/15077, but the implementation has changed enough since then that it's hard to test (we pass the test in `Base`, but the arrays are not sized the same as described [here](https://github.com/JuliaLang/julia/issues/15077), and I believe if it were sized the same, the test would fail, since we don't have the appropriate checks in place.  Cc: @StefanKarpinski 

That said, that's no worse than the current situation, and it should be rare.  I'll try to test a little later (but if you have time and want to take a poke at it, please do).  Either way, I plan to merge this evening (PST) if there are no other comments, and if a fix is needed for the above, we can fix this constructor at the same time.

phaverty:
This new constructor algo works for very short OrderedDicts.  Let me think about this one a bit more.

PR #224, opened 2016-11-14
RFC: Remove HashDict

kmsquire:
* `HashDict` was originally copied from Base and used as a basis for both
  `Ordered` and `Unordered` dictionaries.  However, Jeff Bezanson's
  `OrderedDict` is implemented slightly differently and is committed here, so
  there's no need for this type anymore.


kmsquire:
Julia nightly is timing out.  All other builds pass.

kmsquire:
I split out the `DefaultDict` changes into #233, so waiting until that is merged, and will rebase.

tkelman:
Actually this is what broke Gadfly. I'm going to revert this.

kmsquire:
Wondering why Gadfly is using HashDict...

tkelman:
It's probably not using it directly, but it is doing something with `push!` into one of the other special Dict types that used to call into this I think. Perhaps try deprecating it instead, see if that can be done in a way that doesn't break Gadfly then work with Gadfly maintainers to move them to something else going forward?

kmsquire:
Ok, I'll take a closer look when I can get in front of a computer (probably next weekend).

Issue #223, opened 2016-11-12
N-dimensional (pre-allocating) circular buffer


femtotrader:
Hello,

I noticed that current implementation of circular buffer is only 1 dimension.

A generalized approach of this problem will be nice (especially for [implementing streaming timeseries](https://github.com/femtotrader/TimeSeriesIO.jl/issues/5))

So, instead of pushing a single value in a 1D circular buffer,
- it will be possible to push an `Array{T,1}` in a 2D circular buffer
- it will be possible to push an `Array{T,2}` in a 3D circular buffer
- it will be possible to push an `Array{T,N-1}` in a N-D circular buffer

I also noticed that current implementation of circular buffer is using a Vector with size increasing (until capacity is reached). Maybe pre-allocating array will be a more efficient approach ?

@tbreloff I noticed you are original author.
What is your opinion about this ?

femtotrader:
WIP:

```julia

"""
New items are pushed to the back of the list, overwriting values in a circular fashion.
"""
type CircularBuffer{T, dims} <: AbstractArray{T, dims}
    capacity::Int
    first::Int
    buffer::Array{T, dims}
    used::Int
    colons::Array{Colon}

    function CircularBuffer(capacity::Int, others_dims...)
        a = Array{T, dims}(capacity, others_dims...)
        colons = Array{Colon}(others_dims)
        new(capacity, 1, a, 0, colons)
    end
end

function CircularBuffer(T, capacity::Int, others_dims...)
    n_dims = length(others_dims) + 1
    CircularBuffer{T,n_dims}(capacity::Int, others_dims...)
end

function _buffer_index(cb::CircularBuffer, i::Int)
    n = length(cb)
    if i < 1 || i > n
        throw(BoundsError("CircularBuffer out of range. cb=$cb i=$i"))
    end
    idx = cb.first + i - 1
    if idx > n
        idx - n
    else
        idx
    end
end

function Base.getindex(cb::CircularBuffer, i::Int)
    if ndims(cb) < 2
        cb.buffer[_buffer_index(cb, i)]    
    else
        cb.buffer[_buffer_index(cb, i), cb.colons...]
    end
end

function Base.setindex!(cb::CircularBuffer, data, i::Int)
    cb.buffer[_buffer_index(cb, i), cb.colons...] = data
    cb
end

function Base.push!(cb::CircularBuffer, data)
    # if full, increment and overwrite, otherwise push
    if length(cb) == cb.capacity
[...]

PR #228, opened 2016-11-15
[WIP] N dim circular buffer

femtotrader:
Related issue #223

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -1,13 +1,23 @@
-
 """
 New items are pushed to the back of the list, overwriting values in a circular fashion.
 """
-type CircularBuffer{T} <: AbstractVector{T}
+type CircularBuffer{T, N} <: AbstractArray{T, N}
     capacity::Int
     first::Int
-    buffer::Vector{T}
+    buffer::Array{T, N}
+    used::Int
+    colons::Vector{Colon}
 
-    CircularBuffer(capacity::Int) = new(capacity, 1, T[])
+    function CircularBuffer(capacity::Int, others_dims...)
+        a = Array{T, N}(capacity, others_dims...)
+        colons = Vector{Colon}(N - 1)
+        new(capacity, 1, a, 0, colons)
+    end
+end
+
+function CircularBuffer(T, capacity::Int, others_dims...)
+    N = length(others_dims) + 1
+    CircularBuffer{T,N}(capacity::Int, others_dims...)
 end
 
 function _buffer_index(cb::CircularBuffer, i::Int)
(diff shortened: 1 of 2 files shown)


src/circular_buffer.jl
```diff
@@ -1,44 +1,50 @@
...
-        idx - n
-    else
-        idx
-    end
+    idx = mod1(cb.first + i - 1, n)
 end
 
 function Base.getindex(cb::CircularBuffer, i::Int)
-    cb.buffer[_buffer_index(cb, i)]
+    cb.buffer[_buffer_index(cb, i), cb.colons...]
```
tbreloff: This seems suboptimal. If you're set on going down the Multidim path, maybe this should be a generated function and you can remove the colons field?

src/circular_buffer.jl
```diff
@@ -1,44 +1,50 @@
...
+type CircularBuffer{T, N} <: AbstractArray{T, N}
     capacity::Int
     first::Int
-    buffer::Vector{T}
+    buffer::Array{T, N}
+    used::Int
+    colons::Vector{Colon}
+
+    function CircularBuffer(capacity::Int, others_dims...)
+        a = Array{T, N}(capacity, others_dims...)
```
tbreloff: For context, I didn't do this originally because the capacity might have been much larger than the typical size. Allocating all of the capacity is only better when you actually fill the buffer.

src/circular_buffer.jl
femtotrader: Probably. Some help will be great.

src/circular_buffer.jl
```diff
@@ -1,44 +1,50 @@
...
 end
 
 function Base.getindex(cb::CircularBuffer, i::Int)
-    cb.buffer[_buffer_index(cb, i)]
+    cb.buffer[_buffer_index(cb, i), cb.colons...]
 end
 
 function Base.setindex!(cb::CircularBuffer, data, i::Int)
-    cb.buffer[_buffer_index(cb, i)] = data
+    cb.buffer[_buffer_index(cb, i), cb.colons...] = data
```
femtotrader: I tried to use this code in https://github.com/femtotrader/TimeSeriesIO.jl for a live/stream time series resampler.
see https://github.com/femtotrader/TimeSeriesIO.jl/blob/master/test/test_stream_time_array_ohlcv.jl
and noticed that `setindex!` might be fixed like this:

```julia
function Base.setindex!(cb::CircularBuffer, data, i::Int, args...)
    if length(args) == 0
        colons = cb.colons
    else
        colons = cb.colons[1:length(args)]        
    end
    cb.buffer[_buffer_index(cb, i), args..., colons...] = data
    cb
end
```

Here is my test

```julia
cb = NDCircularBuffer(Int, 3, 4)
#println(convert(Array, cb))
@test size(cb) == (0, 4)
println(cb)
push!(cb, [10, 11, 12, 13])
@test size(cb) == (1, 4)
push!(cb, [20, 21, 22, 23])
@test size(cb) == (2, 4)
push!(cb, [30, 31, 32, 33])
@test size(cb) == (3, 4)
push!(cb, [40, 41, 42, 43])
push!(cb, [50, 51, 52, 53])
@test cb[end-2] == [30,31,32,33]
@test cb[end-1] == [40,41,42,43]
@test cb[end] == [50,51,52,53]
println(cb.buffer)

cb[end, 2] = 99
@test cb[end] == [50,99,52,53]
```

Without this tiny fix `cb[end]` was `[50,51,52,53]` so `setindex!` wasn't working correctly

PR #231, opened 2016-11-20
Make Accumulator an immutable subtype of Associative

oxinabox:
Accumulator can be immutable -- the contents of the `map` changes, the `map` never does.
`immutable` should be preferred to `type` where possible, as it is more performant.

And accumulator should be a subtype of Associative -- because it is.
This lets it be used as a dispatch.

### DIFF
--- a/src/accumulator.jl
+++ b/src/accumulator.jl
@@ -1,6 +1,6 @@
 # A counter type
 
-type Accumulator{T, V<:Number}
+immutable Accumulator{T, V<:Number} <: Associative{K,V}
     map::Dict{T,V}
 end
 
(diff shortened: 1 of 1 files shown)


src/accumulator.jl
```diff
@@ -1,6 +1,6 @@
 # A counter type
 
-type Accumulator{T, V<:Number}
+immutable Accumulator{T, V<:Number} <: Associative{K,V}
```
kmsquire: `K` => `T` (or vice versa)

tkelman:
Making this a subtype of Associative introduces an ambiguity which managed to break Gadfly...

tkelman:
There is an ambiguity, but I think what broke Gadfly was something else.

Issue #235, opened 2016-12-05
Pushing to heap does not return heap


tawheeler:
I just noticed that calling `push!` on a binary heap does not return the modified heap, as is convention.

```
using DataStructures
h = binary_minheap(Int)
push!(h, 1)
```

kmsquire:
Should be an easy fix. Could you submit a PR?

ararslan:
Fixed by #236?

tawheeler:
Yessir

PR #236, opened 2016-12-05
push! on BinaryHeap now returns heap rather than nothing

tawheeler:
Fixes issue #235


src/heaps/binary_heap.jl
```diff
@@ -138,6 +138,7 @@ function push!{T}(h::BinaryHeap{T}, v::T)
     valtree = h.valtree
     push!(valtree, v)
     _heap_bubble_up!(h.comparer, valtree, length(valtree))
+    h
```
kmsquire: Convention in `Base` seems moving to using an explicit return for functions longer than one line, but we haven't moved to that convention here yet.  So, this seems fine to me (and we can use an explicit return when someone is up to updating most or all of the return values at once.)

PR #237, opened 2016-12-07
Add warning on unordered to ordered conversion. Fixes #216

### LINKED ISSUE #216: Should there be conversion from Dict to OrderedDict?
Since `convert(Vector{Int}, Set([1,2,3]))` is not defined, shouldn't `convert(OrderedDict{Int, Int}, Dict(1=>2, 3=>4))` also be undefined? `OrderedDict` cares about the ordering, but the conversion gives it a random one, which isn't meaningful.

cstjean:
I tried to match the tone from Base, but please let me know if anything sounds off.

### DIFF
--- a/src/default_dict.jl
+++ b/src/default_dict.jl
@@ -135,6 +135,7 @@ for _Dict in [:Dict, :OrderedDict]
     end
 end
 
+isordered{T<:DefaultOrderedDict}(::Type{T}) = true
 
 ## This should be uncommented to provide a DefaultSortedDict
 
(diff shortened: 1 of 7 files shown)


kmsquire:
@cstjean, sorry for the delay on this, but would you be able to add some tests?

kmsquire:
(Also, there is a minor conflict.)

cstjean:
Can I test the warnings? Or do you mean to make sure that `is_ordered(Dict)==true`?

kmsquire:
Just testing `is_ordered` (for all of the dictionaries for which it is defined) would be fine.  Base runs julia with `--depwarn=error`, which could be caught, but I don't think packages are tested that way by default.

cstjean:
Tested and rebased.

kmsquire:
Thanks!

kmsquire:
For testing the deprecation see https://github.com/JuliaLang/julia/issues/20348.

PR #238, opened 2016-12-15
Move PriorityQueue over from Base

ararslan:
Ref https://github.com/JuliaLang/julia/issues/18497

This PR copies the definitions, tests, and documentation from `Base.Collections.PriorityQueue` into DataStructures in preparation for their imminent removal from Base.

cc @StefanKarpinski


ararslan:
I've done something horribly wrong but I'm not sure what it is.

ararslan:
This works on 0.5 and nightly but not on 0.4. Any guidance for how to get this working on 0.4 would be much appreciated.

kmsquire:
One small issue (present elsewhere  DayaStructures.jl) is that we conflate basic data structures (heaps) with abstract data types (priority queues). It would be nice to disentangle these. 

Any chance to share code here with Heaps?

ararslan:
I'm sure that's possible but this is well outside my realm of knowledge. Perhaps this change is better left for someone who's more familiar with these things to take on.

vtjnash:
should open / move https://github.com/JuliaLang/julia/issues/4372 here when this merges

ararslan:
Okay, I've preserved all authorship information for the moved code, which is why this suddenly looks like an enormous mess.

I've preserved the code from Base which provides heap operations on arrays for convenience (other packages, e.g. QuadGK, use that). I hope that's okay.

StephenVavasis:
I would be against dropping 0.4 support at this time.  My own code has encountered a bug in 0.5.0 (significant performance loss for arrays with 6 or more subscripts).  This bug is patched in 0.6.0-dev, but the patch has not yet been backported.  So I am still using 0.4.  I suspect I am not alone in this regard.

ararslan:
Do you have any suggestions for how to make the code in this PR work on 0.4 then, @StephenVavasis?

StephenVavasis:
What does copymutable do?  I didn't find it in the 0.5.1 predocumentation.  Would deepcopy be a reasonable substitute?

ararslan:
From the current master docstring:
> Make a mutable copy of an array or iterable `a`. For `a::Array`, this is equivalent to `copy(a)`, but for other array types it may differ depending on the type of `similar(a)`. For generic iterables this is equivalent to `collect(a)`.

I'm not really sure what the closest analog would be in 0.4, but I guess `copy` or `deepcopy`, since in here it's only being used on `AbstractArray`s?

ararslan:
Huh, nice. I guess that did it! This actually works on 0.4 now! Thanks so much, @StephenVavasis!

ararslan:
If this is good to go, can someone merge this and tag the package at their earliest convenience so that https://github.com/JuliaLang/julia/issues/19800 can make it into Base before the 0.6 feature freeze?

ararslan:
As an aside, whoever merges this will probably want to ensure that they do not squash on merge, otherwise the history will not be preserved (AFAIK).

tkelman:
Can we get this in please? Even if it doesn't fit perfectly with the rest of the package's functionality, this is a much better place for it and it can be modified going forward.

tkelman:
I don't want to step on any toes of the people who do most of the maintenance work on this package, but we're trying to get base feature frozen for 0.6 so in a hurry here. If I don't hear a strong objection I'm likely to merge this.

Issue #219, opened 2016-10-15
Proposal: B-trees


alooow:
Hi! I was thinking about contributing to this package. I'm interested in B-trees. I saw they were also mentioned in previous issues and topics on julia-dev. Please let me know if you would like to have them in this package? I would be more than happy to help.

StephenVavasis:
I believe that the data structures SortedDict, SortedSet and SortedMultDict are currently losing performance because of cache-misses during tree-traversal operations.  (I did some limited testing on this issue two years ago and drew this conclusion.)  There is a proposal in the literature, Bender, Demaine and Farach-Colton, SIAM J. Comput. 35 (2006) 341-358, for a B-tree-like data structure that significantly reduces cache misses.  I have not had the time to implement this myself, but you might be interested in giving it a try.  The idea would be to build a drop-in replacement (i.e., same API) for the current source file balanced_tree.jl, which is based on 2-3 trees, and then see if there are performance gains for large SortedDict and SortedSet collections.

alooow:
All right, I'll take a look at it, thanks!

femtotrader:
I'm also interested in [B-tree](https://en.wikipedia.org/wiki/B-tree) ... more exactly in a particular form of [binary search tree](https://en.wikipedia.org/wiki/Binary_search_tree) called [AVL tree](https://en.wikipedia.org/wiki/AVL_tree) see #503

timholy:
I wonder if it would be better to put trees in a separate package(s). There are *so many* variants. ProfileView contains a left-child right-sibling tree implementation that should probably be split out; there is also https://github.com/Keno/AbstractTrees.jl

femtotrader:
Maybe @Keno should consider moving AbstractTrees to JuliaCollections?

o314:
Tree will still be a very large generic abstract container.
A *short* list of the kinship of tree :

- tree
- bst
- rbtree
- merkle tree
- ubtree
- btree
- b+tree
- b*tree
- bwtree
- rtree
- kdtree
- quadtree
- octtree
- bsptree
- trie
- lsmtree
- ...

`Graph`s have been made autonomous, `Array` kind of.
It make sense that trees are promoted too.

`AbstractTree` could be bring back to a `Trees` package

PR #241, opened 2016-12-27
WIP: Memory packed arrays

alooow:
I am working on the implementation of oblivious B-trees, as proposed in issue  #219, based on the paper _Cache-Oblivious B-Trees, Bender, Demaine and Farach-Colton, SIAM J. Comput. 35 (2006) 341-358_ 

  The structure is quite complicated, but on the bottom of everything lies a memory packed array (description of it can be found in the above paper, as well as in: _An adaptive packed-memory array, Bender, Hu, ACM Transactions on Database Systems v.32 i.4, (2007)_ ). For use in the B-trees the structure must offer an "insert after" method and a "delete" method. Presented implementation is still a subject to change. One of the problems is that the array must be always within lower and upper threshold. I tested the current implementation only with the lower threshold set to 0, as otherwise an empty array wouldn't  be within the threshold, preventing from inserting any element. I will try to add some test by the end of this or future week. As the array is a basis for future work on the cache-oblivious b-trees, I would really appreciate if someone could look at it and share his opinion.   

  I was also wondering if the memory packed array would be useful not only for the cache oblivious B-trees. I am not sure, what is right now the best way of inserting an element after a given index (I tried only insert!(collection, index, item)). I was thinking about making some tests and checking if memory packed array is a faster solution for this issue.

### DIFF
--- a/src/memory_packed_array.jl
+++ b/src/memory_packed_array.jl
@@ -65,7 +65,7 @@ function scan{D}(A::MemoryPackedArray{D},from::Int,to::Int)
 	local nr = 0;
 	local i = from;
 	local j = to-1;
-	while(i<j)
+	while i<j
 		if(A.exists[i])
 			nr = nr+1;
 		end
@@ -97,7 +97,7 @@ end
 function ancestor{D}(A::MemoryPackedArray{D},left::Int,right::Int,seg_pos::Int, seg_size::Int)
 	local nr::Int;
 	if (seg_pos & 1)==1 ## we are in left child;
-		nr = scan(A,right,right+seg_size+1);
+		nr = scan(A,right,right+seg_size);
 		right = right+seg_size;
 	else 
 		nr = scan(A,left-seg_size,left);
@@ -136,13 +136,11 @@ function rebalance{D}(A::MemoryPackedArray{D},left::Int,right::Int,number_of_ele
 		A.store[index] = elemnt;
 		A.exists[index] = true;
 	else
-		j = j-1; ## last last taken place
-		print(j," ",k)
+		j = j-1; ## last taken place
 		while i>=left && j>k
 			A.store[i] = A.store[j];
 			A.exists[i] = true;
 			j = j-1;
-			println(i-gap)
 			i = ceil(Int,i-gap);
 		end
 		if i>=left ## then j==k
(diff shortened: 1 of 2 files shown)


StephenVavasis:
This looks like a good start!  I'll mention a few stylistic points about Julia:

- the 'local' keyword is rarely needed
- statements generally do not need a semicolon terminator
- Usually j+=1 is preferred to j=j+1.
- The expression `sum(A.exists[from:(to-1)])` will needlessly create a temporary array.  You can either replace this with an explicit loop, or if you are targeting 0.6.0, then you can use a new generator expression to loop over the subarray. 
- it is probably best to avoid using `size` as a variable name since this is also the name of a function in Base.

With regard to the data structure itself: if you are planning to develop a drop-in replacement for the current 2-3-tree structure that underlies `SortedDict`, `SortedMultiDict` and `SortedSet`, then almost certainly you will need 'up' pointers at the leaves of the tree.  These are necessary to support the token interface to the data structures, and they are not mentioned in the Bender et al. paper quoted above.  As far as I can see, your code can already handle 'up' pointers if you designate D (the data type of the array) to be a tuple or small immutable one of whose fields is an 'up' pointer.

kmsquire:
Looking forward to this!

tjgreen:
Is this pull request still active?  I just stumbled upon it, after doing my own implementation of (adaptive) packed-memory arrays, which I was considering packaging up for candidacy to DataStructures.jl.  A second question I have for folks here is whether there are any efforts underway to create a standard library of paged data structures, such as the cache-oblivious B-trees mentioned here.  I don't think these would fit properly into DataStructures.jl, because they require a notion of page-access/buffer pool implemented by components TBD.

alooow:
I'm sorry, I had a lot going on last year and  I didn't have time give this issue proper attention.  @tjgreen,   I am new to cache oblivious algorithms (got involved through this implementation) so my implementation of MPA can still be very faulty. During the last commits I was still fixing the delete procedure (should be working now).  If you think your implementation works better, or if you have any suggestions I will be happy to hear them out.

kmsquire:
I'm going to close this for now, and suggest that if @alooow (or anyone else) wants to pursue this further, it can probably be done in a separate package.

PR #250, opened 2017-01-10
Implement start/next/done for CircularDeque

kdmurray91:
Hi,

This PR simply adds the `start`/`next`/`done` methods required for idiomatic for-loop iteration over CircularDeques.

I'm a relative Julia novice, so any tips (especially regarding performance) are appreciated.

Cheers,
Kevin

### DIFF
--- a/src/circ_deque.jl
+++ b/src/circ_deque.jl
@@ -70,20 +70,25 @@ end
     v
 end
 
-@inline function Base.getindex(D::CircularDeque, i::Integer)
-    @compat @boundscheck 1 <= i <= D.n || throw(BoundsError())
+# getindex sans bounds checking
+@inline function _unsafe_getindex(D::CircularDeque, i::Integer)
     j = D.first + i - 1
     if j > D.capacity
         j -= D.capacity
     end
     @inbounds ret = D.buffer[j]
-    ret
+    return ret
+end
+
+@inline function Base.getindex(D::CircularDeque, i::Integer)
+    @compat @boundscheck 1 <= i <= D.n || throw(BoundsError())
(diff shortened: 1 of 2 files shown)


test/test_circ_deque.jl
```diff
@@ -78,6 +78,11 @@ end
     empty!(D)
     unshift!(D, 40)
     @test front(D) == back(D) == 40
+
+    # Test iteration over loop
+    D = CircularDeque{Int}(5)
+    for i in 1:5 push!(D, i) end
+    @test collect(i for i in D) == collect(1:5)
```
kmsquire: Generators aren't available on Julia 0.4.  To maintain compatibility, you could enclose the generator inside an array:

```julia
collect([i for i in D])
```

src/circ_deque.jl
```diff
@@ -80,6 +80,11 @@ end
     ret
 end
 
+# Iteration via getindex
+Base.start(d::CircularDeque) = 1
+Base.next(d::CircularDeque, i) = (d[i], i+1)
+Base.done(d::CircularDeque, i) = i == length(d) + 1
```
kmsquire: Using `getindex` is a good first pass.  It does do a number of checks that might (?) not be necessary during iteration, so if you wanted, you could try to implement a more complicated version of these functions which access the buffer directly. Not sure if the actual benefit.

kdmurray91: I've pushed an "enhanced" version that uses a new `unsafe_getindex` to avoid bounds checking via getindex.

kmsquire:
Thanks for the contribution!

PR #251, opened 2017-01-11
Add Base.unshift! for CircularBuffers

kdmurray91:
Hi,

This PR adds `unshift!` to circular buffers. I've also `@testset`-ified the tests, and added tests of unshifting items.

On an unrelated note, I notice that the extended functions for `CircularBuffer` (push, getindex, setindex etc.) are not `@inline`. Is there any point in inlining these?

Cheers,
K

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -43,6 +43,17 @@ function Base.push!(cb::CircularBuffer, data)
     cb
 end
 
+function Base.unshift!(cb::CircularBuffer, data)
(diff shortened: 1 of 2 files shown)


kmsquire:
> On an unrelated note, I notice that the extended functions for `CircularBuffer` (push, getindex, setindex etc.) are not `@inline`. Is there any point in inlining these?

Not at a computer, so I'm lazy to check, but if the definitions are short enough, the compiler should automatically inline them. 

Best way to answer this would be to write some timing tests and see if there is a difference.

PR #252, opened 2017-01-15
RFC: revert "revert remove HashDict"

kmsquire:
This removes `HashDict` again, and fixes some issues uncovered by basing `DefaultDict`s on `Base.Dict` rather than `HashDict`.

Previously, `getindex` always called `get!(h::Associative, key, default)` to return a default value.

When DefaultDict was based on HashDict (now deleted) and the default value was calculated using a function, a version of this function with signature
```julia
get!(h::HashDict, key, default::Base.Callable)
```
was called.

`Base.Dict` has no such function (nor does `DataStructures.OrderedDict`), so the (incorrect) value returned was the function object which generated the default value, not the default value itself.

Additionally, there were no tests for `DefaultDicts` with defaults generated by a function.

* Fixed so that `getindex` calls the `do` function form of `get!` on `DefaultDict`s/`DefaultOrderedDict`s when appropriate
* Added tests for `DefaultDict`s with defaults generated by functions

**Edit:** Also, add back functionality for pushing tuples to `DefaultDict`s.


tkelman:
This fixes the Gadfly problem I hope?

PR #253, opened 2017-01-16
RFC: Fix delegation for *Dict functions returning the underlying *Dict

kmsquire:
* When called on an `Associative` container, `sizehint!`, `empty!`, and
  `setindex!` all return the container
* The `@delegate` macro is not aware of this, and instead returned
  the wrapped object when used to define, e.g., `DefaultDict`s
* Fixed by creating a second macro, `@delegate_return_parent`,
  which works like `@delegate`, but returns the parent container.

**Edit:** `delete!` also returns the container.


kmsquire:
I noticed this while working on #252.  Should be uncontroversial, but I'm open to suggestions for alternative fixes, if anyone has any suggestions.

PR #254, opened 2017-01-16
RFC: Deprecate functionality to push! tuples onto DefaultDict, DefaultOrderedDict

kmsquire:
* For `Base.Dict`, only `Pair`s can be pushed.
* To make `Default*Dict`s consistent, deprecate this functionality

This would affect Gadfly, and in particular, fixing the deprecations would make the code somewhat less readable. (see https://github.com/kmsquire/Gadfly.jl/commit/b885108046bc3fd7e451b2407bc92a56a91f0b64)

An alternative would be to add the ability to push tuples (and maybe other possible pair datatypes) to Base Julia.

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -109,7 +109,7 @@ module DataStructures
     @deprecate iter(s::Stack) s
     @deprecate iter(q::Queue) q
 
-    # Remove when Julia 0.7 (or whatever version is after v0.6) is released
+    # Remove when Julia 1.0 (or whatever version is after v0.6) is released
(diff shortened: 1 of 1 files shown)


tkelman:
I forget, was Gadfly the only thing that broke when these were accidentally removed?

Might be good to only add new deprecations after a last 0.4-compatible release is made and master goes 0.5-only, since many other packages have already dropped 0.4 support and wouldn't necessarily fix any newly introduced deprecations there.

kmsquire:
After thinking about this, my preference would actually be to try to add back support for creating Dicts from tuples to base.

PR #257, opened 2017-01-17
Julia 0.6 fixes

andreasnoack:
Is there a better way to handle the new `where` syntax than `eval(parse(...))`?

### DIFF
--- a/src/balanced_tree.jl
+++ b/src/balanced_tree.jl
@@ -125,15 +125,15 @@ type BalancedTree23{K, D, Ord <: Ordering}
     deletionchild::Array{Int,1}
     deletionleftkey::Array{K,1}
     function BalancedTree23(ord1::Ord)
-        tree1 = Array{TreeNode{K}}(1)
+        tree1 = Vector{TreeNode{K}}(1)
         initializeTree!(tree1)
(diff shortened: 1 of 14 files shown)


src/balanced_tree.jl
```diff
@@ -119,21 +119,21 @@ type BalancedTree23{K, D, Ord <: Ordering}
...
     freedatainds::Array{Int,1}
-    useddatacells::IntSet 
+    useddatacells::IntSet
     # The next two arrays are used as a workspace by the delete!
     # function.
     deletionchild::Array{Int,1}
     deletionleftkey::Array{K,1}
     function BalancedTree23(ord1::Ord)
-        tree1 = Array(TreeNode{K}, 1)
+        tree1 = Array{TreeNode{K}}(1)
```
tkelman: should use Vector when the dimension is know to be 1

andreasnoack: I don't really see why that is important. It is implied by the number of arguments.

tkelman: the compiler has to do more work if it's not explicit

andreasnoack:
Updated and rebased. I've also enabled testing on nightly to avoid that we break the package again.

PR #260, opened 2017-01-20
RFC: Remove eval from DefaultDict constructor

kmsquire:
* eval was being used to create the `DefaultDict` type, because it required some v0.6 specific changes.  
  However, with `@compat`, eval isn't needed (unless I missed something?)

(The actual change to the constructor itself is needed, though!)

Cc: @andreasnoack

### DIFF
--- a/src/default_dict.jl
+++ b/src/default_dict.jl
@@ -14,38 +14,20 @@
 # subclassed.
 #
 
-_oldDefaultDictBase = """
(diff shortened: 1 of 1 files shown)


tkelman:
( `include_string` is also a possibly-less-scary-looking alternative to `eval(parse())` that could be used in similar circumstances )

PR #262, opened 2017-01-27
Add top_with_handle(::MutableBinaryHeap) function

### LINKED ISSUE #217: Feature request: a function to get the handle when peeking a MutableBinaryHeap with `top`
This is basically what I'm asking for:

``` julia
function top_with_handle(heap::MutableBinaryHeap)
    el = theap.nodes[1]
    t, i = el.value, el.handle
    return t, i
end
```

This is especially useful to know what element was picked, in order to mutate it afterwards with `update!`. I need it for the particular case of `MutableBinaryHeap`, but I reckon it may be useful for other cases as well.

I could submit a PR with that implementation if there is interest. I'm not sure about the name though, suggestions are welcome.

carlobaldassi:
Fixes #217: add an interface to get the handle of the top element of a mutable binary heap, in order to be able to call `update!` on it.

### DIFF
--- a/doc/source/heaps.rst
+++ b/doc/source/heaps.rst
@@ -23,9 +23,11 @@ All heaps in this package are derived from ``AbstractHeap``, and provide the fol
 Mutable heaps (values can be changed after being pushed to a heap) are derived from
 ``AbstractMutableHeap <: AbstractHeap``, and additionally provides the following interface::
 
-  i = push!(h, v)       # adds a value to the heap and and returns a handle to v
+  i = push!(h, v)              # adds a value to the heap and and returns a handle to v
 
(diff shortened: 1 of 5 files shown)


kmsquire:
At some point, we'll want to consolidate the different names applied to the same concept: `SortedDict` and friends have a concept of a token, which is a direct reference to an item in a dictionary.  For heaps, `handle` is used.  Arguably, a reference (ala `Base.Ref`) is quite similar, although the implementation in `Base` is specifically meant to be a pointer reference.

Since `handle` was already used for `Heap`s, let's keep it for now and figure out a plan going forward.

Otherwise, this LGTM.

carlobaldassi:
I don't seem to have the permissions to delete the `cb/tophandle` branch. @kmsquire can you do it?

PR #263, opened 2017-01-28
Fix typing in Binary Heap

ChrisRackauckas:
The `valtree` in the binary heap is not strictly typed. In some codes I have, this causes a 2x performance reduction in the entire code (<.1% is data structure related) due to checking if a binary heap is empty. The simple fix is to make it `Vector{T}` instead of the non-concerete `Array{T}`.

Reference for performance problems: https://github.com/JuliaDiffEq/StochasticDiffEq.jl/issues/13

### DIFF
--- a/src/heaps/binary_heap.jl
+++ b/src/heaps/binary_heap.jl
@@ -104,7 +104,7 @@ end
 
 type BinaryHeap{T,Comp} <: AbstractHeap{T}
     comparer::Comp
-    valtree::Array{T}
+    valtree::Vector{T}
 
     function BinaryHeap(comp::Comp)
         new(comp, Vector{T}(0))


kmsquire:
Thanks!

ChrisRackauckas:
Could a patch be tagged with this change? This'll have a large effect on some of the main DifferentialEquations.jl methods. It's blocking a release which inlines a few more things which previously needed a function barrier to stop the propagation of this type instability.

ChrisRackauckas:
Though I understand if you want to wait until after you migrate to JuliaContainers

kmsquire:
I'll try to tag this today. The move to JuliaContainers shouldn't affect tagging, as a redirect will be in place.

ChrisRackauckas:
Thanks!

kmsquire:
@ChrisRackauckas, wondering if this is urgent or can wait a few days to get some additional changes in?

Reasoning isn't `JuliaContainers` (or whatever org that becomes), but more that there are changes in `SortedDict` constructors that warrant a minor version bump, as well as some upcoming changes in `SortedSet`s and `SortedMultiDict`s (with pull requests) that warrant a minor version bump, and I would rather only bump the minor version once.

If this is urgent, I can just this change on a branch without the `SortedDict` changes and tag (cc: @tkelman), and then tag all of the sorted container changes at once (this commit will still be in the history, albeit with a different SHA).

Let me know your thoughts.

ChrisRackauckas:
It's not urgent. If it takes a week or two I wouldn't mind. I mentioned it because the tag frequency in this repo looks unpredictable, and that is a small enough change that I was hoping to make sure that it wouldn't just sit two months waiting for a larger change to happen. But since it sounds like you have something else on your plate that you're churning through, I can wait.

kmsquire:
If it starts to take a while (say longer than a week), bump this and I'll
put it on a branch and tag it. But my hope is that the other changes are in
over the next few days.

ChrisRackauckas:
Thanks!

PR #266, opened 2017-01-29
RFC: Update `SortedSet` constructors

kmsquire:
* `SortedSet` constructors can now take nothing, an ordering, an iterable,
  or an ordering and an interable, with or without specifying the set
  element type

### DIFF
--- a/doc/source/sorted_containers.rst
+++ b/doc/source/sorted_containers.rst
@@ -206,20 +206,18 @@ Constructors for Sorted Containers
   may also be constructed via ``SortedMultiDict(K[], V[], o)`` where
   the ``o`` argument is optional.
 
-``SortedSet(iter,o)``
+``SortedSet()``
+``SortedSet{K}()``
+  Construct a SortedSet{Any} with ``Forward`` ordering.
+
+``SortedSet(iter,o=Forward)``
+``SortedSet{K}(iter,o=Forward)``
+``SortedSet(o, iter)``
+``SortedSet{K}(o, iter)``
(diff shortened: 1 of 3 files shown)


kmsquire:
@StephenVavasis (or anyone else), any comments here?  These changes to the `SortedSet` constructors are roughly equivalent to the ones I made to `SortedDict`s.  
CI has already passed (a minor refactor is going through CI again right now), so I'll probably merge Sunday evening or Monday if there aren't any comments.

StephenVavasis:
Kevin,

This looks fine to me.  In another pull request from today, @timholy noted that some of the constructors are causing ambiguity warnings in recent 0.6.0 masters, so possibly you can check for this?  Apparently there is a tool to search for possible ambiguities.  

Also, I mentioned on another branch that, as of 0.5.0, there is now an AbstractSet abstract class in Base, so maybe SortedSet should be declared as a subtype of AbstractSet.  As far as I can see, there are no generic methods yet defined for AbstractSet.

-- Steve

kmsquire:
Thanks for the review.  Yes, it would be good to make SortedSet a subtype of AbstractSet. At the least, it would let SortedSets be used where Sets are expected (assuming that functions restrict their inputs to AbstractSet instead of Set, if they do so at all).

I'm not going to do that in this PR, because at the least, AbstractSet isn't defined in Julia v0.4 or Compat.jl (I assume?), and we're still supporting that version. 

I did see @timholy's PR, and will look closer at it. I suspect that there won't be problems with SortedSets with this PR, and probably wouldn't be if I finished the same refactor of the SortedMulitDict constructors.  But the test is good.

PR #267, opened 2017-01-29
RFC: SortedMultiDict constructor updates

kmsquire:
Update SortedMultiDict constructors to accept type parameters
* Mirrors recent updates to SortedDict constructors

@StephenVavasis, can you please review?

One change, mirrored in all of the other `Dict`s, is that I deprecated `SortedMultiDict(ks,vs)` (constructor from arrays).  However, `SortedMultiDicts` are different enough that I would like feedback on that change (and anything else.)

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -118,4 +118,14 @@ module DataStructures
 
     @deprecate DefaultOrderedDict(default, ks, vs) DefaultOrderedDict(default, zip(ks, vs))
     @deprecate DefaultOrderedDict{K,V}(::Type{K}, ::Type{V}, default) DefaultOrderedDict{K,V}(default)
+
+    function SortedMultiDict{K,V}(ks::AbstractVector{K},
+                                  vs::AbstractVector{V},
+                                  o::Ordering=Forward)
+        Base.depwarn("SortedMultiDict(ks, vs, o::Ordering=Forward) is deprecated.\n" * "Use SortedMultiDict(o, zip(ks,vs)) or SortedMultiDict(zip(ks, vs))", :SortedMultiDict)
+        if length(ks) != length(vs)
+            throw(ArgumentError("SortedMultiDict(ks,vs,o): ks and vs arrays must be the same length"))
+        end
+        SortedMultiDict(o, zip(ks,vs))
+    end
 end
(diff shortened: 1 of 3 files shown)


kmsquire:
`SortedMultiDict` tests still need updating.

tkelman:
don't mind me, just double-checking that travis is running properly

kmsquire:
It took a while, but this PR finally has full coverage of the diff.

@StephenVavasis, this is similar to the changes to `SortedDict` and `SortedSet`.  If you have the chance, can you take a gander at it?  Either way, I'll plan to merge later this week.

StephenVavasis:
Kevin,

It looks good to me.  However, it is now difficult for me to follow the logic to see exactly which constructor is invoked because the rules are fairly complicated.   As you know, the pitfall is that we don't want a user to think he/she is creating a `SortedDict` for `Int64=>Float64` (say) but has actually created one that is `Any=>Any` because he/she snagged the wrong constructor.   In this scenario, there won't be an error message, but there will be a significant performance hit.

Could I suggest that in the test sets in `test_sorted_containers.jl`, in the sections that test constructors, we also check the types of the constructed containers?  For example, after line 1634 of `test_sorted_containers.jl`, we could insert:
```julia
@test keytype(sd1) == (@compat ASCIIString) && valtype(sd1) == Int
```
and so on for the other constructor invocations in this section?  If you want, I can code this up and make a separate PR.

-- Steve

kmsquire:
Thanks for the review, @StephenVavasis!

> However, it is now difficult for me to follow the logic to see exactly which constructor is invoked because the rules are fairly complicated. 

Yes, I agree that can be hard to follow.  If you have any suggestions on reorganizing the code without losing functionality, a PR would be very welcome.

> Could I suggest that in the test sets in `test_sorted_containers.jl`, in the sections that test constructors, we also check the types of the constructed containers? For example, after line 1634 of `test_sorted_containers.jl`, we could insert:
>
> ```
> @test keytype(sd1) == (@compat ASCIIString) && valtype(sd1) == Int
> ```

I think that testing the constructed types is a good idea.  However, I think it's clearer to write it as

```julia
@test typeof(sd1) == @compat SortedDict{ASCIIString, Int}
```

If you would do that in a separate PR, that would be great.

Unless you would prefer I wait, I'll merge this tonight or tomorrow, barring further comments.

PR #268, opened 2017-02-09
Fix 0.6 depwarns

yuyichao:


kmsquire:
Hi @yuyichao, I can probably find it, but could you point to the change in Base that requires this?  I think it significantly detracts from readability, and want to discuss that more in the original issue.  (Will look for it later if you don't respond.)

yuyichao:
The corresponding PR is https://github.com/JuliaLang/julia/pull/20308, which does not require this syntax but this is the simplest way I can find to make it work on old versions.

yuyichao:
Note that it is possible to get some form of the new syntax to work on old versions but it'll either require duplicating the code or quite involved hack in `@compat`. Given that the `::Type{...}` trick works and seems to be accepted by people so far, I'm not too excited about implementing anything fancy in `@compat` now.....

PR #269, opened 2017-02-10
indexing interface for MutableBinaryHeap

CarloLucibello:
we should give users the opportunity to access the elements of a MutableBinaryHeap, since this can be done efficiently


kmsquire:
Sorry not to get to this sooner.  Thanks for the contribution!

PR #273, opened 2017-03-03
Fix SortedSet ambiguities on julia v0.5, v0.6

kmsquire:
SortedSet has a couple of constructors which look like
```julia
SortedSet(o::Ordering, iter) = sortedset_with_eltype(o, iter, eltype(iter))
SortedSet(iter, o::Ordering=Forward) = sortedset_with_eltype(o, iter, eltype(iter))
```

Because this causes an ambiguity to be shown on Julia v0.4, the following constructor was added for that version only:
```julia
SortedSet(o1::Ordering,o2::Ordering) =
    throw(ArgumentError("SortedSet with two parameters must be called with an Ordering and an interable"))
```

However, DataStructure tests recently started testing for ambiguities, and the lack of this definition on v0.5 and v0.6 causes problems.  So this PR just defines it for all julia versions.

### DIFF
--- a/src/sorted_set.jl
+++ b/src/sorted_set.jl
@@ -19,22 +19,18 @@ end
 SortedSet() = SortedSet{Any,ForwardOrdering}(Forward)
 SortedSet{O<:Ordering}(o::O) = SortedSet{Any,O}(o)
 
-if VERSION < v"0.5"
-    # To address ambiguity warnings on Julia v0.4
-    SortedSet(o1::Ordering,o2::Ordering) =
-        throw(ArgumentError("SortedSet with two parameters must be called with an Ordering and an interable"))
-end
+# To address ambiguity warnings on Julia v0.4
+SortedSet(o1::Ordering,o2::Ordering) =
+    throw(ArgumentError("SortedSet with two parameters must be called with an Ordering and an interable"))
 SortedSet(o::Ordering, iter) = sortedset_with_eltype(o, iter, eltype(iter))
 SortedSet(iter, o::Ordering=Forward) = sortedset_with_eltype(o, iter, eltype(iter))
 
 @compat (::Type{SortedSet{K}}){K}() = SortedSet{K,ForwardOrdering}(Forward)
 @compat (::Type{SortedSet{K}}){K,O<:Ordering}(o::O) = SortedSet{K,O}(o)
 
-if VERSION < v"0.5"
-    # To address ambiguity warnings on Julia v0.4
-    @compat (::Type{SortedSet{K}}){K}(o1::Ordering,o2::Ordering) =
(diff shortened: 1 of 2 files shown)


kmsquire:
[AppVeyor](https://ci.appveyor.com/project/kmsquire/datastructures-jl) and CodeCov (obviously) ran, but neither showed up under "checks" here.  @tkelman, any thoughts/suggestions?  Might it have to do with the move to JuliaCollections?

kmsquire:
(Closing and reopening to retrigger.)

kmsquire:
Okay, I revoked and reauthorized AppVeyor.  JuliaCollections wasn't originally authorized, but now it should be.  Might have to do the same for CodeCov,

kmsquire:
(And I could have done that without revoking access, through the Authorized Applications interface.)

StephenVavasis:
@kmsquire, I am trying to add a few tests to test_sorted_containers.jl, but something unexpected is occurring.  In Julia 0.4.6, I downloaded the latest master via `Pkg.checkout("DataStructures")`, and then I ran `Pkg.test("DataStructures")`, which worked without a hitch.  However, when I ran the test file for sorted containers directly from the REPL via `include(joinpath(Pkg.dir("DataStructures"), "test/test_sorted_containers.jl"))` I received many error message.  These messages were legitimate, e.g., when I copy line 321 of `test_sorted_containers.jl` to the REPL, I indeed get an error message:

```julia
julia> m02 = DataStructures.SortedDict{Int,Float64}()
ERROR: MethodError: `convert` has no method matching convert(::Type{DataStructures.SortedDict{Int64,Float64,Ord<:Base.Order.Ordering}})
This may have arisen from a call to the constructor DataStructures.SortedDict{Int64,Float64,Ord<:Base.Order.Ordering}(...),
since type constructors fall back to convert methods.
Closest candidates are:
  convert{T}(::Type{T}, ::T)
  call{T}(::Type{T}, ::Any)
 in call at essentials.jl:57
```

So I'm quite confused why I don't see these error message from `Pkg.test("DataStructures")`?

kmsquire:
I took a quick look, and nothing obvious jumps out at me. I'll try to look later, but in the mean time, you could try to see if running the lines in `run rests.jl` manually first, affect whether or not you get errors.

StephenVavasis:
@kmsquire, I got the same error messages from:
```julia
 include(joinpath(Pkg.dir("DataStructures"), "test/runtests.jl"))
```

StephenVavasis:
@kmsquire, never mind, the error messages all went away when I restarted Julia.  There must have been some old versions of DataStructures.jl (older than the current master) polluting my namespace.

kmsquire:
Ok, glad to hear it wasn't anything major. Cheers!

PR #274, opened 2017-03-08
Move not_iterator_of_pairs function from sorted_dict.jl to dict_support.jl

kmsquire:
~~This function was moved to dict_support.jl, but wasn't removed from its original location.~~

### DIFF
--- a/src/dict_support.jl
+++ b/src/dict_support.jl
@@ -1,6 +1,12 @@
 # support functions
 
-# These functions are defined in Base, but are not exported,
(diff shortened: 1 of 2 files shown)


kmsquire:
Whoops.. I only had that change locally. I should probably push the other modified file.

kmsquire:
Okay, got it right this time.

Moving this function from sorted_dict.jl to dict_support.jl, because it's used by both SortedDicts and SortedMultiDicts.

PR #278, opened 2017-03-19
Ignore .DS_Store files

toblux:


kmsquire:
@toblux, thanks for paying attention to details like `.DS_Store` being committed, and for taking the time to make this pull request. 

For `.gitignore`, a rule of thumb that is good to follow and that we try to follow here, is to use it for files that are generated by this repository (such as the build directory already in there).

Obviously, `.DS_Store` isn't something that we want in the repo, and this one slipped in (thanks for your other PR that removed it).

Generally, for OS-generated files like this (and editor generated files, etc.), we encourage users to set up a global `.gitignore` file, and include files like this that are specific to their own workflow.

Cheers!

PR #286, opened 2017-05-20
Reduce type strictness for accumulator

oxinabox:
Accumulator was overly strictly typed.
This was breaking the ability to use it for `SubString`s and `String`s interchangeably.

There is no need to type-things any stricter than the backing dictionary does.


kmsquire:
LGTM.

PR #290, opened 2017-06-01
Add SortedVector type and tests

dpsanders:
Since I needed a SortedVector type, here it is.

The hardest part was writing the binary search.
A simpler implementation would be to just to push the new element onto the end and then use `sort!`, but that seemed to be slightly slower in my not-extensive-nor-rigorous tests.

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -8,7 +8,7 @@ module DataStructures
                  show, dump, empty!, getindex, setindex!, get, get!,
                  in, haskey, keys, merge, copy, cat,
                  push!, pop!, shift!, unshift!, insert!,
-                 union!, delete!, similar, sizehint!,
+                 union!, delete!, similar, sizehint!, resize!,
                  isequal, hash,
                  map, reverse,
                  first, last, eltype, getkey, values, sum,
@@ -51,6 +51,8 @@ module DataStructures
     export orderobject, ordtype, Lt, compare, onlysemitokens
 
     export MultiDict, enumerateall
+
+    export SortedVector
 
     import Base: eachindex, keytype, valtype
 
@@ -101,6 +103,8 @@ module DataStructures
     export PriorityQueue, peek
 
     include("priorityqueue.jl")
+
+    include("sorted_vector.jl")
 
     # Deprecations
 
--- a/src/sorted_vector.jl
+++ b/src/sorted_vector.jl
@@ -1,0 +1,94 @@
+immutable SortedVector{T, F<:Function}
+    data::Vector{T}
+    by::F
+
+    function SortedVector(data::Vector{T}, by::F)
+        new(sort(data), by)
+    end
+end
+
+
+SortedVector{T,F}(data::Vector{T}, by::F) = SortedVector{T,F}(data, by)
+SortedVector{T}(data::Vector{T}) = SortedVector{T,typeof(identity)}(data, identity)
+
+function show(io::IO, v::SortedVector)
+    print(io, "SortedVector($(v.data))")
+end
+
+
+
+getindex(v::SortedVector, i::Int) = v.data[i]
+length(v::SortedVector) = length(v.data)
+
+
+function insert!{T}(v::SortedVector{T}, i::Int, x::T)
+    push!(v.data, i, x)
+    return v
+end
+
+function push!{T}(v::SortedVector{T}, x::T)
+    i = binary_search(v, x)
+    insert!(v.data, i, x)
+    return v
+end
+
+function deleteat!(v::SortedVector, i::Int)
+    deleteat!(v.data, i)
+    return v
+end
+
+function resize!(v::SortedVector, n::Int)
+    resize!(v.data, n)
+    return v
+end
+
+
+"""
+Do binary search for item `x` in *sorted* vector `v`.
+Returns the lower bound for the position of `x` in `v`.
+"""
+function binary_search(v::SortedVector, x)
+
+    by = v.by
+
+    a, b = 1, length(v)
+
+    if by(x) < by(v[a])
+        return 1
+
+    elseif by(x) > by(v[b])
+        return b + 1
+    end
+
+    m = (a + b) ÷ 2  # mid-point
+
+    while abs(a - b) > 1
+
+        # @show a, b, v[a], v[b]
+
+        if by(v[m]) == by(x)
+            return m
+        end
+
+        if by(x) < by(v[m])
+            b = m
+        else
+            a = m
+        end
+
+        m = (a + b) ÷ 2  # mid-point
+
+    end
+
+    # @show a, b, v[a], v[b]
+
+    if by(v[a]) == by(x)
+        return a
+    end
+
+    if by(v[b]) == by(x)
+        return b
+    end
+
+    return b  # a is a lower_bound; insert in position b
+end
--- a/test/test_sorted_vector.jl
+++ b/test/test_sorted_vector.jl
@@ -20,7 +20,7 @@
         v = SortedVector([(5, "a"), (2, "b")], x->x[1])
         @test v.data == [(2, "b"), (5, "a")]
 
-        push!(v, (3, "z")
+        push!(v, (3, "z"))
         @test v.data == [(2, "b"), (3, "z"), (5, "a")]
 
         push!(v, (6, "z"), (4, "y"))
--- a/test/test_sorted_vector.jl
+++ b/test/test_sorted_vector.jl
@@ -1,0 +1,34 @@
+@testset "SortedVector" begin
+
+    @testset "With standard ordering" begin
+
+        v = SortedVector([5, 2])
+        @test v.data == [2, 5]
+
+        push!(v, 3)
+        @test v.data == [2, 3, 5]
+
+        push!(v, 4, 6)
+        @test v.data == [2, 3, 4, 5, 6]
+
+        push!(v, 1)
+        @test v.data == [1, 2, 3, 4, 5, 6]
+    end
+
+    @testset "With ordering function" begin
+
+        v = SortedVector([(5, "a"), (2, "b")], x->x[1])
+        @test v.data == [(2, "b"), (5, "a")]
+
+        push!(v, (3, "z")
+        @test v.data == [(2, "b"), (3, "z"), (5, "a")]
+
+        push!(v, (6, "z"), (4, "y"))
+        @test v.data == [(2, "b"), (3, "z"), (4, "y"), (5, "a"), (6, "z")])
+
+        push!(v, (1, "x"))
+        @test v.data == [(1, "x"), (2, "b"), (3, "z"), (4, "y"), (5, "a"), (6, "z")])
+
+    end
+
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -31,7 +31,8 @@ tests = ["int_set",
          "multi_dict",
          "circular_buffer",
          "sorting",
-         "priorityqueue"]
+         "priorityqueue",
+         "sorted_vector"]
 
 if length(ARGS) > 0
     tests = ARGS


StephenVavasis:
This functionality is already available in the SortedSet container.  Furthermore, SortedSet will in general be more efficient because insertion and deletion to SortedSet require O(log n) operations, whereas insertion and deletion in SortedVector require O(n) operations.  

In addition, Base already contains the functions: issorted, searchsorted_first, and searchsorted_last to deal with sorted vectors.

If you believe that, despite what is already available. there is a use-case for the new SortedVector, then possibly you could document the use-case.  Also, you might consider following the same interface provided by SortedSet and the other sorted containers.  For SortedSet, the second type parameter is an order object rather than a function.

dpsanders:
Thanks, I see your point.

One possible answer is the overhead for storing the set, e.g.
```
julia> s = SortedSet([7, 5, 1])
SortedSet([1, 5, 7],
Base.Order.ForwardOrdering())

julia> Base.summarysize(s)
424

julia> v = SortedVector([7, 5, 1])
SortedVector([1, 5, 7])

julia> Base.summarysize(v)
32
```

I understand that the rotations etc. needed to balance the set can be expensive, but it's not very clear to me how to benchmark things like that which modify the structure being tested.

dpsanders:
Also, in my application, I will need to have multiple values that compare as being equal (with a non-standard comparison that only compares the first element of a tuple). This doesn't seem to be possible using `SortedSet`.

dpsanders:
I can't see from the docs how to use a non-standard ordering, e.g. `x->x[1]` for a tuple.

StephenVavasis:
If you want a structure that will preserve multiple insertions of keys that are distinct under == but test as equals under the sort order, then your best bet would be SortedMultiDict{K,Void}.  The SortedMultiDict container allows multiple keys that test as equal by the sort order.  

You can implement a test based on the first entry either using the Lt type from Base
(use `Lt((x,y) -> x[1] < y[1])` as the order object), or implement this as your own type -- something like the following (untested)

```
immutable OrderByFirstEntry <: Ordering
end

eq(::OrderByFirstEntry, a, b) = a[1] == b[1]
lt(::OrderByFirstEntry, a, b) = a[1] < b[1]
```
and then use `s = SortedMultiDict(K,Void,OrderByFirstEntry)(OrderByFirstEntry())` to create the initially empty object.

dpsanders:
Thanks. One of the reasons I need a sorted vector or related is that I will frequently need to remove all of those entries that are larger than a certain value. Is that functionality available / easy to implement in `SortedMultiDict`? It feels like it should just corresponding to lopping off a branch from the tree and doing a rebalance.

StephenVavasis:
In principle, it is possible to chop off an entire subtree of the tree and then rebalance, but it would be a lot of code to write.  (Have a look at the `delete!` function in `balanced_tree,jl`.)   Depending how often you need to carry out this operation, a Vector solution might be better than SortedMultiDict.

dpsanders:
OK thanks. I guess that would be a high complexity operation? (O(number deleted items) or similar?)

StephenVavasis:
Currently, to delete k items from a sorted multidict would require O(k*log n) operations.  If new special-purpose code were written for deleting a block of k consecutive items, then I think this would require O(k + log n) operations.  To delete the last k items from a vector requires only O(1) time (since it corresponds to resize).

kmsquire:
> A simpler implementation would be to just to push the new element onto the end and then use sort!, but that seemed to be slightly slower in my not-extensive-nor-rigorous tests.

Just for curiosity: if you run those tests, but use `TimSort` from `SortingAlgorithms.jl`, is that faster?  That algorithm targets "mostly sorted" lists.

dpsanders:
I guess the slow part is moving the data around, which `TimSort` would also (presumably) end up doing.

`SortedVector`s seem to be working well in my application. It seems a shame to just keep this functionality inside my package, when `DataStructures.jl` is the obvious home for it, so any objection to merging this?

src/sorted_vector.jl
```diff
@@ -0,0 +1,52 @@
+"""
+A `SortedVector` behaves like a standard Julia `Vector`, except that its elements are stored in sorted order, with an optional function `by` that determines the sorting order in the same way as `Base.searchsorted`.
+"""
+immutable SortedVector{T, F<:Function}
+    data::Vector{T}
+    by::F
+
+    function SortedVector(data::Vector{T}, by::F)
+        new(sort(data), by)
```
kmsquire: I think you need to pass `by` to `sort()` here.

Should also pass through `rev` and `alg` as parameters to `sort()`.

src/sorted_vector.jl
```diff
@@ -0,0 +1,52 @@
...
+SortedVector{T}(data::Vector{T}) = SortedVector{T,typeof(identity)}(data, identity)
+
+function show(io::IO, v::SortedVector)
+    print(io, "SortedVector($(v.data))")
+end
+
+
+
+getindex(v::SortedVector, i::Int) = v.data[i]
+length(v::SortedVector) = length(v.data)
```
kmsquire: These and a few functions below can be simplified using `@delegate`

test/test_sorted_vector.jl
```diff
@@ -0,0 +1,34 @@
...
+        push!(v, (3, "z"))
+        @test v.data == [(2, "b"), (3, "z"), (5, "a")]
+
+        push!(v, (6, "z"), (4, "y"))
+        @test v.data == [(2, "b"), (3, "z"), (4, "y"), (5, "a"), (6, "z")]
+
+        push!(v, (1, "x"))
+        @test v.data == [(1, "x"), (2, "b"), (3, "z"), (4, "y"), (5, "a"), (6, "z")]
+
+    end
```
kmsquire: Please add a testset using

```julia
v = SortedVector([1,2,3], x->-x)
```

src/sorted_vector.jl
kmsquire: (And for that matter, `rev` could/should probably be a separate field here as well)

kmsquire:
Also for reference: 
* https://github.com/JuliaLang/Juleps/blob/master/Find.md
* https://github.com/JuliaLang/julia/issues/22354#issuecomment-308314699

kmsquire:
Cc: @nalimilan

src/sorted_vector.jl
dpsanders: Good catch, thanks.

`rev` seems to be redundant with `by`.

nalimilan:
Thanks for Ccing me, @kmsquire. Indeed `SortedVector` would be very useful to allow getting rid of the `searchsorted*` functions in favor of dedicated methods for the generic find/search functions. There seems to be some support by core devs about this.

However, I'm not sure it would be OK to deprecate `searchsorted*` functions in Base and require people to use DataStructures. Since essential features like sorted dicts are also in the package, it may be fine. If not, `SortedVector` would have to be included in Base. That's not a lot of code, but the trend has generally been to shrink Base as much as possible.

src/sorted_vector.jl
nalimilan: `rev` isn't redundant with `by`: `sort` takes  both arguments. Would be better to make them keyword arguments, to match `sort`.

It would also be useful to provide a way to sort the data in-place, and to declare that the input is already sorted. Not sure what's the best way to do this; maybe just a `sorted` keyword argument.

src/sorted_vector.jl
```diff
@@ -0,0 +1,52 @@
+"""
+A `SortedVector` behaves like a standard Julia `Vector`, except that its elements are stored in sorted order, with an optional function `by` that determines the sorting order in the same way as `Base.searchsorted`.
+"""
+immutable SortedVector{T, F<:Function}
+    data::Vector{T}
```
nalimilan: Should use `AbstractVector` everywhere (here, via a type parameter).

src/sorted_vector.jl
```diff
@@ -0,0 +1,52 @@
+"""
+A `SortedVector` behaves like a standard Julia `Vector`, except that its elements are stored in sorted order, with an optional function `by` that determines the sorting order in the same way as `Base.searchsorted`.
```
nalimilan: Wrap at 92 chars. Also, better refer to `sort` than to  `searchsorted`. The docstring should also mention explicitly what happens on insert.

src/sorted_vector.jl
```diff
@@ -0,0 +1,52 @@
...
+    print(io, "SortedVector($(v.data))")
+end
+
+
+
+getindex(v::SortedVector, i::Int) = v.data[i]
+length(v::SortedVector) = length(v.data)
+
+
+function insert!{T}(v::SortedVector{T}, i::Int, x::T)
```
nalimilan: Shouldn't use `::T`, just leave `push!` do the conversion automatically if possible. Same below.

Though this method should probably throw an error, since it doesn't respect the contract of inserting `x` at position `i`.

src/sorted_vector.jl
```diff
@@ -0,0 +1,52 @@
...
+    push!(v.data, i, x)
```
nalimilan: `i` is incorrect here, right?

src/DataStructures.jl
```diff
@@ -102,6 +104,8 @@ module DataStructures
 
     include("priorityqueue.jl")
 
+    include("sorted_vector.jl")
```
nalimilan: No underscore.

test/test_sorted_vector.jl
nalimilan: Should also test all methods defined for the type. Can you also test that an error is thrown when calling `setindex!`?

src/sorted_vector.jl
```diff
@@ -0,0 +1,52 @@
...
+    by::F
+
+    function SortedVector(data::Vector{T}, by::F)
+        new(sort(data), by)
+    end
+end
+
+
+SortedVector{T,F}(data::Vector{T}, by::F) = SortedVector{T,F}(data, by)
+SortedVector{T}(data::Vector{T}) = SortedVector{T,typeof(identity)}(data, identity)
```
nalimilan: You can get rid of this function by adding `by=identity` to the previous one.

src/sorted_vector.jl
```diff
@@ -0,0 +1,52 @@
...
+
+
+function insert!{T}(v::SortedVector{T}, i::Int, x::T)
+    push!(v.data, i, x)
+    return v
+end
+
+function push!{T}(v::SortedVector{T}, x::T)
+    i = searchsortedfirst(v.data, x, by=v.by)
+    insert!(v.data, i, x)
```
JeffBezanson: I believe this should just check that the pushed item is greater than the current last item in the vector. Otherwise it doesn't implement the contract that `pop!` returns the last thing you `push!`ed.

src/sorted_vector.jl
JeffBezanson: I think the constructor shouldn't call `sort`. You end up needing `SortedVector(v, dont_sort=true)`, instead of simply writing `SortedVector(v)`, `SortedVector(sort(v))`, `SortedVector(sort!(v))` as needed.

To support different orderings, `SortedVector` should have a `Base.Order.Ordering` type parameter.

dpsanders:
The documentation for pop! says that it returns the last element in the structure, not the last element added to the structure. 

To me push! just means "add this element to the data structure"; I don't see why it should be in the last place.

dpsanders:
Many thanks for the detailed review @nalimilan. I don't think I'll have time to do all that any time soon though.

src/sorted_vector.jl
dpsanders: If the constructor doesn't call `sort`, you can end up with a non-sorted `SortedVector`. Or am I missing something? Or just call `issorted` first to see if it needs sorting? (But is that much faster than just calling `sort` if it's already sorted?)

src/sorted_vector.jl
dpsanders: What is the use case for using an `AbstractVector`?

nalimilan: I don't know, one could use a `DataArray`, a `CategoricalArray`... Is there any reason not to support it?

kmsquire:
@nalimilan said
> However, I'm not sure it would be OK to deprecate `searchsorted*` functions in Base and require people to use DataStructures. Since essential features like sorted dicts are also in the package, it may be fine. If not, `SortedVector` would have to be included in Base. That's not a lot of code, but the trend has generally been to shrink Base as much as possible.

I agree.  I had a longer response with a more detailed proposal for what specifically to do with `searchsorted*` (in a type stable way) ready for https://github.com/JuliaLang/julia/issues/22354#issuecomment-308314699, but I thought it was getting too far from the topic there.  I'll open an issue in `julialang`.

src/sorted_vector.jl
dpsanders: I see. I was just worried that there are some subtypes of AbstractVector that wouldn't support the necessary operations.

nalimilan: AFAICT you only use functions in the `AbstractArray` interface, right? Anyway, even if a method fails for some peculiar types, it cannot be worse than if you only accept `Vector`.

StephenVavasis: If I understand correctly, there will be a big performance hit for making the inner container an `AbstractVector`. The compiler won't be able to ascertain the actual type of the inner container at compile-time so there will be a lot more run-time dispatching.  To avoid this problem, you would need to make both the data type and the inner container type parameters of `SortedVector`.

dpsanders: Making the container type a parameter is what @nalimilan was suggesting.

src/sorted_vector.jl
dpsanders: Ah, I understand the point. But it should definitely not be possible to create the object without any sorting at all.

src/DataStructures.jl
dpsanders: Have a look at the other file names in the package.

nalimilan: Funny, my comment was based on `priorityqueue.jl`, which is the only example shown in the diff, and apparently also the only exception.

kmsquire:
bump

nalimilan:
Actually, `searchsorted` functions are used in Base for sparse matrices, so the replacement will have to be defined in Base too. Though we could use a very simpler wrapper called e.g. `Sorted` which would work with any iterable, but be read-only. `SortedVector` could live in DataStructures and support more operations.

dpsanders:
Sorry, dropped the ball on this one.
If anybody would like to finish this off, feel free...

tpapp:
I am interested in reviving this, since it would be nice to have a vector-like container that simply indicates that the contents are sorted.

However, from the long discussion above it seems there was some resistance to the idea. I wonder if the package maintainers would indicate under what conditions they would *consider* merging something like this, otherwise I can just start another package for it.

tpapp:
Made a package for this, comments welcome:
https://github.com/tpapp/SortedVectors.jl

PR #292, opened 2017-06-07
`@inline` binary heap `top` call

ChrisRackauckas:
This will allow `@inbounds` to propagate into this call.

### DIFF
--- a/src/heaps/binary_heap.jl
+++ b/src/heaps/binary_heap.jl
@@ -147,6 +147,6 @@ end
 
 Returns the element at the top of the heap `h`.
 """
-top(h::BinaryHeap) = h.valtree[1]
+@inline top(h::BinaryHeap) = h.valtree[1]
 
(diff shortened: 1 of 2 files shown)


yuyichao:
is this function too large to be automatically inlined?

ChrisRackauckas:
In profiling I found this to be a surprisingly expensive operation, so I assumed this must be part of it. I'll see if I can find out more directly. 

(But I also thought that inline-ing will only let `@inbounds` through if you `@inline`)

oxinabox:
The inlining properation rules are a bit complex (for good reasons).
https://docs.julialang.org/en/latest/devdocs/boundscheck/

AFAICT @ChrisRackauckas  is generally right, though there might be a better way (eg with `@propergate_inbounds`).
But in general a lot of the code needs an audit over its use of `@inline` as a lot was written for versions of julia with marginally (or significantly) different inlining rules.

PR #295, opened 2017-06-15
fix broken find_root for non-int disjoint sets

### LINKED ISSUE #294: DisjointSets with other data types: find_root returns internal integer
DisjointSets internally maps to IntDisjointSet using a Dict. But the `find_root` function returns the internal integer mapping instead of the root element:

https://github.com/JuliaCollections/DataStructures.jl/blob/master/src/disjoint_set.jl#L132

Since the dict represents a one-to-one mapping,  it should be ok to to return key corresponding to the integer value, e.g.,  by reversing the dict:

```
find_root{T}(s::DisjointSets{T}, x::T) = map(reverse,s.intmap)[find_root(s.internal, s.intmap[x])]
```
I considered this as a quick fix, but  the time-complexity of the above `find_root` implementation unfortunately depends on the size and it could be constant time.  I think that it would be better to extend the `DisjointSets` type to have a reverse intmap as well. This would ensure constant time `find_root` at the cost of a little extra  bookkeeping (constant time) in the `push!` method. If you think this is an acceptable solution, I will submit a pull request on this.

cth:
This is my attempt to resolve #294

### DIFF
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -107,19 +107,19 @@ end
 
 @compat type DisjointSets{T}
     intmap::Dict{T,Int}
-    revmap::Dict{Int,T}
+    revmap::Vector{T}
     internal::IntDisjointSets
 
     function (::Type{DisjointSets{T}}){T}(xs)    # xs must be iterable
         imap = Dict{T,Int}()
-	rmap = Dict{Int,T}()
+        rmap = Vector{T}()
         n = length(xs)
         sizehint!(imap, n)
-	sizehint!(rmap, n)
+        sizehint!(rmap, n)
         id = 0
         for x in xs
             imap[x] = (id += 1)
-	    rmap[id] = x
+            push!(rmap,x)
         end
         new{T}(imap, rmap, IntDisjointSets(n))
     end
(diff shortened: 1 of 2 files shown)


src/disjoint_set.jl
```diff
@@ -107,17 +107,21 @@ end
 
 @compat type DisjointSets{T}
     intmap::Dict{T,Int}
+    revmap::Dict{Int,T}
     internal::IntDisjointSets
 
     function (::Type{DisjointSets{T}}){T}(xs)    # xs must be iterable
         imap = Dict{T,Int}()
+	rmap = Dict{Int,T}()
```
kmsquire: Please set your editor to use spaces instead of tabs.

src/disjoint_set.jl
```diff
@@ -129,17 +133,18 @@ num_groups(s::DisjointSets) = num_groups(s.internal)
...
+find_root{T}(s::DisjointSets{T}, x::T) = s.revmap[find_root(s.internal, s.intmap[x])]
 
 in_same_set{T}(s::DisjointSets{T}, x::T, y::T) = in_same_set(s.internal, s.intmap[x], s.intmap[y])
 
 for f in (:union!, :root_union!)
     @eval begin
-        ($f){T}(s::DisjointSets{T}, x::T, y::T) = ($f)(s.internal, s.intmap[x], s.intmap[y])
+        ($f){T}(s::DisjointSets{T}, x::T, y::T) = s.revmap[($f)(s.internal, s.intmap[x], s.intmap[y])]
     end
 end
```
kmsquire: I realize that the for loop for this definition was there before, but since there are only two functions and one definition inside the loop per function, I think it would be clearer if both functions were simply written out.

src/disjoint_set.jl
```diff
@@ -107,17 +107,21 @@ end
 
 @compat type DisjointSets{T}
     intmap::Dict{T,Int}
+    revmap::Dict{Int,T}
```
kmsquire: Couldn't this just be a `Vector{T}` of the size of the set?  That would be more efficient (in both time and space) than hashing the number.

cth: Agreed, this performance-wise better. I've updated my pull to accommodate this change.

kmsquire:
Sorry not to get to this sooner.  Thanks for the contribution!

Issue #297, opened 2017-06-24
DisjointSets: set sizes


tejus-gupta:
I noticed that DisjointSets doesn't store set sizes. I wanted to use DisjointSets for an algorithm where I need to quickly find an element's set size. Is set size not stored since users don't usually need this feature and an user can separately store set sizes and update them during set merge operations?

Just to make it clear, I am looking for a function ```find_set_size(a, n)``` which returns n's set size in DisjointSets a. I would be happy to send a pull request.

kmsquire:
HI @tejus-gupta, I was the one who suggested submitting a pull request, and one of the maintainers of DataStructures.jl.  

I didn't implement DisjointSet, but I believe it was patterned after [disjoint sets in boost](http://www.boost.org/doc/libs/1_64_0/libs/disjoint_sets/disjoint_sets.html), which does not store the set size.

Anyway, yes, please do submit a pull request.  Presumably, you'll use a `Vector` to keep track of the size?  If so, let's add a type parameter there, which is either `Int` or `Void`.  `Void` arrays exist, but don't take up any space, so the type parameter should be set to `Void` if the user does not need to keep track of the size of the sets.

Cc: @timholy

timholy:
My own contributions to DisjointSet are quite modest, but I'd be happy to help review a PR. And thanks to you both.

PR #298, opened 2017-06-26
Added function to get element set size in DisjointSets

tejus-gupta:
Added function ```get_size``` for DisjointSets. See #297.

### DIFF
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -139,11 +139,11 @@ get_size(s::IntDisjointSets{<:Void}, x::Integer) = throw(ErrorException("Disjoin
 #
 ############################################################
 
-@compat type DisjointSets{T}
+@compat type DisjointSets{T, S<:Union{Void, Int}}
     intmap::Dict{T,Int}
-    internal::IntDisjointSets
+    internal::IntDisjointSets{S}
 
-    function (::Type{DisjointSets{T}}){T}(xs, store_size::Bool=false)    # xs must be iterable
+    function (::Type{DisjointSets{T, S}}){T, S}(xs)    # xs must be iterable
         imap = Dict{T,Int}()
         n = length(xs)
         sizehint!(imap, n)
@@ -151,9 +151,10 @@ get_size(s::IntDisjointSets{<:Void}, x::Integer) = throw(ErrorException("Disjoin
         for x in xs
             imap[x] = (id += 1)
         end
-        new{T}(imap, IntDisjointSets(n, store_size))
+        new{T, S}(imap, IntDisjointSets{S}(n))
     end
 end
+IntDisjointSets(xs, store_size::Bool=true) = store_size ? IntDisjointSets{eltype(s), Int}(n) : IntDisjointSets{eltype(s), Void}(n)
 
 length(s::DisjointSets) = length(s.internal)
 num_groups(s::DisjointSets) = num_groups(s.internal)
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -23,7 +23,7 @@ type IntDisjointSets{T<:Union{Void, Int}}
     IntDisjointSets{T}(n::Integer) where {T<:Int} = new(collect(1:n), zeros(Int, n), ones(Int, n), n)
     IntDisjointSets{T}(n::Integer) where {T<:Void} = new(collect(1:n), zeros(Int, n), Vector{Void}(), n)
 end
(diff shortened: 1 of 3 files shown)


src/disjoint_set.jl
```diff
@@ -84,17 +87,48 @@ function root_union!(s::IntDisjointSets, x::Integer, y::Integer)
...
     push!(s.ranks, 0)
+    push!(s.sizes, 1)
     s.ngroups += 1
     return x
 end
 
+# get size of set containing element x
+#
+get_size(s::IntDisjointSets{<:Int}, x::Integer) = s.sizes[find_root(s, x)]
+get_size(s::IntDisjointSets{<:Void}, x::Integer) = error("DisjointSets wasn't initialized with store_size = true")
```
kmsquire: Can you please change this to throw an `ErrorException`, rather than calling `error`?  We should be moving away from `error` in this package.

tejus-gupta: :+1:

src/disjoint_set.jl
```diff
@@ -109,15 +143,15 @@ end
     intmap::Dict{T,Int}
     internal::IntDisjointSets
 
-    function (::Type{DisjointSets{T}}){T}(xs)    # xs must be iterable
+    function (::Type{DisjointSets{T}}){T}(xs, store_size::Bool=false)    # xs must be iterable
```
timholy: This is a bit unfortunate because there's no way to inferrably create a `DisjointSet`. But if you add a 2nd parameter then it will be.

src/disjoint_set.jl
```diff
@@ -109,15 +143,15 @@ end
     intmap::Dict{T,Int}
     internal::IntDisjointSets
```
timholy: This is now an abstract type. I suspect you want to add another parameter to `DisjointSets`.

src/disjoint_set.jl
```diff
@@ -13,14 +13,17 @@
...
     ranks::Vector{Int}
+    sizes::Vector{T}
     ngroups::Int
 
     # creates a disjoint set comprised of n singletons
-    IntDisjointSets(n::Integer) = new(collect(1:n), zeros(Int, n), n)
+    IntDisjointSets{T}(n::Integer) where {T<:Int} = new(collect(1:n), zeros(Int, n), ones(Int, n), n)
+    IntDisjointSets{T}(n::Integer) where {T<:Void} = new(collect(1:n), zeros(Int, n), Vector{Void}(), n)
 end
+IntDisjointSets(n::Integer, store_size::Bool=false) = store_size ? IntDisjointSets{Int}(n) : IntDisjointSets{Void}(n)
```
timholy: Not inferrable, but given that there are inferrable alternatives this isn't terrible. Perhaps a code comment explaining this would be helpful.

src/disjoint_set.jl
kmsquire: I believe that `Val` is often discouraged, but I'm wondering if this would be a valid use case?

tejus-gupta: I have added the second parameter for DisjointSet. Is there a way to write a constructor such that ```DisjointSets{Int}([1, 2, 3])``` still works?

kmsquire: Sorry, I missed this!  Yes, you can write a constructor so that `DisjointSets{Int}([1, 2, 3])` still works. Add

```julia
(::Type{DisjointSets{T}}){T}(xs) = DisjointSets{T, Void}(xs)
```

as an external constructor for `DisjointSets`.

kmsquire:
@tejus-gupta, wondering if you still need this for any of your work in `ImageFeatures`?  Are you available to fix this pull request up?

kmsquire:
bump.  Cc: @timholy

kmsquire:
This might be easier to fix up now that we've dropped support for Julia v0.5 in master.

kmsquire:
Bump.

eulerkochy:
What's the status of this PR? 
Umm, can I take it on from here?

tejus-gupta:
Yeah sure. It is mostly ready; you may need to update it for the latest Julia version.

I wanted it for [this PR](https://github.com/JuliaImages/ImageSegmentation.jl/pull/4), but I managed by separately storing set sizes and updating them during merge ops.

Sorry for not replying to earlier comments, I missed the notifications.

eulerkochy:
It's cool. A very subtle amount of work left in this one, I'll get to it straight away!

PR #300, opened 2017-07-03
Dequeuing (priority queue) with values

### LINKED ISSUE #289: PriorityQueue - peek versus dequeue!
Moved from https://github.com/JuliaLang/julia/issues/18517

-----------------

`peek` returns the pair `(key,value)` whereas `dequeue!` only returns the `key`:

``` julia
using Base.Collections
pq = PriorityQueue()
enqueue!(pq, key, value)
key, value = peek(pq)
key = dequeue!(pq)
```

Can `dequeue!` return the pair?

To be consistent with the rest of the interface, should we rename some of these to `push!` and `pop!`?

sbromberger:
Closes #289. Adds `dequeue_with_val!` to PriorityQueue - returns a tuple (key, value). Implemented also for specific dequeuing of a known key.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
 julia 0.4
-Compat 0.9.4
+Compat 0.17.0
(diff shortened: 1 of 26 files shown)


sbromberger:
Erm. Needs an update.

PR #301, opened 2017-07-03
dequeue_with_val! - closes #289

### LINKED ISSUE #289: PriorityQueue - peek versus dequeue!

sbromberger:
Closes #289. Adds dequeue_with_val! to PriorityQueue - returns a tuple (key, value). Implemented also for specific dequeuing of a known key.

### DIFF
--- a/test/test_priorityqueue.jl
+++ b/test/test_priorityqueue.jl
@@ -96,13 +96,13 @@ end
 priorities2 = Dict(zip('a':'e', 5:-1:1))
 pq = PriorityQueue(priorities2)
 try
-    dequeue_with_val!(pq, 'g')
+    dequeue_pair!(pq, 'g')
     error("should have resulted in KeyError")
 catch ex
     @test isa(ex, KeyError)
 end
-@test dequeue_with_val!(pq) == ('e', 1)
-@test dequeue_with_val!(pq, 'b') == ('b', 4)
+@test dequeue_pair!(pq) == Pair('e', 1)
+@test dequeue_pair!(pq, 'b') == Pair('b', 4)
 @test length(pq) == 3
 
 # low level heap operations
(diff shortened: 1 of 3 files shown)


kmsquire:
Thanks, @sbromberger!  Will take a closer look. 

At a high level, I think it would be better to return a `Pair` and have the function named `dequeue_pair`.  Is there a reason the function should return a tuple?

sbromberger:
@kmsquire no reason. I just thought a tuple made sense, but feel free to change it (and the name), as long as we can get the value back :)

test/test_priorityqueue.jl
```diff
@@ -93,6 +93,18 @@ while !isempty(pq)
     @test 10 != dequeue!(pq)
 end
 
+priorities2 = Dict(zip('a':'e', 5:-1:1))
+pq = PriorityQueue(priorities2)
+try
+    dequeue_pair!(pq, 'g')
+    error("should have resulted in KeyError")
```
kmsquire: Better as `throw ErrorException(...)`. We're trying to remove uses of `error` in lieu of throwing errors.

sbromberger: Yeah, that syntax confused me as well but I figured it was better to be consistent with what was there. I'll fix it.

sbromberger:
Thanks, @kmsquire. Is it possible to tag a new version?

kmsquire:
Yep.  (I meant to squash this during the merge, and thought I had, but it turns out I didn't.  Oh, well.)

kmsquire:
https://github.com/JuliaLang/METADATA.jl/pull/10165

PR #302, opened 2017-07-08
Define counter(::Generator)

cstjean:
The call to `_default_eltype` comes from the [`Set` constructor](https://github.com/JuliaLang/julia/blob/903644385b91ed8d95e5e3a5716c089dd1f1b08a/base/set.jl#L12). Also, it might be worth defining a `pushmany!` function; the `for x in blah; push!(c, x) end` loop happens four times already in that file.


src/accumulator.jl
```diff
@@ -24,6 +24,14 @@ function counter{T}(seq::AbstractArray{T})
     return ct
 end
 
+function counter{T<:Base.Generator}(gen::T)
+    ct = counter(Base._default_eltype(T))
```
kmsquire: I wonder if there is a better way to do this?  My only concern is that this is meant to be an internal function and could (in theory) go away...

Anyway, this is useful functionality.  Sorry to let it languish!

cstjean: I wondered that too, but it's how [`Base.Set`](https://github.com/JuliaLang/julia/blob/b46c74cc9c61769aade33e407c202bf88aaf39d1/base/set.jl#L20) does it, and similarly for `Dict`. It's defined as `_default_eltype(itrt::ANY) = Core.Inference.return_type(first, Tuple{itrt})`, so I don't think we can do it correctly using exported functionality.

kmsquire:
Fixed merge conflicts against master.  Will merge once CI passes.  

I need to open up an issue against Base regarding `_default_eltype`.

kmsquire:
As before, (incorrect) coverage changes are https://github.com/JuliaCI/Coverage.jl/issues/132.

PR #311, opened 2017-07-23
Fix most of 0.7 depwarns

yuyichao:
Remove old code for 0.4 and 0.5-.


yuyichao:
I realize maybe I should define "most of 0.7 depwarns" more precisely since I've been copying that commit message around. It's all depwarn in loading and most if not all depwarn in the test not including the type definition depwarn...

PR #316, opened 2017-08-15
PriorityQueue updates (Fixes #312)

### LINKED ISSUE #312: Why PriorityQueue(K,V) instead of PriorityQueue{K,V}()?
It seems to me that we should have `PriorityQueue{K,V}()`, instead of `PriorityQueue(K,V)` (where `K`,`V` are types), since this is more consistent with the way type parameters are treated in Julia.

kmsquire:
* Deprecate PriorityQueue(K::Type, V::Type) in lieu of PriorityQueue{K,V}()
* Update tests for parametric constructor


kmsquire:
@ararsalan, thanks for the review. I'll add a few more tests to cover new code, and then merge.

PR #317, opened 2017-08-18
empty!() for CircularDeque, add it to docs, add some other missing things to docs

KadeG:


src/circular_buffer.jl
```diff
@@ -9,6 +9,11 @@ type CircularBuffer{T} <: AbstractVector{T}
     (::Type{CircularBuffer{T}}){T}(capacity::Int) = new{T}(capacity, 1, T[])
 end
 
+function Base.empty!(cb::CircularBuffer)
+    cb.buffer = typeof(cb.buffer)[]
+    cb
+end
```
kmsquire: I think you also need to reset `cb.first = 1`.  Is there a reason you didn't just call `empty!(cb.buffer)`?

kmsquire: Also, could you add a quick test for this?

KadeG: Changes make. Test added. You're suggestion is much faster. Thanks for the help!

kmsquire:
Coverage reduction is a problem in Coverage.jl: https://github.com/JuliaCI/Coverage.jl/issues/132

LGTM!  Thanks for the contribution.

Issue #325, opened 2017-09-05
New syntax for PriorityQueue?


sbromberger:
So, 0.7 breaks LightGraphs (no big deal, we'll fix it), but I'm confused by the following syntax change:

```
julia> PriorityQueue{Int, Float64, Base.Order.ForwardOrdering}()
ERROR: MethodError: no method matching DataStructures.PriorityQueue{Int64,Float64,Base.Order.ForwardOrdering}()

julia> PriorityQueue{Int, Float64, Base.Order.ForwardOrdering}(Base.Order.ForwardOrdering())
DataStructures.PriorityQueue{Int64,Float64,Base.Order.ForwardOrdering} with 0 entries
```

Is the second example really the right way to do this now?

ETA: it looks like
```
PriorityQueue{Int, Float64}(Base.Order.ForwardOrdering())
```
works as well.

kmsquire:
The purpose of changing the `PriorityQueue` constructor was mostly to simplify the syntax. Your updated syntax

```julia
PriorityQueue{Int, Float64}(Base.Order.ForwardOrdering())
```

is correct for general orderings.  However, since `ForwardOrdering` is the default, you can also just do

```
PriorityQueue{Int, Float64}()
```

which puts it in line with `Dict` constructors.

We could add a special case to make 

```
PriorityQueue{Int, Float64, Base.Order.ForwardOrdering}()
```

work, but it would only be for `ForwardOrdering` (or anything else we special cased), since, e.g., currently, `ReverseOrdering` takes another ordering as a parameter (and type parameter), and it isn't possible to construct one without the internal Ordering parameter (at least without some changes to `Base`).

Also note that the whole `Ordering` framework only exists because it wasn't possible to inline `Function` parameters until Julia v0.5 (I think), and is slated to go away.  So it's probably best to avoid them when you can anyway.  

(And we'll need to update this interface here once again to handle that change, when it happens.)

sbromberger:
The problem was with https://github.com/JuliaGraphs/LightGraphs.jl/blob/dd952cfef9d7a359d9da87108337502737aaef60/src/traversals/maxadjvisit.jl#L52.

kmsquire:
I get a proper deprecation warning for that:

```julia
julia> DataStructures.PriorityQueue(Int, Int, Base.Order.Reverse)
[...]

PR #326, opened 2017-09-05
Added deprectation for PriorityQueue{K,V,ForwardOrdering}()

kmsquire:
* This previously worked, and was broken with the release of DataStructures v0.7


kmsquire:
Cc: @sbromberger

sbromberger:
it's not just `ForwardOrdering()` though. See https://github.com/JuliaCollections/DataStructures.jl/issues/325#issuecomment-327254145

kmsquire:
As noted there, this should be the only case that might have actually broken any code.  Everything else should have been generated a deprecation warning, but should have still worked.  

If that's not the case, though, please let me know soon so it can be fixed--thanks!

PR #328, opened 2017-09-12
Fix DataStructures precompile

TotalVerb:
Close #327.

This is the smallest workaround I could imagine for the 0.7 macro changes, which may not be fixed ever/for a while.

The tests still break on 0.7 but at least it gets past the loading phase.

PR #329, opened 2017-09-12
Fix loading on 0.7

yuyichao:
Replaces https://github.com/JuliaCollections/DataStructures.jl/pull/328

### DIFF
--- a/src/delegate.jl
+++ b/src/delegate.jl
@@ -1,7 +1,17 @@
 # by JMW
+
+function unquote(e::Expr)
+    @assert e.head == :quote
+    return e.args[1]
+end
+
+function unquote(e::QuoteNode)
+    return e.value
+end
+
 macro delegate(source, targets)
     typename = esc(source.args[1])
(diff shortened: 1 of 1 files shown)


kmsquire:
Test crashes should have been fixed by https://github.com/JuliaLang/julia/pull/23700.

@yuyichao @tkelman (or anyone), after such a fix is merged in mainline Julia, I assume rerunning the tests will pull in the latest build--is this true?  If so, how soon after a merge can the tests be rerun, in order to have the fix in place?

yuyichao:
The nightly haven't been updated AFAICT as of a few minutes (to an hour) ago, which is longer than what I usually expect it to take. I've verified locally that the tests doesn't crash anymore though there are still test failures.

Those failures looks unrelated (`nextind` method error) and tests passes on released versions so I'll just merge this now.

yuyichao:
And I might come back to check this later after finish dealing with other packages.

kmsquire:
Thanks Yichao

PR #333, opened 2017-10-03
Add pre-allocation to circular_buffer.jl

sairus7:
Add pre-allocation fill! method for circular buffers.

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -67,6 +67,13 @@ function Base.append!(cb::CircularBuffer, datavec::AbstractVector)
     cb
 end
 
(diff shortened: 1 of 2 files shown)


oxinabox:
Can this have tests added?
Also might as well use `Base.fill!(cb.buffer, data)` rather than ` cb.buffer[:] = data`

sairus7:
I've noticed that preallocated buffer indexing should use another logic (internal cb.length is required for proper indexing). Also, it shouldn't use any functions changing the constant size. 

Is it worth adding another circular buffer type with constant size? I consider it will be faster with some memory overhead.

How do you recommend to do this, (or not)?

kmsquire:
Superceded by #405

Issue #303, opened 2017-07-08
Two-argument `merge!` and `push!` are the same for accumulators


cstjean:
```julia
ct1() = counter(abs.(-5:5))
ct2() = counter([1,2,1,4,5,6])

push!(ct1(), ct2()) == merge!(ct1(), ct2())   # true
```

Why? Semantically, it doesn't make much sense, and pragmatically, it's a needless special-case that precludes having a counter-of-counter.

kmsquire:
I agree that this doesn't make much sense. The `push!` function taking two accumulators should get renamed to `merge!` and the alternative which accepts two or more accumulators should be change to accept 3 or more. 

(Probably needs a deprecation as well, with an end point of Julia v0.7 or 1.0)

Issue #285, opened 2017-05-20
Should `pop!(::Accumulator, v)` set the count of `v` to zero, or just decrement it


oxinabox:
Currently `push!(ct::Accumuator, v)` increase the count of `v` by 1.
`pop!` is the logical oposite of `push!`, so it could be expected to decrement it by 1.
But currently it removes `v` entirely -- setting it's count to zero.

The manual for `pop!` in base is:

> 
> Base.pop! — Method.
> 
> pop!(collection, key[, default])
> 
> Delete and return the mapping for key if it exists in collection, otherwise return default, or throw an error if default is not specified.

And also:

> 
> Base.pop! — Method.
> 
> pop!(collection) -> item
> 
> Remove an item in collection and return it. If collection is an ordered container, the last item is returned.

You can see the definition is for indexed collections with Dict-like behavour.
This is our current behavour.

The second definition is for non-indexed collections with Stack like behavour.
(though the main user `Vector` is indexed, the index is not used).

If we think of Accumulator as being like a `DefaultDict{T, Int}`, with a default value of `0`.
Then the current behavior of resetting the count to zero makes sense.

If on the otherhand, we think of it as a Multiset / Bag (#133).
Then it makes sense for `pop!` to delete only the last copy of that item added to the bad.
i.e. reducing its count by one.

kmsquire:
@oxinabox, do you have a use case that suggests one use case over the other?

To me, this suggests that `push!` and `pop!` are the wrong functions to be using for accumulating values, or at least that it would probably be useful to have an alternative to `pop!` when treating an Accumulator like a Multiset/Bag.  

Maybe `pop_one!`?

oxinabox:
Could deprecate `push!`, and  just have `inc` and `dec`.
Maybe also deprecate `pop!` for `reset!`.
Thus thinking about this as a thing that counts -- a counter if you like :-P.

I only noticed this because I wanted to decrements the count sometimes in a for-loop.
I thought since I was using `push!` elsewhere in the loop for incrementing, `pop!` would decrement.
Right now the only way to `dec` is `push!(counter, key,  -1)`
Set index also needs overloading -- with sent index overloaded there is the nice: `counter[key]-=1`

cstjean:
I agree with @oxinabox: if `push!` increments, `pop!` should decrement. Perhaps `delete!` should be used to delete an element? It's already used this way in `dict`. I can't imagine too many uses of the current `pop!`.

[...]

Issue #248, opened 2017-01-07
OrderedDict fails to infer types of generators.


bicycle1885:
On 3afb7a24c304c67a0b3d432af89a2550333eb071, it fails to infer element types as follows:
```
julia> typeof(Dict(x => 2x for x in [1,2,3]))  # this works
Dict{Int64,Int64}

julia> typeof(OrderedDict(x => 2x for x in [1,2,3]))  # but this doesn't
DataStructures.OrderedDict{Any,Any}
```

I'm using Julia 0.5.0.

oxinabox:
This is because julia does not do inference on the types of generators.
This is basically an issue for JuliaLang/julia.
I believe there is a good reason not to.
Like that it can't always be done.
And so the behavior of the code should not depend on it.

The `eltype` for generators is not set:
```
julia> Base.iteratoreltype(typeof(x=>2x for x in [1,2,3]))
Base.EltypeUnknown()

julia> eltype(x=>2x for x in [1,2,3])
Any
```

One could force it, via something like

```
force_eltype(::Type{T}) where {T} = eltype(T) #fallback
force_eltype(gen::T) where {T} =force_eltype(T)

function force_eltype(::Type{Base.Generator{B,F}}) where{B,F}
    back_et = force_eltype(B)
    gen_ets = Base.return_types(F.instance, (back_et,))
    Union{gen_ets...}
end
```

But I'm sure there is a good reason not to.

oxinabox:
It seems we are already doing something like this with `Base._default_eltype` for accumulators.
So maybe this is Ok

cstjean:
I don't like using `Base._default_eltype` either, but if Base cheats for its `Dict` implementation, IMHO DataStructures can/should cheat too, if only to provide a consistent user experience. There are some [comments on Discourse about it](https://discourse.julialang.org/t/can-eltype-deduce-the-element-type-of-a-generator/6429/13), apparently the `eltype` of generators is a contentious topic.

oxinabox:
I agree.

laborg:
This is fixed by now (Julia 1.9.3 and DataStructures 0.18.5:
```
julia> typeof(Dict(x => 2x for x in [1,2,3]))  
Dict{Int64, Int64}

julia> typeof(OrderedDict(x => 2x for x in [1,2,3])) 
OrderedDict{Int64, Int64}
```

PR #334, opened 2017-10-20
Make counters use inc!, dec!, reset! and merge!, also adds setindex!

oxinabox:
This resolves https://github.com/JuliaCollections/DataStructures.jl/issues/303
and https://github.com/JuliaCollections/DataStructures.jl/issues/285

In particular it gets rid of `pop!` entirely for `dec!`, with a deprecation warning.
After the deprecation period is over, we can add `pop!` back for the Bag/multiset interpretation of an accumulator.
And then maybe add it as thing that all ClassifiedCollections have too,
(since then all of them will have that)

Possibly `reset!` should be named `delete!` and should have the normal `delete!` return value of the whole collection. Right now it returns the count of the deleted item. Like `pop!`

It added new exports for `inc!`, `dec!` and `reset!`.
Technically if wanted to reduce exports `inc` could stay as `push!`, and `reset!` could be come `delete!`.
But need `dec!` as otherwise can't deprecate `pop!`.

### DIFF
--- a/src/accumulator.jl
+++ b/src/accumulator.jl
@@ -26,7 +26,7 @@ end
 
 eltype_for_accumulator(seq::T) where T = eltype(T)
 function eltype_for_accumulator(seq::T) where {T<:Base.Generator}
-    @static if VERSION < v"0.7.0-DEV.2096"
+    @static if VERSION < v"0.7.0-DEV.2104"
         Base._default_eltype(T)
     else
         Base.@default_eltype(T)
--- a/src/accumulator.jl
+++ b/src/accumulator.jl
@@ -24,7 +24,7 @@ function counter(seq)
     return ct
 end
 
-eltype_for_accumulator(seq::T) = eltype(T)
+eltype_for_accumulator(seq::T) where T = eltype(T)
 function eltype_for_accumulator(seq::T) where {T<:Base.Generator}
     @static if VERSION < v"0.7.0-DEV.2096"
         Base._default_eltype(T)
--- a/src/accumulator.jl
+++ b/src/accumulator.jl
@@ -9,30 +9,22 @@ end
 Accumulator(::Type{T}, ::Type{V}) where {T,V<:Number} = Accumulator{T,V}(Dict{T,V}())
 counter(T::Type) = Accumulator(T,Int)
 
-counter(dct::Dict{T,Int}) where {T} = Accumulator{T,Int}(copy(dct))
+counter(dct::Dict{T,V}) where {T,V<:Integer} = Accumulator{T,V}(copy(dct))
 
 """
-    counter{T}(seq::AbstractArray)
+    counter(seq)
 
 Returns an `Accumulator` object containing the elements from `seq`.
 """
-function counter(seq::AbstractArray{T}) where T
-    ct = counter(T)
+function counter(seq)
+    ct = counter(eltype(seq))
     for x in seq
-        push!(ct, x)
+        inc!(ct, x)
     end
     return ct
 end
 
-function counter(gen::T) where {T<:Base.Generator}
-    ct = counter(Base._default_eltype(T))
-    for x in gen
-        push!(ct, x)
-    end
-    return ct
-end
-
-copy(ct::Accumulator{T,V}) where {T,V<:Number} = Accumulator{T,V}(copy(ct.map))
+copy(ct::Accumulator) = Accumulator(copy(ct.map))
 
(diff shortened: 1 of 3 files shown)


src/accumulator.jl
```diff
@@ -9,30 +9,22 @@ end
 Accumulator(::Type{T}, ::Type{V}) where {T,V<:Number} = Accumulator{T,V}(Dict{T,V}())
```
kmsquire: As with other constructors, we should deprecate this constructor, and add a constructor for `Accumulator{T,V<:Integer}()`.

Probably not that urgent, as the primary interface to this seems to be the `counter` function.  Can be done in a separate PR (unless you want to do it here).

src/accumulator.jl
```diff
@@ -9,30 +9,22 @@ end
...
+    ct = counter(eltype(seq))
     for x in seq
-        push!(ct, x)
+        inc!(ct, x)
     end
     return ct
 end
 
-function counter(gen::T) where {T<:Base.Generator}
-    ct = counter(Base._default_eltype(T))
```
kmsquire: This code was just added recently (by @cstjean).  I'm going to suggest leaving it in until the `eltype` of a generator can be more easily determined.  @cstjean, any thoughts?

src/accumulator.jl
```diff
@@ -61,32 +56,80 @@ done(ct::Accumulator, state) = done(ct.map, state)
...
+"""
+    dec!(ct, x, [v=1])
 
-function push!(ct::Accumulator, r::Accumulator)
-    for (x, v) in r
-        push!(ct, x, v)
+Decrements the count for `x` by `v` (defaulting to one)
+"""
+dec!(ct::Accumulator, x, a::Number) = (ct[x] -= a)
+dec!(ct::Accumulator{T,V}, x) where {T,V} = dec!(ct, x, one(V))
```
kmsquire: Okay to leave as is, but I think this can just be 

```julia
dec!(ct::Accumulator, x, a::Number = 1) = (ct[x] -= a)
```

I guess it might be worth looking at the code generated for each of these in different cases, but in general, until it is shown to be better to create more complicated code, I think simpler code is better.

src/accumulator.jl
cstjean: Is there a reason to take it out? It provides much better typing of the accumulator.

oxinabox: oop, I actually  though I had it covered by adding general iterators.
I didn't know `_default_eltype` was different to `eltype`, but now I see it is different.

With that said:
I wonder why Base doesn't have `eltype(T::Type{Generator})=_default_eltype(T)`.
The fact that Base does not do this leaves me worried.
It is an undocumented, unexported method,
that seems like it should be used and yet, it is not.

I will put that back in.
This is closely related to  #248.

oxinabox:
@kmsquire  thanks.
I'll get to those changes.
What are you thoughts on the API changes the go with this?

kmsquire:
> What are your thoughts on the API changes the go with this?

Kind of ambivalent, which is why I think I didn't review this when you first pushed it.  😜

Generally, I think having a uniform API between different data structures is good (which is probably why it was `push!` and `pop!` before, although I didn't write this code).  But it doesn't always fit, so I think that your plan to change these (at least temporarily) to `inc!` and `dec!` are fine.

Regarding `reset!` vs `delete!` (and the corresponding return value), it would be good to have some sample code in order to see what makes the most sense.  Since you've taken the most interest, I'm going to defer to your judgment here.

Thanks again, for working on cleaning things up here.

oxinabox:
I have made the changed to bring back the the eltype forcing.
In a 0.7 compatible way.
Which should unbreak like all recent PRs that are broken.
So I'ld like to merge this soon so I can prod them.
Since other that that it was passed review If I don't get any further feedback by tonight (~10 hrs).
I'll go ahead and merge.
(but not tag a release)

kmsquire:
The unsuccessful appveyor builds were caused by missing julia 0.7 windows downloads

See 
* https://ci.appveyor.com/project/kmsquire/datastructures-jl/build/1.0.230/job/eba59leatyxr4a9p
* https://discourse.julialang.org/t/travis-download-of-julia-latest-winxx-exe-broken/7045

Coverage decrease is also spurious (`@static` line isn't covered).

See 
* https://coveralls.io/builds/14165929/source?filename=src%2Faccumulator.jl#L29
* https://github.com/JuliaCI/Coverage.jl/issues/139

Merging

Issue #339, opened 2017-11-07
Implement unrolled linked lists


PallHaraldsson:
I proposed at https://github.com/adolgert/Lists.jl/issues/1

I.e. https://en.wikipedia.org/wiki/Unrolled_linked_list

See e.g.:
http://www.sanfoundry.com/java-program-implement-unrolled-linked-list/

And got this response:

"That seems like a great idea."

I couldn't find the issue here; I thought I had filed here, and see now the other project seems dead.

oxinabox:
Isn't this what our Deque is?
(honest question)

https://github.com/JuliaCollections/DataStructures.jl/blob/master/src/deque.jl

kmsquire:
Yes, that implementation is pretty much equivalent.

PallHaraldsson:
"Isn't this what our Deque is?" Ok, great then, I didn't realize as a deque can be implemented another way, e.g. non-unrolled list. If you can really do all the same things as with a list, e.g. add or drop from any position, then this issue can be closed; but maybe document then better that it can work as/is an unrolled linked-list.

PallHaraldsson:
There's also at least on generalization: https://en.wikipedia.org/wiki/T-tree

I see 2-3 tree in the code; is there no 3+ tree in the code, and isn't it better to have? And deque only blocked/unrolled list? I so on WP:

the B-tree and T-tree, data structures that are similar to unrolled linked lists in the sense that each of them could be viewed as an "unrolled binary tree"

kmsquire:
> If you can really do all the same things as with a list, e.g. add or drop from any position,

The underlying `DequeBlock` data structure is [here](https://github.com/JuliaCollections/DataStructures.jl/blob/master/src/deque.jl#L8-L23), and the corresponding `Deque` implementation is below that.

`DequeBlock` and `Deque` could, in theory, be generalized to add or drop from any position, but there is no functionality to do so right now.  Some comments:

1.  Arguments were made by @lindahua, the original author of this code, that, for `Deque` (and `Stack` and `Queue`), there is expected functionality (push/pop/enqueue/dequeue) for each data structure, and that it's better to keep the implementations simple, in order to keep them performant.  One could argue differently of course.
   (This is a paraphrase from some old conversations here that I don't want to take the time to find right now, so @lindahua, please speak up if I'm misrepresenting you!)

[...]

PR #341, opened 2017-11-09
Make "unrolled link list" clear, and link to Wikipedia [ci skip]

PallHaraldsson:


PallHaraldsson:
At least since/if not usable as such a list, a first step to complete https://github.com/JuliaCollections/DataStructures.jl/issues/339.

PallHaraldsson:
Seems like I fixed from .md to .rst syntax (I learned something, not why both exist, or why here changed to this format, from MD, in the past..). Adding [ci skip] did probably nothing.. but anyway seems ok to just merge.

oxinabox:
Thanks.
I think it is good to have these summeries in the readme.
It would be nice to have them in the full docs too.

I will merge once travis etc completes.
The `[ci skip]` needs  to be in the commit message, not the PR title.
(unfortunately)

PallHaraldsson:
"I will merge once travis etc completes."

You mean you can't? I can't merge either way, I thought just a permission problem.

I'm not sure why CI is so slow, at least here where no actual code change.. but I'm in no hurry.

kmsquire:
As noted in https://github.com/JuliaCollections/DataStructures.jl/issues/339#issuecomment-343750565, the underlying `DequeBlock` implementation is very similar to an unrolled linked-list, but access to/insertion to/deletion from arbitrary positions isn't implemented.

PR #342, opened 2017-11-09
Doc: "unrolled linked list" change [ci skip]

PallHaraldsson:


PallHaraldsson:
I could add "readme" to this (the similar one already pending). Or just wait for it?

PallHaraldsson:
Hmm, I remember [ci skip] working before, if I start with it..

oxinabox:
just keep them separate.
Weird that travis seems to still be trying to run even with the [ci skip]

oxinabox:
thanks

kmsquire:
As I mentioned in the other issue, while the underlying implementation for `Dequeue`s is similar to an unrolled linked list, I don't think this PR is actually accurate.  (Sorry, I thought I mentioned that here as well...)

oxinabox:
Yeah I saw that . It was my thought that while not perfectly accurate, it was more accurate than not, and so better than not having this .
Sorry if I over stepped.
I can either fix the docs to be "implemented similar to a unrolled linked list",
or make a PR or issue to fix it so it **is** an unrolled linked list.

kmsquire:
👍 Either way works for me.  Whatever you have time for.

PR #344, opened 2017-11-12
delete! for PriorityQueue

### LINKED ISSUE #343: Implement delete!(pq::PriorityQueue, key)
Hello,

I need to be able to remove a key from a `PriorityQueue`.
I noticed that

```julia
delete!(pq::PriorityQueue, key)
```

is not implemented. Maybe it could be useful for some other users

I haven't look closely at `PriorityQueue` code, so please excuse my ignorance about its internal mechanism.

But I think I need to remove key from `pq.index` `Dict` but also remove a `Pair` from  `pq.xs`

What is your opinion about?

```julia
import Base: delete!
function delete!(pq::PriorityQueue, key)
    idx = pq.index[key]
    delete!(pq.index, key)
    deleteat!(pq.xs, idx)
    pq
end
```

Kind regards

scls19fr:
Closes #343


oxinabox:
CI is being bad, but this LGTM.
I can merge this if you are happy with it
(and we changed the settings so  we don't need CI to pass first)

kmsquire:
@oxinabox, it's better if we let CI pass first.  While they're not perfect, they're there for a reason--to make sure as much as possible works!  If something is broken about CI, it should ideally be fixed, not ignored.

In some cases, of course, waiting for CI to pass or fixing problems is not reasonable or possible for any of us (e.g., the latest Julia Windows builds haven't been updated, so the Julia 0.7 Appveyor tests are failing).  

Anyway, I just restarted the failing job on Travis, which seems to have been a timeout.  If that succeeds, it also might fix the coverage issues.

oxinabox:
I agree, and was likewise waiting. (I thought I kicked Travis to retry)
The comment was mores an an aside.

We should actually be passing on AppVeyor even with the old 0.7 builds as the version conditional I added to accumulators (which was what was broken) is linked to the exact build number.
THough it seemed to me recent AppVeyor failures have been downloading julia binary issues.
Which were in particular the CI tests I was thinking of when I mentioned bypassing them.

scls19fr:
Personally, I won't have fixed doctest now.
I will first tackle Documenter.jl / doctest issue and after tackle this one

src/priorityqueue.jl
```diff
@@ -324,6 +324,29 @@ function next(pq::PriorityQueue{K,V}, i) where {K,V}
     return (pq.xs[idx], i)
 end
 
+"""
+    delete!(pq, key)
+
+Delete the mapping for the given key in a priority queue, and return the priority queue.
+# Examples
+```jldoctest
+julia> q = PriorityQueue(["a","b","c"],[2,3,1],Base.Order.Forward)
```
kmsquire: `PriorityQueue`s no longer accept two iterables during construction (matching the corresponding change in other data structures like `Dict`s).  Some alternatives include

```
q = PriorityQueue(Base.Order.Forward, "a"=>2, "b"=>3, "c"=>1)
q = PriorityQueue(Base.Order.Forward, zip(["a","b","c"],[2,3,1]))
q = PriorityQueue(zip(["a","b","c"],[2,3,1]), Base.Order.Forward)
q = PriorityQueue(Base.Order.Forward, [("a", 2), ("b", 3), ("c", 1)])
```

(The same change needs to be made in the docstring for the `struct`, if you're up for it.)

test/test_priority_queue.jl
```diff
@@ -176,6 +176,12 @@ end
     @test dequeue_pair!(pq) == ('e'=> 1)
     @test dequeue_pair!(pq, 'b') == ('b'=>4)
     @test length(pq) == 3
+
+    # delete!
+    pq = PriorityQueue(["a","b","c"],[2,3,1],Base.Order.Forward)
```
kmsquire: Same change needed here.

scls19fr: That's sad to see that https://github.com/JuliaDocs/Documenter.jl/issues/527 is not fixed!
We should have been able to catch this kind of error!

scls19fr: Maybe some people at JuliaCon hackaton should try to fix thiss first because that's very annoying not to be able to rely on doctests!

scls19fr:
Closed in favor of #428

PR #349, opened 2017-12-05
Use KeySet instead of KeyIterator

martinholters:
...in argument types of `in` and `next` on newer Julia versions.

Admittedly, I don't know whether the `KeySet` interface promises anything that DataStructures currently doesn't hold, but at least this makes DataStructures load again on Julia 0.7.


oxinabox:
LGTM,
what about `SortedDict` ?

StephenVavasis:
I haven't understood what this thread is about.  What is the issue or problem for `SortedDict`?

oxinabox:
It may or may not be a  a problem for `SortedDict`.
But Base has changed the name of `KeyIterator` to `KeySet`.
So if `keys(::StortedDict)` was returning a `KeyIterator` then it would need to be updated

But actually I see that `keys(::SortedDict)` returns its own type:
`DataStructures.SDMKeyIteration`
so I guess it does not need to be changed.

So I think this can be merged now

martinholters:
Worth tagging a new release to have a version working on current Julia master?

PR #350, opened 2017-12-08
Remove unnecessary `global` declaration in IntSet test

martinholters:
Fixes the
```
WARNING: deprecated syntax "implicit assignment to global variable `x`".
Use "global x" instead.
```
when running the tests.


kmsquire:
Thanks!

PR #351, opened 2017-12-15
Associative -> AbstractDict

ararslan:


src/DataStructures.jl
```diff
@@ -54,6 +54,10 @@ module DataStructures
 
     import Base: eachindex, keytype, valtype
 
+    if VERSION < v"0.7.0-DEV.2951"
```
timholy: Perhaps better in Compat.jl?

oxinabox:
Closing in favour of #353 which uses Compat and does a few other things.

PR #352, opened 2017-12-18
Add argument to Default*Dict to tell it to pass the key to the default function

iamed2:
I find I want the default value to be some function on the key pretty often. Motivating example added to the docs.

Is there a better name than `passkey`?

~EDIT: Oops, still need to add a few tests~

### DIFF
--- a/test/test_default_dict.jl
+++ b/test/test_default_dict.jl
@@ -15,6 +15,10 @@ import DataStructures: DefaultDictBase
     ddb = DefaultDictBase{Int, Float64}(0.0)
     @test isa(ddb, DefaultDictBase{Int, Float64, Float64, Dict{Int,Float64}})
     @test isa(DefaultDictBase(1.0, ddb), DefaultDictBase{Int, Float64, Float64, Dict{Int,Float64}})
+
+    @test isa(DefaultDictBase{Int, String}(String), DefaultDictBase{Int, String, typeof(String), Dict{Int, String}})
+    @test isa(DefaultDictBase{Int, String}(String; passkey=true), DefaultDictBase{Int, String, typeof(String), Dict{Int, String}})
+    @test isa(DefaultDictBase{Int, String}(String; passkey=false), DefaultDictBase{Int, String, typeof(String), Dict{Int, String}})
 end
 
 ##############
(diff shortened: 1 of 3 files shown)


kmsquire:
I think this is fine. 

If we’re worried about the extra test, `passkey` could be a type parameter. But probably not worth the trouble. 

Failure on nightly is unrelated (but needs to be addressed separately.

oxinabox:
I agree good feature.
This is actually why I use normal `Dict`, and `get!` rather than a `DefaultDict` 

I kind of think this would be nice to have default as `passkey=true`.
its would be breaking, and I can't see a sensible deprecation path.
We could push this version tag a release, change to `passkey=true` tag a breaking release.  
It would be crazy breaking.
But julia 0.7 is going to be crazy braking anyway.
So it is a good time for it.

--------

Is it possible to pretend we had arrow types so that we pass the key if the function has a 1 arg form, 
and we don't if it doesn't.

Like using reflection: `methodexists(f, (Any,))`,
or via catching method errors?

Possibly we could do this as a path to having a deprecation, even if it is slow.

------------

Overall I'ld be fairly happy to merge as is (I'll just do a code review),
but I wanted to throw these thoughts out first.

doc/source/default_dict.rst
```diff
@@ -59,6 +61,22 @@ Examples using ``DefaultDict``::
   # inner set of DefaultDicts is zero
   dd = DefaultDict(AbstractString, DefaultDict, () -> DefaultDict(AbstractString,Int,0))
 
+  # use DefaultDict to cache an expensive function call
```
oxinabox: Could link to wikipedia on Memoization

test/test_default_dict.jl
```diff
@@ -92,6 +96,33 @@ end
     @test e['e'] == 9
     @test f['e'] == 0
 
+    # passkey
```
oxinabox: this is a good time to use a nexted testset, rather than a comment delimited block.
We don't use them enough in DataStructures.jl.
I assume because they didn't used to exist.

Arguably, it would be weird to start now.
And better to go and clean everything at once.
But idk that that is a good argument.

test/test_default_dict.jl
```diff
@@ -92,6 +96,33 @@ end
...
+    @test calls == 1
+    @test length(g) == 1
+    @test g["baz"] == 3
+    @test calls == 2
+    @test length(g) == 2
+    @test g["foobar"] == 6
+    @test calls == 2
+    @test length(g) == 2
+
+    # passkey but wrongly
```
oxinabox: this too

kmsquire:
While I like the idea, I’ve never felt the need to pass in the key to the default function, and I use `DefaultDict`s pretty frequently. So I personally prefer to leave `passkey=false`. (That also matches Python’s `defaultdict`, which was the inspiration for this version... although that’s less important.)

iamed2:
Back from vacation, updating this now :)

oxinabox:
Why didn't I merge this 2 months ago, I went looking for this feature today.

PR #353, opened 2017-12-22
Void -> Nothing and Associative -> AbstractDict

fredrikekre:


fredrikekre:
Oh, also done in #351

oxinabox:
LGTM, 
I has a quick check and I can't see any docs that need to change.
So I am going to merge this, once tests pass.
If there is something I missed, we can make another PR

fredrikekre:
There was a recent `symdiff` change on julia master leading to this, but I think it is fine to merge this anyways and fix that in a separate PR.
```
OrderedSet: Test Failed at /home/travis/.julia/v0.7/DataStructures/test/test_ordered_set.jl:199
  Expression: isequal(symdiff(OrderedSet([1, 2, 3, 4]), OrderedSet([2, 4, 5, 6])), OrderedSet([1, 3, 5, 6]))
   Evaluated: isequal([1, 3, 5, 6], OrderedSet{Int64}([1, 3, 5, 6]))
```

oxinabox:
yes, I am happy with that, (and I don't think we need to wait for the mac builds to finish (infact I think we should disable those, as they are slow and uninformative)).
So merging now

Issue #345, opened 2017-11-12
Continuous integration should break when doctests are broken


s-celles:
Hello,

I did a (voluntary) mistake in https://github.com/JuliaCollections/DataStructures.jl/pull/344/files#diff-dc614dd5c3fe25430697f06e62e01395R344

But Travis shows a green badge
https://travis-ci.org/JuliaCollections/DataStructures.jl/jobs/301055719

I think, continuous integration should break when doctests are broken

Kind regards

oxinabox:
To do it we'ld need to add Documentor.jl to the Requires (right?).
Which I think is Ok.
So this just needs to be done

scls19fr:
I don't think adding Documentor to `REQUIRE` is enough. 
I don't even know if that's necessary as Sphinx is used for building documentation

oxinabox:
What i mean to say is we'ld need to migrate from Sphinx to Documentor.jl.

scls19fr:
Ok so that's a separate issue

kmsquire:
> What i mean to say is we'ld need to migrate from Sphinx to Documentor.jl.

Commented in #343 before I saw this, (and just saw #346), but yes, this is what is needed.

scls19fr:
Related Documenter issue https://github.com/JuliaDocs/Documenter.jl/issues/527

scls19fr:
Looks quite similar than https://github.com/JuliaReach/LazySets.jl/issues/71 which was addressed using https://github.com/JuliaReach/LazySets.jl/pull/467

PR #361, opened 2018-01-05
[WIP] Fix doctests

s-celles:
Should help #345

Break willingly a doctest to be sure that doctests are running

If that's not the case ie CI is passing despite a broken doctest, we should fix the testing machinery to ensure that doctests break as expected.

After, we should fix the doctest I willingly broke and CI should pass correctly


scls19fr:
https://travis-ci.org/JuliaCollections/DataStructures.jl/jobs/325394206 shows that CI is passing... which is not a good news... it means that doctests are not run.

PR #368, opened 2018-01-14
Allow failures for CI with Julia Nightly

scls19fr:


oxinabox:
I think this is for the best, as it makes the CI more informative.

oxinabox:
Thanks, I forgot about this and should have merged it a month ago. 
My bad

PR #372, opened 2018-02-17
Add implementations of pop! and shift! to CircularBuffer

tjgreen:
Previously we had only `push!` and `unshift!`; this PR adds the corresponding deletion operations.  I also modified the existing code to allocate the entire underlying buffer up-front, and to introduce an explicit length field.  This ensures that `shift!` and `unshift!` run in O(1) time, rather than O(n).

I also noticed that most accesses to `cb.buffer` could safely be annotated `@inbounds`, since the `CircularBuffer` code already performs its own bounds-checking, but I'll save that for another PR.

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -15,14 +15,14 @@ function Base.empty!(cb::CircularBuffer)
     cb
 end
 
-function _buffer_index(cb::CircularBuffer, i::Int)
+function _buffer_index_checked(cb::CircularBuffer, i::Int)
     if i < 1 || i > cb.length
         throw(BoundsError(cb, i))
     end
-    _mod_buffer_index(cb, i)
+    _buffer_index(cb, i)
 end
 
-function _mod_buffer_index(cb::CircularBuffer, i::Int)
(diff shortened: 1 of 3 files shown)


src/circular_buffer.jl
```diff
@@ -4,21 +4,26 @@ New items are pushed to the back of the list, overwriting values in a circular f
...
-    n = length(cb)
-    if i < 1 || i > n
-        throw(BoundsError("CircularBuffer out of range. cb=$cb i=$i"))
+    if i < 1 || i > cb.length
+        throw(BoundsError(cb, i))
     end
+    _mod_buffer_index(cb, i)
+end
+
+function _mod_buffer_index(cb::CircularBuffer, i::Int)
```
oxinabox: I don't like this function name, thre must be a better name than that.


Maybe `_buffer_index`, and then the current `_buffer_index` could be renamed `_buffer_index_checked`

tjgreen: I agree, `_buffer_index_checked` and `_buffer_index` would be better names.

oxinabox:
LGTM, thanks.

PR #374, opened 2018-03-20
Fixed keys() and findfirst()

dfdx:
Comments for fixes: 

 * unlike `_default_eltype` function, macro `@default_eltype` accepts iterator, not type
 * since `keys()` for `OrderedSet` should mimic array indexing, it's keys should be integers in range `1:length(s)`, not keys of an underlying dict, which are just elements of the set

There's still a couple of deprecation warnings, but tests pass locally.

### DIFF
--- a/test/test_ordered_set.jl
+++ b/test/test_ordered_set.jl
@@ -138,7 +138,11 @@
     s = OrderedSet([1,3,5,7])
     @test findfirst(isequal(1), s) == 1
     @test findfirst(isequal(7), s) == 4
-    @test findfirst(isequal(2), s) == 0
+    if VERSION >= v"0.7.0-DEV.3399"
+        @test findfirst(isequal(2), s) == nothing
+    else
+        @test findfirst(isequal(2), s) == 0    
+    end    
 
     # setdiff
     @test isequal(setdiff(OrderedSet([1,2,3]), OrderedSet()),        OrderedSet([1,2,3]))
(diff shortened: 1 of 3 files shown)


dfdx:
Since behavior for `findfirst` and `symdiff` changed in Julia 0.7, tests now fail on 0.6. @ararslan is `aa/0.7` branch intended to be merged into master and support 0.6 as well (in which case I'll make separate test cases) or are you going to make this branch 0.7 only?

ararslan:
>  is aa/0.7 branch intended to be merged into master and support 0.6 as well

Yes, I intended to maintain support for 0.6 as well.

test/test_ordered_set.jl
```diff
@@ -138,7 +138,7 @@
     s = OrderedSet([1,3,5,7])
     @test findfirst(isequal(1), s) == 1
     @test findfirst(isequal(7), s) == 4
-    @test findfirst(isequal(2), s) == 0
+    @test findfirst(isequal(2), s) == nothing
```
ararslan: This should work on 0.6 if you do
```julia
if VERSION >= v"0.7.0-DEV.3399"
    @test findfirst(isequal(2), s) == nothing
else
    @test findfirst(isequal(2), s) == 0    
end
```
which should eventually be cleaned up automatically by FemtoCleaner, or if you prefer simplicity:
```julia
@test coalesce(findfirst(isequal(2), s), 0) == 0
```

dfdx: I like the explicit version + FemtoCleaner, at least it's a  very clear message for a causal reader of this snippet. 

As for `symdiff`, I believe the change was introduce in [this commit](https://github.com/JuliaLang/julia/commit/84da967df4aa38b6b29de11dcf861ddf64139463), though I need to figure out exact Julia version for it. When you need to find such a version, do you just rebuild Julia or there's a smarter way to do it?

ararslan: In the Julia source tree, there's a super cool shell script called `commit-name.sh`. If you know the commit SHA, you can get the corresponding version like so:
```
$ contrib/commit-name.sh 84da967df4aa38b6b29de11dcf861ddf64139463
0.7.0-DEV.4377
```

dfdx: Oh, that's much easier than rebuilding Julia which I've been doing the whole day! :) However, this behavior turns to have been introduced somewhere earlier, I'll continue to dig into it tomorrow.

(As plan B we can simply stick to 0.7.0-DEV.4377 since it's guaranteed to work and many commits between it and Julia 0.6 turn to be broken anyway).

ararslan: My guess is that the `symdiff` change happened in https://github.com/JuliaLang/julia/pull/23528, i.e. 0.7.0-DEV.3127. We can just go with that, it's probably close enough even if it isn't exactly right.

dfdx:
Fixed tests on both - 0.6 and 0.7.

ararslan:
Perfect, thanks so much!

PR #375, opened 2018-03-23
add most_common for accumulator

oxinabox:
This is basically #198 but with select and sort.
And up to date and without conflicts.

I left off optimising the `most_common(A, 1)` case, because that optimization should be done to
 `select` in Base.
I think

### DIFF
--- a/src/accumulator.jl
+++ b/src/accumulator.jl
@@ -165,7 +165,7 @@ julia> nlargest(counter("abbbccddddda"),2)
 ```
 """
 nlargest(acc::Accumulator) = sort!(collect(acc), by=last, rev=true)
-nlargest(acc::Accumulator, n) = select!(collect(acc), 1:n, by=last, rev=true)
+nlargest(acc::Accumulator, n) = @compat partialsort!(collect(acc), 1:n, by=last, rev=true)
 
 
 """
@@ -179,7 +179,7 @@ For obvious reasons this will not include zero counts for items not encountered.
 (unless those elements are added to he accumulator directly, eg via `acc[foo]=0)
 """
 nsmallest(acc::Accumulator) = sort!(collect(acc), by=last, rev=false)
-nsmallest(acc::Accumulator, n) = select!(collect(acc), 1:n, by=last, rev=false)
+nsmallest(acc::Accumulator, n) = @compat partialsort!(collect(acc), 1:n, by=last, rev=false)
 
 
 
--- a/src/accumulator.jl
+++ b/src/accumulator.jl
@@ -1,4 +1,4 @@
-# A counter type
+#A counter type
 
 struct Accumulator{T, V<:Number} <: AbstractDict{T,V}
     map::Dict{T,V}
@@ -138,14 +138,16 @@ Returns its former count.
 reset!(ct::Accumulator, x) = pop!(ct.map, x)
 
 """
-     most_common(acc::Accumulator, [k])
+     nlargest(acc::Accumulator, [n])
 
-Returns a sorted vector of the `k` most common elements, with their counts.
-If `k` is ommitted, the full sorted collection is returned.
+Returns a sorted vector of the `n` most common elements, with their counts.
+If `n` is omitted, the full sorted collection is returned.
+
+This corresponds to Python's `Counter.most_common` function.
 
 Example
 ```
-julia> most_common(counter("abbbccddddda"))
+julia> nlargest(counter("abbbccddddda"))
 
 4-element Array{Pair{Char,Int64},1}:
  'd'=>5
@@ -154,7 +156,7 @@ julia> most_common(counter("abbbccddddda"))
  'a'=>2
 
 
-julia> most_common(counter("abbbccddddda"),2)
+julia> nlargest(counter("abbbccddddda"),2)
 
 2-element Array{Pair{Char,Int64},1}:
  'd'=>5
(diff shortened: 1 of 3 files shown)


nalimilan:
Maybe it would be better to make this a method of `partialsort` (which was `select` in 0.6)? The one-argument method could then be just `sort`.

oxinabox:
While yes, as you can see in the code, these are basically just wrappers of `sort` and `select` with defaults set (which are the opposite of the normal defaults).

To me calling it `sort` would feel wrong.
I think `sort` should almost always be `sort(x::T)::T where T`
That is to say, sorting something of type `T` should give a result of type `T` 

`most_common` is the name used in python.
And just matching python was what i was going for.

If  we wanted to be a bit free-er,
I think not having it at all,  and just implementing it on top of an `prioryqueue` might be nicer.

nalimilan:
> To me calling it sort would feel wrong.
> I think sort should almost always be sort(x::T)::T where T

Since this isn't possible for data structures without an ordering, it makes sense to me to return a different structure in that case, be it a vector of pairs or an ordered dict.

oxinabox:
My instinct disagrees.
Since this isn't possible for data structures without an ordering, it makes sense **to me** to use a different function name in that case. e.g. `most_common`.
but we could go either way.

To be a bit more clear on my instincts (intraspecting here),
I think I expect the return type to duck-type as if it were a sorted version of the the input type.
E.g.  `sort(::Dict)::OrderedDict` feels fine to me.

src/accumulator.jl
```diff
@@ -137,6 +137,34 @@ Returns its former count.
...
+
+julia> most_common(counter("abbbccddddda"),2)
+
+2-element Array{Pair{Char,Int64},1}:
+ 'd'=>5
+ 'b'=>3
+
+```
+"""
+most_common(acc::Accumulator) = sort(collect(acc), by=last, rev=true)
```
ararslan: You should be able to `sort!` instead of `sort`, assuming `collect` is not a no-op.

src/accumulator.jl
```diff
@@ -137,6 +137,34 @@ Returns its former count.
 """
 reset!(ct::Accumulator, x) = pop!(ct.map, x)
 
+"""
+     most_common(acc::Accumulator, [k])
+
+Returns a sorted vector of the `k` most common elements, with their counts.
+If `k` is ommitted, the full sorted collection is returned.
```
ararslan: omitted (one m)

test/test_accumulator.jl
```diff
@@ -112,33 +112,44 @@
...
+	@testset "most common" begin
+		@test most_common(counter("abbbccddddda")) == ['d'=>5, 'b'=>3, 'a'=>2, 'c'=>2]
+		@test most_common(counter("abbbccddddda"),2) == ['d'=>5, 'b'=>3]
+		@test most_common(counter("a")) == ['a'=>1]
 
-    #deprecations
-    ctd = counter([1,2,3])
-    @test ctd[3]==1
+		@test_throws BoundsError most_common(counter("a"),2)
+	end
```
kmsquire: Tabs -> spaces.

src/accumulator.jl
```diff
@@ -137,6 +137,34 @@ Returns its former count.
...
+julia> most_common(counter("abbbccddddda"),2)
+
+2-element Array{Pair{Char,Int64},1}:
+ 'd'=>5
+ 'b'=>3
+
+```
+"""
+most_common(acc::Accumulator) = sort(collect(acc), by=last, rev=true)
+most_common(acc::Accumulator, k) = select(collect(acc), 1:k, by=last, rev=true)
```
kmsquire: Should probably change `select` to `Compat.partialsort!`.

test/test_accumulator.jl
```diff
@@ -112,33 +112,44 @@
...
-    # non-integer uses
-    acc = Accumulator(Symbol, Float16)
-    acc[:a] = 1.5
-    @test acc[:a] ≈ 1.5
-    push!(acc, :a, 2.5)
-    @test acc[:a] ≈ 4.0
-    dec!(acc, :a)
-    @test acc[:a] ≈ 3.0
+    @testset "non-integer uses" begin
+        acc = Accumulator(Symbol, Float16)
```
kmsquire: Not related to your changes, but let's deprecate this constructor, and add

```
Accumulator{T,V}() where {T,V<:Number} = Accumulator{T,V}(Dict{T,V}())
```

Not necessary in this PR, but it should be a three line change, so if you have time...

nalimilan:
Ah, since `nlargest` and `nsmallest` already exist, I agree it makes sense to use these functions.

oxinabox:
I agree also.
I will make these changes

oxinabox:
I have not done the deprecation of `Accumulator(::DataType,::DataType)`
but I have done everything else.

The docs still talk about counts because I can't think of other words to use

oxinabox:
Thoughts?

oxinabox:
Bump

oxinabox:
I am going to rebase this and if none has coments before I do so and tests all pass, merge it

src/accumulator.jl
```diff
@@ -165,7 +165,7 @@ julia> nlargest(counter("abbbccddddda"),2)
 ```
 """
 nlargest(acc::Accumulator) = sort!(collect(acc), by=last, rev=true)
-nlargest(acc::Accumulator, n) = select!(collect(acc), 1:n, by=last, rev=true)
+nlargest(acc::Accumulator, n) = @compat partialsort!(collect(acc), 1:n, by=last, rev=true)
```
ararslan: This doesn't need `@compat`; the function is available upon `using Compat`.

oxinabox:
@ararslan Fixed,
I (or you) will (can) merge if nothing further

ararslan:
Should be squashed on merge.

PR #376, opened 2018-03-24
Update wording in accumulator docs

ludvigk:


oxinabox:
These are not docs you have updated, these are just comments.

oxinabox:
Yes, this is an improvement.
Thanks

PR #373, opened 2018-03-19
Fix 0.7 depwarns

ararslan:

PR #378, opened 2018-04-20
Use deprecated bindings from Base for deprecations

### LINKED ISSUE #377: UndefVarError: find not defined
I did a `Pkg.update()` today and now get the "find not defined" error when running this in a jupyter notebook: 
```julia
Pkg.status("DataStructures")
 - DataStructures                0.8.0
```

```julia
using DataStructures
find

WARNING: both DataStructures and Base export "find"; uses of it in module Main must be qualified
UndefVarError: find not defined
```

I also checked the doc:
```julia 
?find
search: find findn findnz findin findmin findmax findkey findprev findnext

Couldn't find find
Perhaps you meant find, findn, bind, sind, findin, findnz, fill, filt, fd or fld

No documentation found.
Binding find does not exist.
```
which seems also strange.

When running the same in the REPL it works:
```julia
julia> find
find (generic function with 10 methods)
 ```
 so it might be related to jupyter, but... with release 0.7.4 the same code in jupyter works fine.

ararslan:
Fixes #377. Ref #373.

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -143,12 +143,21 @@ module DataStructures
         PriorityQueue(o, zip(ks,vs))
     end
 
-    @deprecate find(x::Union{SortedDict,SortedSet}, k) findkey(x, k)
-    @deprecate endof(x::OrderedSet) lastindex(x)
-    for T in [:IntSet, :Deque, :CircularDeque]
-        @eval @deprecate shift!(x::$T) popfirst!(x)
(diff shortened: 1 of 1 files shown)


ararslan:
Actually, this introduces method overwrite warnings because Compat refers to the same functions.

ararslan:
Should be all good now.

ararslan:
Looks like we don't have AppVeyor fast-fail enabled here, so the queue will take a while...

PR #383, opened 2018-05-21
Sorted containers -> new iteration protocol

StephenVavasis:
On branch newiterationsortedcontainers
 Changes to be committed:
	modified:   ../docs/src/sorted_containers.md
	modified:   DataStructures.jl
	modified:   balanced_tree.jl
	modified:   container_loops.jl
	modified:   ../test/test_sorted_containers.jl

(introduced in 0.7.0-DEV). It should be backwards compatible with 0.6.2.

function when applied to subranges (i.e.,  inclusive(a,b,c) or
of the length of the subrange.  (There should be no value returned

 - IntSet was changed to BitSet (name change in 0.7.0-DEV)
 - Small updates to documentation
 - Some assert statements in balanced_tree.jl that were present

### DIFF
--- a/docs/src/sorted_containers.md
+++ b/docs/src/sorted_containers.md
@@ -71,7 +71,7 @@ then there may be a loss of performance compared to:
 k,v = deref((sc,st))
 ```
 
-because the former needs an extra heap allocation step for `tok`.
+because the former may need an extra heap allocation step for `tok`.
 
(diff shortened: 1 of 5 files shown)


StephenVavasis:
Sorry... please ignore the PR for a bit... I didn't test it properly.  (Git/Pkg fooled me into thinking I was testing the branch, but I wasn't.)

PR #384, opened 2018-05-22
Sorted containers -> new iteration protocol

StephenVavasis:
On branch newiterationsortedcontainers
 Changes to be committed:
	modified:   ../docs/src/sorted_containers.md
	modified:   DataStructures.jl
	modified:   balanced_tree.jl
	modified:   container_loops.jl
	modified:   ../test/test_sorted_containers.jl

(introduced in 0.7.0-DEV). It should be backwards compatible with 0.6.2.

function when applied to subranges (i.e.,  inclusive(a,b,c) or
of the length of the subrange.  (There should be no value returned

 - IntSet was changed to BitSet (name change in 0.7.0-DEV)
 - Small updates to documentation
 - Some assert statements in balanced_tree.jl that were present

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 doc/build
 docs/build/
 docs/site/
+.gitignore
+*~
+
--- a/docs/src/sorted_containers.md
+++ b/docs/src/sorted_containers.md
@@ -71,7 +71,7 @@ then there may be a loss of performance compared to:
 k,v = deref((sc,st))
 ```
 
-because the former needs an extra heap allocation step for `tok`.
+because the former may need an extra heap allocation step for `tok`.
 
 The notion of token is similar to the concept of iterators used by C++
 standard containers. Tokens can be explicitly advanced or regressed
@@ -406,8 +406,8 @@ past-end token. Time: O(1)
 ## Iteration Over Sorted Containers
 
 As is standard in Julia, iteration over the containers is implemented
-via calls to three functions, `start`, `next` and `done`. It is usual
-practice, however, to call these functions implicitly with a for-loop
+via calls to the function `Base.iterate`. It is usual
+practice, however, to call this function implicitly with a for-loop
 rather than explicitly, so they are presented here in for-loop notation.
 Internally, all of these iterations are implemented with semitokens that
 are advanced via the `advance` operation. Each iteration of these loops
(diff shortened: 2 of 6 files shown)


StephenVavasis:
The CI tests on my pull request are failing because of the ambiguity-detector is detecting ambiguities in 0.7.  However, these ambiguities are being caused by source files other than the ones modified in this PR.

StephenVavasis:
Question for the experts:  Is there a good way to test that the new definitions in my PR for `IteratorEltype` and `IteratorSize` are correct?  I didn't know what to put in the test set to test these functions.

kmsquire:
Hi Stephen, you’ll probably want to ask this on Discourse, and/or ping Keno
directly.

Cheers,
   Kevin

StephenVavasis:
The CI checks are now passed, but the coverage has dropped.  One reason the coverage dropped is because I didn't test the new `length` and `eltype` methods, so I'm adding more tests for these now.  However, the coverage will probably be depressed because of the branch on which version in container_loops.jl, unless the coverage tool merges coverage from 0.6 and 0.7. (does it?)

StephenVavasis:
As an aside: I know Base has some undocumented infrastructure for iterating over  keys (e.g., `Base.Keyset`) which is not used in container_loops.jl.  What would be the advantage of using this infrastructure?  Is it required for compatibility with 0.7?

StephenVavasis:
I inserted additional tests in `test_sorted_containers.jl` for new `eltype` and `length` methods in order to improve coverage.  The testing in fact uncovered bugs, so I also revised `container_loops.jl` again.  But the coverage still seems to be low, so before this is merged, I'd like to test the coverage locally.  Is the GitHub coverage tool the same as `Coverage.jl`?  Is the tool run in Julia 0.6 or 0.7 or both?

tpapp:
Since [#28365](https://github.com/JuliaLang/julia/pull/28365) introduced deprecation messages, it would be great to get this merged.

StephenVavasis:
I can try to get back to work on this.  I just looked at source file `DataStructures.jl` in the master branch.  It still imports `start`, `done`, `next` but not `iterate`.  I would have presumed that someone else aside from this PR has updated that file since May.  Is there another branch besides master and this PR where there is already work on the new iteration protocol?

StephenVavasis:
Withdrawn-- superseded by #426 .

PR #386, opened 2018-05-25
Replaceordereddict

StephenVavasis:
Addresses problems with making OrderedDict compatible with 0.7 via a complete rewrite.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 doc/build
 docs/build/
 docs/site/
+.gitignore
+*~
+
--- a/docs/src/sorted_containers.md
+++ b/docs/src/sorted_containers.md
@@ -71,7 +71,7 @@ then there may be a loss of performance compared to:
 k,v = deref((sc,st))
 ```
 
-because the former needs an extra heap allocation step for `tok`.
+because the former may need an extra heap allocation step for `tok`.
 
 The notion of token is similar to the concept of iterators used by C++
 standard containers. Tokens can be explicitly advanced or regressed
@@ -406,8 +406,8 @@ past-end token. Time: O(1)
 ## Iteration Over Sorted Containers
 
 As is standard in Julia, iteration over the containers is implemented
-via calls to three functions, `start`, `next` and `done`. It is usual
-practice, however, to call these functions implicitly with a for-loop
+via calls to the function `Base.iterate`. It is usual
+practice, however, to call this function implicitly with a for-loop
 rather than explicitly, so they are presented here in for-loop notation.
 Internally, all of these iterations are implemented with semitokens that
 are advanced via the `advance` operation. Each iteration of these loops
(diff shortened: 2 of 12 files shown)


kmsquire:
First, some history: I wrote the first version of `OrderedDict`, which was reasonably performant.  @JeffBezanson rewrote this version to be more performant, especially for iteration. He originally targeted Julia itself (https://github.com/JuliaLang/julia/pull/10116). I brought it here when it became clear that it wasn't going to be merged there.  (I thought I rewrote the commit message for him to be the owner, but it seems my git-foo wasn't up to snuff yet.)

Regarding this version:
* Ideally, the two commits would be in separate merge requests, especially since they have nothing to do with one another.
* I appreciate the goal of making the code easier to use and more maintainable.  That said, I find quite doubtful that this type of rewrite can ever be as performant as the previous version.
* That said, I have no time to work on this myself right now, so in order to get things working, I'm okay with having something like this merged.  If has the time and inkling to improve it and maintain it in the future, great.
* If anyone else has the time to look to see if the current implementation might be fixed with a more surgical approach, that would be nice...

Going to review this now.

StephenVavasis:
Thanks for the history!  I suspected something like that must have occurred. I will try to rebase these changes on the latest master and split into two PR's. In the next comment I'll post my thoughts on partially recovering performance.

StephenVavasis:
I see three ways to get a more performant `OrderedDict` beyond what this PR offers.

1. Go back to the old approach in which the maintainers of `Dict` pinky-swear that they will also maintain `OrderedDict`.  I see this as unlikely; it appears that the core developers are pushing as much out of Base as possible, presumably in order to be able to more easily delegate responsibility for code maintenance.

2. Ask for a complete refactoring of dict.jl to split off a high-level interface from a low-level substrate.  In principle, this is the best solution, but everyone seems very busy right now.

3. Try to improve the performance of the code in this PR.  There are two separate sources of performance losses, which I'll expand on now.

First, as observed by Martin Holters, if one repeatedly inserts and deletes keys and then carries out many loops over the whole OrderedDict, then the performance of this PR can be arbitrarily worse (i.e., a ratio greater than any constant factor) compared to the old implementation.  This can be fixed as follows: Every time a key is added, one checks whether a1 has become more than twice as long as d1, and if so, one runs a compression on a1.  This will return the performance ratio to a constant factor in all cases, I believe.

Second, the new code often makes multiple probes (as many as 3) into the underlying Dict for a single operation.  To fix this, a token interface to Dict is needed.  There is already an open issue for this here: https://github.com/JuliaLang/julia/issues/24454, so maybe we can press for this issue to be addressed.

If these two fixes are applied, then the new code for OrderedDict will still be slower than the old; the ratio will depend on how long hashing takes, but I expect the ratio to range from 1.25 to 2.

StephenVavasis:
This PR is no longer needed because of #448.

PR #390, opened 2018-06-21
sprinkle at-inline and at-propagate_inbounds around

s-broda:
Timings:

Before:
```
julia> using DataStructures

julia> using BenchmarkTools

julia> cb=CircularBuffer{Float64}(3)
0-element DataStructures.CircularBuffer{Float64}

julia> pusher(cb)=(@inbounds for i=1:10^6; push!(cb, 3.);end; nothing)
pusher (generic function with 1 method)

julia> @benchmark pusher(cb)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     6.944 ms (0.00% GC)
  median time:      6.995 ms (0.00% GC)
  mean time:        7.029 ms (0.00% GC)
  maximum time:     8.994 ms (0.00% GC)
  --------------
  samples:          713
  evals/sample:     1
```

After:
```
julia> @benchmark pusher(cb)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.400 ms (0.00% GC)
  median time:      2.414 ms (0.00% GC)
  mean time:        2.422 ms (0.00% GC)
  maximum time:     3.129 ms (0.00% GC)
  --------------
  samples:          2069
  evals/sample:     1
```

Similar results for larger buffers.

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -15,14 +15,14 @@ function Base.empty!(cb::CircularBuffer)
     cb
 end
 
-function _buffer_index_checked(cb::CircularBuffer, i::Int)
-    if i < 1 || i > cb.length
+Base.@propagate_inbounds function _buffer_index_checked(cb::CircularBuffer, i::Int)
+    @boundscheck if i < 1 || i > cb.length
         throw(BoundsError(cb, i))
     end
     _buffer_index(cb, i)
 end
 
-function _buffer_index(cb::CircularBuffer, i::Int)
+@inline function _buffer_index(cb::CircularBuffer, i::Int)
     n = cb.capacity
     idx = cb.first + i - 1
     if idx > n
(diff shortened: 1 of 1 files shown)


oxinabox:
Those performance improvements are nice
Good stuff.
And to me this looks completely correct.
(I'm not super familiar with the inbound propergation macros but this match's the manual).

I'm a bit unhappy about the 0.7 tests failing, but I think that us unrelated.
I'll double check tomorrow and get back to you.

If this doesn't happen in next few days feel free to pung me. I might have forgotten.

oxinabox:
For my own future reference here is a link to a build of master I just triggered.
https://travis-ci.org/JuliaCollections/DataStructures.jl/builds/395401162
If it fails on 0.7 then that means this PR didn't break 0.7 and so i think am all clear to merge it.

oxinabox:
Yes, master is failing, and in the same way.
And this code looks right, so this is all good.

Thank you for this,

s-broda:
Thanks for merging!
Cheers, 
Simon.

PR #393, opened 2018-06-29
fixes for 0.7

fredrikekre:


fredrikekre:
DataStructures should still be updated for the new iteration protocol, but that is not urgent.

src/DataStructures.jl
```diff
@@ -16,9 +16,7 @@ module DataStructures
                  union, intersect, symdiff, setdiff, issubset,
                  searchsortedfirst, searchsortedlast, in
 
-    using Compat
-    using Compat.InteractiveUtils # for methodswith
-    import Compat: lastindex, pushfirst!, popfirst!
+    using InteractiveUtils # for methodswith
```
oxinabox: `using InteractiveUtils: methodswith`

Also the `dict_support.jl` function that uses it needs to be updated to look for `iterate` rather than `start`/ `done`/ `next`.
Whether as part of this PR or later idk.

src/default_dict.jl
```diff
@@ -52,25 +52,16 @@ DefaultDictBase{K,V}(default::F; kwargs...) where {K,V,F} = DefaultDictBase{K,V,
...
 @delegate_return_parent DefaultDictBase.d [ delete!, empty!, setindex!, sizehint! ]
 
 similar(d::DefaultDictBase{K,V,F}) where {K,V,F} = DefaultDictBase{K,V,F}(d.default; passkey=d.passkey)
-if isdefined(Base, :KeySet) # 0.7.0-DEV.2722
-    in(key, v::Base.KeySet{K,T}) where {K,T<:DefaultDictBase{K}} = key in keys(v.dict.d)
-    next(v::Base.KeySet{K,T}, i) where {K,T<:DefaultDictBase{K}} = (v.dict.d.keys[i], Base.skip_deleted(v.dict.d,i+1))
-else
-    in(key, v::Base.KeyIterator{T}) where {T<:DefaultDictBase} = key in keys(v.dict.d)
-    next(v::Base.KeyIterator{T}, i) where {T<:DefaultDictBase} = (v.dict.d.keys[i], Base.skip_deleted(v.dict.d,i+1))
-end
```
oxinabox: I do not believe you intended to delete both the `if` clause and the `else` clause
I think you intended to leave the `KeySet` code here.
And just get rid of `KeyIterator`

Or perhaps not, since 
https://github.com/JuliaCollections/DataStructures.jl/pull/393/files#diff-69860e53de6ca0b86046a70a1dd966cbR152
defines it for things via the meta-programming loop?

src/int_set.jl
```diff
@@ -5,12 +5,7 @@ import Base: similar, copy, copy!, eltype, push!, pop!, delete!,
              setdiff, setdiff!, symdiff, symdiff!, in, start, next, done,
              last, length, show, hash, issubset, ==, <=, <, unsafe_getindex,
              unsafe_setindex!, findnextnot, first
-import Compat: popfirst!
-if !isdefined(Base, :complement)
-    export complement, complement!
-else
-    import Base: complement, complement!
-end
+export complement, complement!
```
oxinabox: These should move to the main file (`src/DataStructures.j`) with the other exports.
(on their own line like most of the "subpackage" exports)

src/int_set.jl
```diff
@@ -54,7 +49,7 @@ end
 @inline function _resize0!(b::BitVector, newlen::Integer)
     len = length(b)
     resize!(b, newlen)
-    len < newlen && unsafe_setindex!(b, false, len+1:newlen) # resize! gives dirty memory
+    len < newlen && (b[len+1:newlen] .= false) # resize! gives dirty memory
```
oxinabox: should there be an `@inbounds` here?
To match the `unsafe_setindex!`?

src/ordered_dict.jl
```diff
@@ -428,12 +424,6 @@ function start(t::OrderedDict)
 end
 done(t::OrderedDict, i::Int) = i > length(t.keys)
 next(t::OrderedDict, i::Int) = (Pair(t.keys[i],t.vals[i]), i+1)
-
-if isdefined(Base, :KeySet) # 0.7.0-DEV.2722
-    next(v::Base.KeySet{K,T}, i) where {K,T<:OrderedDict{K}} = (v.dict.keys[i], i+1)
-else
-    next(v::Base.KeyIterator{T}, i) where {T<:OrderedDict} = (v.dict.keys[i], i+1)
-end
```
oxinabox: Again, I am not sure you intended to delete both branches of this `if` and `else`

test/test_priority_queue.jl
```diff
@@ -34,23 +34,15 @@ end
     r = rand(1:pmax, n)
     priorities = Dict(zip(1:n, r))
 
+    ks, vs = 1:n+1, rand(1:pmax, n)
+
```
oxinabox: this is correct, but man I do not like how stateful all these old tests are.

test/test_sorted_containers.jl
```diff
@@ -1300,10 +1297,6 @@ end
     factors5 = SortedMultiDict(test_pair_array)
     @test typeof(factors5) == SortedMultiDict{Char,Any,ForwardOrdering}
 
-    println("The following warning is expected:")
-    factors6 = SortedMultiDict(['a', 'b'], [1, 2])
-    @test typeof(factors6) == SortedMultiDict{Char,Int,ForwardOrdering}
-
     #@test factors2 == factors3   # Broken!  TODO: fix me...
```
oxinabox: This line looks like it should have been deleted long ago,
or changed to an `@test_broken`

Not related to this PR though

src/default_dict.jl
fredrikekre: It wasn't needed so thats why I removed it... Perhaps it will hit a fallback and this is an optimization? I can add it back, but would be good to know the reason for them. Perhaps they are just not tested?

oxinabox: I wonder if we can delete the other ones instead?
I would much rather be hitting code defined on an abstract type,
than be generating methods via an eval loop for all subtypes of that abstract type.

Inheritance > Metaprogramming

src/DataStructures.jl
fredrikekre: I think that changes should me made when changing to the new iteration protocol.

fredrikekre: (I note that this is used for a very weird thing also, perhaps just a method error is better but w/e).

oxinabox: Could do, thought it is actually not directly related to them.
Changing to new iteration protocol. is about changing types we define to that protocol.
Changing `dict_supports.jl` is about changing the types we *consume*  to be the new protocol.
But yes it can go in that PR,

fredrikekre:
> Can you reassure me that the times you deleted both branches of an if-else you did so on purpose?

I had a quick look, and it seems that none of those were called, so I think we can leave this as is for now.

oxinabox:
I am satisfied and pleased.
Thanks, nice work

PR #395, opened 2018-07-09
Fix #394 with the missing keys after sorting

### LINKED ISSUE #394: OrderedDict Key Not Found
I'm 90% sure this is something I've screwed up, but for some reason I get a key not found error for ordered dictionaries. My types are quite long and full of data, but I'm effectively making a Dict with integer keys, then sorting it.

For example, I have:

```
DataStructures.OrderedDict{Int64,JAXTAM.GTIData} with 320 entries:
  3   => JAXTAM.GTIData(:nustar, :FPMA, "10002008001", 1.0, 3, 3, [8, 7, 10, 6, 11, 14, 15, 13, 9, 8  …  6, 6, 9, 10…
  4   => JAXTAM.GTIData(:nustar, :FPMA, "10002008001", 1.0, 4, 42, [8, 10, 17, 12, 21, 13, 19, 8, 14, 16  …  14, 10,…
  5   => JAXTAM.GTIData(:nustar, :FPMA, "10002008001", 1.0, 5, 150, [6, 10, 14, 2, 16, 14, 8, 13, 19, 14  …  8, 10, …
  7   => JAXTAM.GTIData(:nustar, :FPMA, "10002008001", 1.0, 7, 389, [12, 16, 15, 13, 8, 15, 20, 6, 10, 11  …  11, 9,…
```

Where `3`, `4`, `5`, `7`, etc... are the integer keys:

```
Base.KeyIterator for a DataStructures.OrderedDict{Int64,JAXTAM.GTIData} with 320 entries. Keys:
  3
  4
  5
  7
```

oxinabox:
This is  my attempt at fixing #394.
Using ` idx = ht_keyindex2(d, key)` or `idx =  ht_keyindex(d, key, true)` 
fixes the problem with the key vanishing,
but causes problems with out of bounds errors.

I tried a few other things.
Then I took a look at how `rehash!`  worked,
and saw that it rebuilds the `slots` based only `keys` and `values` (if nothing has been deleted since last `rehash!`)
And so I figured: Just correct the `keys` and `values` and then `rehash` it.

I have also created a Julia0.6 branch with this change.
https://github.com/JuliaCollections/DataStructures.jl/tree/julia0.6
What I should have done is created a 0.6 branch without this change then PRed that.
So if this PR gets rejected cos it is wrong, that branch should be deleted.
Otherwise, that branch should be used to tag `v0.8.4`

@timholy  this does not need to be ported to OrderedCollections.jl (I tried without looking by just `stash apply`ing the change set, and was informed that) OrderedCollections.jl is missing sorting functionality and that corresponding file.
I think that file should be moved over.
But that can be done with this change applied.

### DIFF
--- a/src/dict_sorting.jl
+++ b/src/dict_sorting.jl
@@ -11,15 +11,9 @@ function sort!(d::OrderedDict; byvalue::Bool=false, args...)
     else
         p = sortperm(d.keys; args...)
     end
-
-    for (i,key) in enumerate(d.keys)
-        idx = ht_keyindex(d, key, false)
-        d.slots[idx] = p[i]
-    end
-
     d.keys = d.keys[p]
     d.vals = d.vals[p]
-
+    rehash!(d)
     return d
 end
 
--- a/test/test_sorting.jl
+++ b/test/test_sorting.jl
@@ -24,4 +24,8 @@
     @test sort(unordered; rev=true) == rev
     @test sort(unordered; byvalue=true) == rev
     @test sort(unordered; byvalue=true, rev=true) == forward
+
+    @testset "Bug DataStructures.jl/#394" begin
+        @test sort(Dict(k=>string(k) for k in 1:3))[1] == "1"
+    end
 end


StephenVavasis:
I'll just remark that I proposed in https://github.com/JuliaCollections/DataStructures.jl/pull/382 and https://github.com/JuliaCollections/DataStructures.jl/pull/386 a complete rewrite of OrderedSet and OrderedDict in terms of Array and Dict (rather than in terms of low-level Julia primitives as in the current code).  The purpose of the rewrite is to improve maintainability.  The rewrite removes some undocumented functionality, namely, use of integer subscripts on OrderedSet, which appeared to be incompletely implemented anyway.

oxinabox:
@StephenVavasis  i do not thing this is mutually exclusive with that.
Either here, or in OrderedCollections.jl

In that this is (I think) simpler than that,
and so can be merged first while you are sorting out the re-basing (and maybe the rebalancing )you mentioned in the last comment.
And then overwritten with your change once that is ready to go.
(While keeping the new test)
If that is the way we are going to go with this.


However::
I recall that I didn't venue to say anything on that thread because I concluded that I had only a limitted understanding of how HashTables worked,
and even less how your SortedDicts etc worked.

As such I didn't venture and opinion there, and I still don't.
Honestly my understanding of if this fix even works is limited, and kind of trial and error.
It passes all the tests.

timholy:
> The rewrite removes some undocumented functionality, namely, use of integer subscripts on OrderedSet, which appeared to be incompletely implemented anyway.

I've already deprecated it in OrderedCollections, see https://github.com/JuliaCollections/OrderedCollections.jl/commit/35ac721d4e609b83958a508d6e1188be81099afb.

PR #402, opened 2018-07-20
Implemented WeakKeyIdDict

mauro3:
This is what I put together in https://github.com/JuliaLang/julia/pull/28161 (closed PR).  I thought it could perhaps live here. Note that a WeakKeyIdDict might go into Base sometime in 1.x: https://github.com/JuliaLang/julia/pull/28182, but who knows when.


mauro3:
Is there any interest in this?  Otherwise please close.

thautwarm:
Any news?

mauro3:
After the great recent interest in this PR (two people in two month! ;-), I updated it such that it passes again tests and fixed conflicts.  Ready to go, as far as I'm concerned.

thautwarm:
Thanks. This is quite useful!

oxinabox:
Per https://github.com/JuliaCollections/DataStructures.jl/issues/479
we are not accepting new data structures til post-1.0.0

I guess a case could be made that this PR was open long before then and so should be an exception?

ericphanson:
+1 for an exception, since I want to use this too :). Or maybe it can be it's own small package?

src/weakkeyid_dict.jl
```diff
@@ -0,0 +1,132 @@
...
+        k = getkey(wkh.ht, WeakRefForWeakDict(kk), secret_table_token)
+        k === secret_table_token && return default
+        return k.w.value::K
+    end
+end
+
+Base.get(wkh::WeakKeyIdDict{K}, key, default) where {K} = lock(() -> get(wkh.ht, WeakRefForWeakDict(key), default), wkh)
+Base.get(default::Base.Callable, wkh::WeakKeyIdDict{K}, key) where {K} = lock(() -> get(default, wkh.ht, WeakRefForWeakDict(key)), wkh)
+Base.get!(wkh::WeakKeyIdDict{K}, key, default) where {K} = lock(() -> get!(wkh.ht, WeakRefForWeakDict(key), default), wkh)
+Base.get!(default::Base.Callable, wkh::WeakKeyIdDict{K}, key) where {K} = lock(() -> get!(default, wkh.ht, WeakRefForWeakDict(key)), wkh)
```
ericphanson, line 102: It would be nice to use `Base.@lock` here instead of the closure form. It can be more performant and keeps stacktraces shorter; I have a recursive call where this almost doubles the number of stack frames. It is exported on Julia 1.9 and present in many versions of Julia, but not present in 1.0 unfortunately, which DataStructures still supports.

PR #405, opened 2018-07-29
Fill CircularBuffer

### LINKED ISSUE #404: Implement Base.fill!(cb, value) method for an empty CircularBuffer
Following https://github.com/JuliaCollections/DataStructures.jl/issues/223#issuecomment-333996424 from @sairus7

we should implement a fill method for an empty `CircularBuffer`

femtotrader:
Closes #404

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -94,7 +94,7 @@ Add an element to the back and overwrite front if full.
 end
 
 """
-    popfirst!(cb::CircularBuffer)
+    popfirst!(cb)
 
 Remove the first item from CircularBuffer.
 """
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -92,6 +92,12 @@ function Base.append!(cb::CircularBuffer, datavec::AbstractVector)
     cb
 end
 
+function Base.fill!(cb::CircularBuffer, data)
+    fill!(cb.buffer, data)
+    cb.length = cb.capacity
+    cb
+end
+
 Base.length(cb::CircularBuffer) = cb.length
 Base.size(cb::CircularBuffer) = (length(cb),)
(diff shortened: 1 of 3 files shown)


src/circular_buffer.jl
```diff
@@ -92,6 +92,12 @@ function Base.append!(cb::CircularBuffer, datavec::AbstractVector)
     cb
 end
 
+function Base.fill!(cb::CircularBuffer, data)
```
oxinabox: This could do with a docstring.
Highlighting, exactly what it does.

In terms of grows the buffer up-to capacity, and fills it entirely.
Overwriting any existing elements.

Some might expect it to fill only unused elements.

test/test_circular_buffer.jl
```diff
@@ -73,4 +73,9 @@
     end
     @test isempty(cb)
     @test_throws ArgumentError popfirst!(cb)
+    
+    # test fill!(cb, data)
```
oxinabox: I recommend when adding new tests to a file in DataStructures,
always use nested `@testsets` per functionality
and  taking the opportunity to update the remaining earlier tests to use testsets too.

The only reason it isn't that way, is because DataStructures.jl is much older than `@testset`

So this would become 

```
@testset "fill!" begin
    cb = CircularBuffer{Int}(3)
    fill!(cb, 42)
    @test Array(cb) == [42, 42, 42]
end
```

src/circular_buffer.jl
femtotrader: Should we have a keyword argument to distinguish the 2 kind of behavior?

What other collections are doing with `fill!`?

femtotrader: `append!`  code could help

femtotrader:
`fill!` now preserves existing elements.

oxinabox:
thanks.
Merging

PR #409, opened 2018-07-29
Modify queue constructor

### LINKED ISSUE #363: Queue{T}() constructor is missing
Hello,

I noticed that `Deque{Int}()` is a possible construction of a deque but `Queue{Int}()` maybe the following constructor is missing to provide a more uniform API

```julia
Queue{T}() where {T} = Queue(T)
```

Same for `Stack{Int}()`

Kind regards

s-celles:
Closes #363


scls19fr:
This PR also split `test_stack_and_queue.jl` into `test_stack.jl` and `test_queue.jl`

src/queue.jl
```diff
@@ -9,8 +9,18 @@ end
 
 Create a `Queue` object containing elements of type `T`.
 """
-Queue(ty::Type{T}) where {T} = Queue(Deque{T}())
-Queue(ty::Type{T}, blksize::Integer) where {T} = Queue(Deque{T}(blksize))
+Queue{T}() where {T} = Queue(Deque{T}())
+Queue{T}(blksize::Integer) where {T} = Queue(Deque{T}(blksize))
+
+function Queue(ty::Type{T}) where {T}
+    Base.depwarn("Queue(T) is deprecated, use Queue{T}() instead")
```
ararslan: I think you can just use
```julia
@deprecate Queue(::Type{T}) where {T} Queue{T}()
```
If that doesn't work, `depwarn` requires a second argument, so this should instead be
```julia
    Base.depwarn("Queue(T) is deprecated, use Queue{T}() instead", :Queue)
```

src/queue.jl
```diff
@@ -9,8 +9,18 @@ end
...
+    Queue{T}()
+end
+
+function Queue(ty::Type{T}, blksize::Integer) where {T}
+    Base.depwarn("Queue(T, blksize) is deprecated, use Queue{T}(blksize) instead")
```
ararslan: Should be able to use
```julia
@deprecate Queue(::Type{T}, blksize::Integer) where {T} Queue{T}(blksize)
```
here as well.

scls19fr:
This PR should be ready now

scls19fr:
Thank @ararslan

PR #416, opened 2018-07-30
Modify Accumulator constructor

### LINKED ISSUE #408: Deprecate Accumulator(T, V) in favor of Accumulator{T, V}()
Following #363

We should probably have for example

```julia
Accumulator{String, Int}()
```

instead of

```julia
Accumulator(String, Int)
```

to provide a more uniform API

scls19fr:
Closes #408


oxinabox:
Yes,
this is good

Issue #412, opened 2018-07-30
Add nested @testsets per functionality


scls19fr:
Hello,

currently unit tests don't use a lot of nested @testsets
more precisely we don't have one testset per functionaly.

Adding nested @testsets per functionality could help to find which tests
are failing.

Kind regards

PR #421, opened 2018-08-03
More nested testsets

scls19fr:
More nested testsets for `test_sorted_containers.jl`

Related #412

Some testsets are still very (too) big... but I didn't have time to split them (if that's even possible).

Spliting `test_sorted_containers` could also be considered but probably not in this PR.

### DIFF
--- a/test/test_sorted_containers.jl
+++ b/test/test_sorted_containers.jl
@@ -275,320 +275,337 @@ eq(::CaseInsensitive, a, b) = isequal(lowercase(a), lowercase(b))
 
 
     @testset "SortedDictBasic" begin
+    
         # a few basic tests of SortedDict to start
         m1 = SortedDict((Dict{String,String}()), Forward)
         @test typeof(m1) == SortedDict{String, String, ForwardOrdering}
-        kdarray = ["hello", "jello", "alpha", "beta", "fortune", "random",
-                   "july", "wednesday"]
-        checkcorrectness(m1.bt, false)
-        for i = 1 : div(size(kdarray,1), 2)
-            k = kdarray[i*2-1]
-            d = kdarray[i*2]
-            #println("- inserting: k = $k d = $d")
-            m1[k] = d
+
+        @testset "testset 1" begin
+            kdarray = ["hello", "jello", "alpha", "beta", "fortune", "random",
+                       "july", "wednesday"]
             checkcorrectness(m1.bt, false)
-            # fulldump(m1.bt)
+            for i = 1 : div(size(kdarray,1), 2)
+                k = kdarray[i*2-1]
+                d = kdarray[i*2]
+                #println("- inserting: k = $k d = $d")
+                m1[k] = d
+                checkcorrectness(m1.bt, false)
+                # fulldump(m1.bt)
+            end
         end
-        i1 = startof(m1)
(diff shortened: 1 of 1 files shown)


StephenVavasis:
Here is the backstory on why test_sorted_containers takes so long to run.  The balanced-tree code is full of nested branching statements.  It contains both `if` blocks and many single-line branches with `? :`, `||` and `&&`.  After I wrote that code, I wanted to make sure every branch was tested, so I wrote a custom coverage tool that checked coverage of each branch including single-line branches.  The test framework was the "sieve" algorithm for identifying prime numbers, which leads to a sequence of irregular tree-traversals.  I found that in order to reach every branch according to my coverage tool, the code needed to run for a long time.  In principle it is possible, though probably quite tedious, to systematically design a sequence of insertion and deletion operations that reach every branch with a much smaller execution time.

scls19fr:
Thanks @StephenVavasis for this explanation.
I have try to isolate this kind of code in "big testset" testsets as much as I can

test/test_sorted_containers.jl
```diff
@@ -275,320 +275,337 @@ eq(::CaseInsensitive, a, b) = isequal(lowercase(a), lowercase(b))
...
-        kdarray = ["hello", "jello", "alpha", "beta", "fortune", "random",
-                   "july", "wednesday"]
-        checkcorrectness(m1.bt, false)
-        for i = 1 : div(size(kdarray,1), 2)
-            k = kdarray[i*2-1]
-            d = kdarray[i*2]
-            #println("- inserting: k = $k d = $d")
-            m1[k] = d
+
+        @testset "testset 1" begin
```
ararslan: It seems to me that if this doesn't separate out the tests based on the particular functionality being tested, it doesn't need to be in its own testset. If it does separate functionality, perhaps a more descriptive name than "testset 1" would be good? 😉

ararslan: Same comment for most of the rest of these; I won't go through and mark them all individually.

scls19fr: For sure "testset 1" is not a very descriptive name! But it can help to catch (with a certain level of distance) where tests are broken. I'm open to suggestions for a more descriptive name for this testset and more broadly for other testsets which need a better naming.

I don't think that diff really help to understand globality of such a PR 

But a visual comparison of
https://github.com/scls19fr/DataStructures.jl/blob/418649bcbd8b476dd5cc22b31646e4d7b7eec887/test/test_sorted_containers.jl
https://github.com/JuliaCollections/DataStructures.jl/blob/master/test/test_sorted_containers.jl
can help to see the general approach

Maybe I shouldn't have tackle this this way... maybe a first should be to split `test_sorted_containers.jl` into several files (some files for unit tests and file(s) for defining functions such as `fulldump`, `checkcorrectness`, ... because a file with nearly 2000 lines of tests is something we should avoid.

With this work on this PR, I noticed that running tests of this file is pretty long, and so we should be able to run just a part of this (I also wonder if Julia tests have timing features and how we can speedup tests)

With Python and [nose](http://nose.readthedocs.io/) we have [nose-timer](https://github.com/mahmoudimus/nose-timer) maybe such a timing plugin should be developed (or enabled)

PR #422, opened 2018-08-03
update iterator protocol for multiple types

Evizero:
Note that this doesn't update everything yet, but its progress.
I am not sure how this will interact with #384 

I marked the remaining occurances of the old iterator interface with `# FIXME`

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -1,8 +1,6 @@
-__precompile__()
-
 module DataStructures
 
-    import Base: <, <=, ==, length, isempty, start, next, done, delete!,
+    import Base: <, <=, ==, length, isempty, iterate, delete!,
                  show, dump, empty!, getindex, setindex!, get, get!,
                  in, haskey, keys, merge, copy, cat,
                  push!, pop!, pushfirst!, popfirst!, insert!, lastindex,
@@ -16,6 +14,7 @@ module DataStructures
                  union, intersect, symdiff, setdiff, issubset,
                  searchsortedfirst, searchsortedlast, in,
                  eachindex, keytype, valtype
+    import Base: start, next, done
 
     using InteractiveUtils: methodswith
 
--- a/src/accumulator.jl
+++ b/src/accumulator.jl
@@ -58,12 +58,7 @@ sum(ct::Accumulator) = sum(values(ct.map))
 
 ## iteration
 
-start(ct::Accumulator) = start(ct.map)
-next(ct::Accumulator, state) = next(ct.map, state)
-done(ct::Accumulator, state) = done(ct.map, state)
-# resolve ambiguity
-next(ct::Accumulator, state::Base.LegacyIterationCompat{I,T,S}) where {I>:Accumulator,T,S} = next(ct.map, state)
-done(ct::Accumulator, state::Base.LegacyIterationCompat{I,T,S}) where {I>:Accumulator,T,S} = done(ct.map, state)
+iterate(ct::Accumulator, s...) = iterate(ct.map, s...)
 
 # manipulation
 
--- a/src/circ_deque.jl
+++ b/src/circ_deque.jl
@@ -86,9 +86,9 @@ end
 end
 
 # Iteration via getindex
-@inline Base.start(d::CircularDeque) = 1
-@inline Base.next(d::CircularDeque, i::Int) = (_unsafe_getindex(d, i), i+1)
-@inline Base.done(d::CircularDeque, i::Int) = i == d.n + 1
+@inline function iterate(d::CircularDeque, i = 1)
+    i == d.n + 1 ? nothing : (_unsafe_getindex(d, i), i+1)
+end
 
 function Base.show(io::IO, D::CircularDeque{T}) where T
     print(io, "CircularDeque{$T}([")
--- a/src/classified_collections.jl
+++ b/src/classified_collections.jl
@@ -34,12 +34,7 @@ keys(cc::ClassifiedCollections) = keys(cc.map)
 
 ## iteration
 
-start(cc::ClassifiedCollections) = start(cc.map)
-next(cc::ClassifiedCollections, state) = next(cc.map, state)
-done(cc::ClassifiedCollections, state) = done(cc.map, state)
-# resolve ambiguity
-next(ct::ClassifiedCollections, state::Base.LegacyIterationCompat{I,T,S}) where {I>:ClassifiedCollections,T,S} = next(ct.map, state)
-done(ct::ClassifiedCollections, state::Base.LegacyIterationCompat{I,T,S}) where {I>:ClassifiedCollections,T,S} = done(ct.map, state)
+iterate(cc::ClassifiedCollections, s...) = iterate(cc.map, s...)
 
 # manipulation
 
--- a/src/default_dict.jl
+++ b/src/default_dict.jl
@@ -51,10 +51,7 @@ DefaultDictBase{K,V}(default::F; kwargs...) where {K,V,F} = DefaultDictBase{K,V,
 
 # most functions are simply delegated to the wrapped dictionary
 @delegate DefaultDictBase.d [ get, haskey, getkey, pop!,
-                              start, done, next, isempty, length ]
-# resolve ambiguity
-next(d::DefaultDictBase, state::Base.LegacyIterationCompat{I,T,S}) where {I>:DefaultDictBase,T,S} = next(d.d, state)
-done(d::DefaultDictBase, state::Base.LegacyIterationCompat{I,T,S}) where {I>:DefaultDictBase,T,S} = done(d.d, state)
+                              iterate, isempty, length ]
 
 # Some functions are delegated, but then need to return the main dictionary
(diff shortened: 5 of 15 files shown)


StephenVavasis:
Is it possible to merge this PR, and then I can resume work on #384?

kmsquire:
LGTM!  I restarted the failed Linux jobs (both were timeouts), but since this passed on OSX and Windows, I think it can be merged as is.  

The drop in coverage is perhaps an issue emitting correct line numbers that the coverage tool can pick up?  
The main (only?) example is https://coveralls.io/builds/18320019/source?filename=src/ordered_dict.jl#L246, where the next line is covered 270,332 times.

kmsquire:
Well, the tests timed out again on nightlies when running `test_sorted_containers.jl`.  This test set is pretty extensive, so it's not that surprising, but perhaps we should consider parallelizing our tests...

oxinabox:
having looked at the code Base uses to parallelize it's tests,
the code needed to do that **nicely** is kinda extensive.

I feel like if we wanted to parallelize our tests **nicely**,
we should make a separate package that generically parallelizes tests.
(Based on the JuliaLang/julia code).

And then DataStructures.jl can depend upon it.

But OTOH,
we should just shunt SortedContainers into it's own package,
and re-export it.
That would improve manythings, including the test run time

Evizero:
`test_sorted_containers.jl` takes ages locally as well. with and without this PR. This may be a regression caused by the new julia version (?)

StephenVavasis:
I'll have a look at improving the test set for sorted containers.  This may take me a couple of days.

StephenVavasis:
I started working on this and discovered that the current DataStructures.jl master apparently does not run under 0.6.  Is this correct-- are we dropping support for 0.6 already?

Evizero:
I think since 0.6 versions are tagged and preserved, we should just focus 0.7 only going forward. there is really no reason to try to support both on master

StephenVavasis:
I've done some profiling and timing of `test_sorted_containers.jl`, and it appears that the reason it is so slow is not because of the tests per se but because of the sheer number of times that the `@test` macro needs to be expanded for this file.  Apparently expanding `@test` is not a super-speedy operation, and there are hundreds of invocations of this macro in `test_sorted_containers.jl`.  (By the way, it is actually faster by about 30% in 0.7 versus 0.6.)

Perhaps I could fix this by rewriting each test set as a no-argument function with no `@test` macros but a lot of assert statements, and then invoking `@test` just once for the function?  To some extent, that would defeat the purpose of the `@test` macro.  Or perhaps it is possible to "pre-expand" macros?  I'm open to suggestions!

oxinabox:
I really think moving it to its own package, and reexporting that, is the solution.
Do you disagree?

StephenVavasis:
I don't exactly see how splitting the package would solve the problem; it seems that splitting just pushes the problem down the road a bit?

oxinabox:
It is unlikely that the number of tests for Sorted Containers will ever substantially increase.
Right now there is almost enough time for them to run, plus everything else.
Thus if they were in their own package, there should be enough time for them all to run -- since there would be nothing else.

So yes, I guess you could say it pushed it down the road a bit.
But I would say it pushes it a fair bit, far enough that I expect we'ld not have to worry about the new package ever timing out on tests.
And once more of DataStructures.jl is broken out, each of those packages too won't have to worry about running out of time,
and nor will the whole since it will have almost no tests.

davidanthoff:
I think it would be great if we could find a way to just quickly merge and tag this here. I don't have an opinion about splitting, other than it sounds more work than we should take on with a 0.7 release around the corner and a need to get a tagged version out that works smoothly on 0.7.

StephenVavasis:
OK, I will try to greatly reduce the time for testing sorted containers.  Does anyone know what is the limit (number of seconds) allowed for testing a package in CI?  Currently on my laptop with Pkg.test("DataStructures") (tagged version, not master), sorted containers takes 40 seconds to test, while all the rest of DataStructures takes 35 seconds.

oxinabox:
@davidanthoff  makes a good point.


Here is my suggested quick hack to speed it up.


A faster version of the `@test` macro, that doesn't produce as pretty an output, but which is a lot faster.
over 10x faster by my count.

```
julia> @time @testset "Foo" begin
           @test 1==ceil(rand())
               @test 10==10ceil(rand())
               end
Test Summary: | Pass  Total
Foo           |    2      2
  0.010400 seconds (2.71 k allocations: 172.127 KiB)
Test.DefaultTestSet("Foo", Any[], 2, false)

julia>

julia> @time @testset "Foo" begin
           @fasttest 1==ceil(rand())
               @fasttest 10==10ceil(rand())
               end
Test Summary: | Pass  Total
Foo           |    2      2
  0.000824 seconds (240 allocations: 29.719 KiB)
Test.DefaultTestSet("Foo", Any[], 2, false)
```

## here is the code

```
using Test
using Test: Returned, Threw, do_test

function get_fast_test_result(ex, source)
    testret = :(Returned($(esc(ex)), nothing, $(QuoteNode(source))))
    result = quote
        try
            $testret
        catch _e
            _e isa InterruptException && rethrow(_e)
            Threw(_e, catch_backtrace(), $(QuoteNode(source)))
        end
    end
    Base.remove_linenums!(result)
    result
end

macro fasttest(ex)
	orig_ex = Expr(:inert, ex)
    result = get_fast_test_result(ex, __source__)
	:(do_test($result, $orig_ex))
end

```

oxinabox:
##Test Script
```


@time (()->(macroexpand(Main, :(@testset "Foo" begin
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
	@fasttest 1==ceil(rand())
	@fasttest 10==10ceil(rand())
	@fasttest 100==100ceil(rand())
end)); nothing))()
```

and the same with the `fast` removed

### @fasttest
0.015670 seconds (9.44 k allocations: 592.848 KiB)
0.014021 seconds (9.44 k allocations: 592.848 KiB)
	
	
###@test

0.018269 seconds (11.34 k allocations: 696.910 KiB)
0.017976 seconds (11.34 k allocations: 696.910 KiB)


So fairly minor improvement -- 14%, but over 2000 uses that, tests might add up

davidanthoff:
> sorted containers takes 40 seconds to test, while all the rest of DataStructures takes 35 seconds.

That cannot even remotely bump against travis limits, right?

StephenVavasis:
OK, I got `test_sorted_containers.jl` down to 10 seconds on my laptopp.  I did this by replacing all the `@test X` with `my_assert(X)` (it turns out that `assert()` has been deprecated-- who knew?), removing uninvoked functions, reducing the sizes of the test containers, making  fewer calls to `checkcorrectness()`, and completely removing all `@test_throw` invocations, which are kind of expensive.  (So the `throw` statements aren't tested.)

However, I am not expert enough with git to know how to insert the new file inside this PR.  Could whoever owns this PR please obtain my new source file and just paste it into your commit?  My file is here:

https://gist.github.com/StephenVavasis/2339e686b36c1eb24645fc55e236caeb

Also, the new version does not need the `Primes` package any more (I wrote `my_primes` instead to speed this up), so you can remove `using Primes` from runtests.jl and remove it from the REQUIRE file.

StephenVavasis:
@oxinabox sorry, I did not see your faster `@test` macro until after I finished my update.  Can you also write a faster `@test_throw` macro?  I'd like to put the test-throws back in the test-set if it's not too costly.

Evizero:
succeeded by #425

PR #423, opened 2018-08-04
Dont precompile on recent 0.7

rened:


Evizero:
This is somewhat included in #422 (without the check)

rened:
Good!

PR #424, opened 2018-08-04
add Project.toml

Evizero:


oxinabox:
The test time dependency on Primes is real, not a StdLib,
but it **doesn't** have a Project.toml file.
As such, I believe this is safe
and will not break JuliaCIBot.



> **Andreas Noack** [1 day ago]
> Let me clarify. The remaining issue is if a test dependency is a real package and has a project file. Using stdlib during tests should be fine. They don’t have to be declared.  
>    
> Btw. I think I’m pretty close to a proper fix where CIBot will be using the new package manager but I’ll have to test it a little further.

oxinabox:
Ok apparently now everything is fine. because new CIBot was just deployed
So lets do this

StephenVavasis:
I required Primes in the old version of `test_sorted_containers.jl`, but this requirement was removed by #426 .  I don't know if any other part of DataStructures requires this package.

PR #426, opened 2018-08-07

StephenVavasis:
This commit rewrites container_loops.jl to incorporate the new iteration protocol
(introduced in 0.7.0-DEV) more thoroughly. It is not compatible with 0.6.

function when applied to subranges (i.e., inclusive(a,b,c) or
of the length of the subrange. (There should be no value returned

    IntSet was changed to BitSet (name change in 0.7.0-DEV)
    Small updates to documentation
    Some assert statements in balanced_tree.jl that were present

### DIFF
--- a/src/balanced_tree.jl
+++ b/src/balanced_tree.jl
@@ -25,6 +25,13 @@
 ##  data nodes.
 
 
+macro invariant(expr)
+end
+
+macro invariant_support_statement(expr)
+end
+
+
 struct KDRec{K,D}
     parent::Int
     k::K
@@ -375,7 +382,7 @@ function insert!(t::BalancedTree23{K,D,Ord}, k, d, allowdups::Bool) where {K,D,O
     ## longer hold undefined references.
 
     if size(t.data,1) == 2
-        # @assert(t.rootloc == 1 && t.depth == 1)
+        @invariant t.rootloc == 1 && t.depth == 1
         t.tree[1] = TreeNode{K}(t.tree[1].child1, t.tree[1].child2,
                                 t.tree[1].child3, t.tree[1].parent,
                                 k, k)
@@ -486,7 +493,7 @@ function insert!(t::BalancedTree23{K,D,Ord}, k, d, allowdups::Bool) where {K,D,O
         ## If p1 is the root (i.e., we have encountered only 3-nodes during
         ## our ascent of the tree), then the root must be split.
         if p1 == t.rootloc
-            # @assert(curdepth == 1)
+            @invariant curdepth == 1
             splitroot = true
             break
         end
@@ -547,7 +554,7 @@ end
 
 function nextloc0(t, i::Int)
     ii = i
-    # @assert(i != 2 && i in t.useddatacells)
+    @invariant i != 2 && i in t.useddatacells
     @inbounds p = t.data[i].parent
     nextchild = 0
     depthp = t.depth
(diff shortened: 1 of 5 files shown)


StephenVavasis:
I forgot to include in the commit message: this PR also rewrites `test_sorted_containers.jl` so that it runs much faster (10 sec versus 30+ sec).

src/balanced_tree.jl
```diff
@@ -631,30 +631,30 @@ function compareInd(t::BalancedTree23, i1::Int, i2::Int)
...
     p2 = t.data[i2].parent
-    curdepth = t.depth
+    # curdepth = t.depth
     while true
-        @assert(curdepth > 0)
+        # @assert(curdepth > 0)
         if p1 == p2
             if i1a == t.tree[p1].child1
-                @assert(t.tree[p1].child2 == i2a || t.tree[p1].child3 == i2a)
+                # @assert(t.tree[p1].child2 == i2a || t.tree[p1].child3 == i2a)
```
kmsquire: Do you think you’ll ever uncomment these again?  If not, then it’s better to just delete them.

StephenVavasis:
I won't ever uncomment those statements, but they do indicate invariants satisfied by the algorithm that may assist someone else in deciphering the code.  Is there a better way to indicate (with a comment) a program invariant?

src/balanced_tree.jl
oxinabox: The curdepth commented out lines can be removed though. They are not asserts

StephenVavasis:
Ok, I removed the commented-out asserts.  The `curdepth` variable in that section of the code existed only to support the `asserts`, so I deleted the commented-out statements involving `curdepth` as well.

kmsquire:
I don’t know of a specific convention regarding invariants, but it would be nice to document them as such, to distinguish them from commented our code. Perhaps add `Invariant:` to the beginning of the line?

oxinabox:
I am sorry, I think I have miscommunicated.

I would not have deleted them.
I think for now best would be to 
define a pair of macros:

`@invariant_assert(expr)`
and `@invariant_supporting(expr)`.

Which for now just effectively comment out their contents.
But later we can add some kind of mode that enables them.
A bit like how `@assert` works in other languages,
except very disabled even in debug mode.

The other reason not to delete them is while we may later wish to inform the compiler of these facts later for optimisation purposes.
(or if not these facts, similar facts in other places. The principle stands)
So we really do want to keep them documented.

StephenVavasis:
OK, I backed out the last commit using `git reset` and changed the commented-out asserts to a new no-op macro as suggested.  However, I don't know how to upload this commit to github.  Below is the error message.  Git is suggesting that I first pull, but of course I don't want to pull; I just want this latest commit to supersede the previous one.  Any suggestions?  I'm not well-versed in git.

```
c:\Users\vavasis\OneDrive - University of Waterloo\GitHub2\DataStructures.jl\src>git push origin container_loop_fixes
To https://github.com/JuliaCollections/DataStructures.jl
 ! [rejected]        container_loop_fixes -> container_loop_fixes (non-fast-forward)
error: failed to push some refs to 'https://github.com/JuliaCollections/DataStructures.jl'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

oxinabox:
I believe this calls for `git push --force`

kmsquire:
As @oxinabox said, `git push --force` is the correct thing to do in this case.

kmsquire:
Bump.  @StephanVavasis, is this good to go?  Do you need to force push?

StephenVavasis:
Should be good to go... there are now no-op macros to indicate program invariants.

PR #427, opened 2018-08-11
Fix confusion in circular buffer

### LINKED ISSUE #429: Confusion in circular buffer
When using pushfirst! in a situation where length != capacity and first != 1
the buffer content is shifted by one internally which results in an incorrect
buffer state.

**How to reproduce:**
```julia
using DataStructures
buffer = CircularBuffer{Int}(5)
map(x -> pushfirst!(buffer, x), 1:8)
pop!(buffer)
pushfirst!(buffer, 9)
```
**Expected buffer state**
```
[9, 8, 7, 6, 5]
```
**Actual state**
```
[4, 8, 7, 9, 5]
```

To prevent duplicate work, see PR #427.

This PR fixes this issue.

In addition to fixing this issue, handling of pushing first is improved. In the previous implementation `pushfirst!(cb.buffer)` was called. This leads to storing inaccessible elements in a range outside of the buffer capacity. Since references to these elements are technically still accessible via the internal cb.buffer array, these elements are not GCed and result in memory leakage.

This is my first PR to a large public project, so please tell me if I can improve something :smiley:

tynsh:
Closes #429

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -118,13 +118,11 @@ and overwrite back if full.
 """
 function pushfirst!(cb::CircularBuffer, data)
     # if full, decrement and overwrite, otherwise pushfirst
-    if length(cb) == cb.capacity
-        cb.first = (cb.first == 1 ? cb.capacity : cb.first - 1)
-        cb[1] = data
-    else
+    cb.first = (cb.first == 1 ? cb.capacity : cb.first - 1)
+    if length(cb) < cb.capacity
         cb.length += 1
-        pushfirst!(cb.buffer, data)
     end
+    cb.buffer[cb.first] = data
     cb
 end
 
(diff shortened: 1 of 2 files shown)


scls19fr:
Thanks @tynsh for this first contribution here.

Please use triple backticks for code section in Github https://help.github.com/articles/creating-and-highlighting-code-blocks/ for better rendering (see my edit)

It's generally better to first [open an issue](https://github.com/JuliaCollections/DataStructures.jl/issues) and after create a PR which closes this issue.

PR comment could be "Closes #..." where ... is the number of this issue

Your PR can't be accept like this because a unit test should show what your are trying to fix
so please add a testset in `test_circular_buffer.jl` named "issue ..."

Anyway happy to see new contributions here!

scls19fr:
Pinging @joshday who contributed to `function pushfirst!(cb::CircularBuffer, data)`

scls19fr:
@kmsquire @oxinabox what is your opinion?

kmsquire:
Thanks for the contribution!

PR #365, opened 2018-01-08
Update default_dict.md

scls19fr:
Closes #364

Issue #360, opened 2018-01-05
Improve docstrings and doc


scls19fr:
Hello,

Doc need now to be improved

Example:
http://juliacollections.github.io/DataStructures.jl/latest/deque.html
(but it also apply to most doc page)

Instead of a usage paragraph with raw Julia code we should use docstrings and `@docs`

Example of docstring for `isempty`
https://github.com/JuliaLang/julia/blob/edeb002a13afdce31dac55706433ebb6d5cedd79/base/essentials.jl#L727

Writing doc this way will help to have a doc which really reflect current code (and avoid issues with documentation of constructors such as exposed in #340 )

Maybe a new release should be tagged before tackling this issue because it will be easier to compare latest version (with @docs) to previous (current) version with raw Julia code in doc.

Kind regards

oxinabox:
I'ld suggest putting this on the backburner until #310.
Than after that is done there will be 5 packages and one can go through and tackle them one at a time.

scls19fr:
I noticed that one reason for splitting this package is that it should help fixing (future) dep warning. Is https://github.com/apps/femtocleaner installed ?

scls19fr:
When trying to build doc, we get:

    !! No docs found for 'find(m::SortedDict, k_)'. [src/sorted_containers.md]
    !! No docs found for 'get(sd::SortedDict,k,v)'. [src/sorted_containers.md]
    !! No docs found for 'get!(sd::SortedDict,k,v)'. [src/sorted_containers.md]

Maybe @mortenpi @fredrikekre or @KristofferC who are active contributor of [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl/graphs/contributors) can tell us why we are facing this

mortenpi:
As far as I can tell, DataStructures does not define `find(m::SortedDict, k_)` at all. As for `get` and `get!`: the docstrings appear to be attached to a different signature, e.g. `get(default_::Union{Function,Type}, m::SortedDict{K,D}, k_) where {K,D}`, not `get(m::SortedDict, k_, default_)`.

scls19fr:
Thanks @mortenpi 
Sorry for noise.

scls19fr:
Doc should explain:
- how to install and use this package (latest tagged version)
- how to install development version and run all unit tests (or run just one test file)
- how to contribute to this repository (both technically and code of conduct)

scls19fr:
[...]

PR #431, opened 2018-08-13
Improve DefaultDict doc

### LINKED ISSUE #364: Deprecated doc for construction of DefaultDict
Hello,

I noticed in doc the following construction of DefaultDict 

https://github.com/JuliaCollections/DataStructures.jl/blame/c31f17f0d0dead44ae7766b36fd880dcae666a9d/doc/src/default_dict.md#L18

but this kind of construction is deprecated so it shouldn't be in doc

```julia
julia> DefaultDict(String, Int, 0)
WARNING: DefaultDict(::Type{K}, ::Type{V}, default) where {K, V} is deprecated, use DefaultDict{K, V}(default) instead.
Stacktrace:
 [1] depwarn(::String, ::Symbol) at .\deprecated.jl:70
 [2] DataStructures.DefaultDict(::Type{String}, ::Type{Int64}, ::Int64) at .\deprecated.jl:57
 [3] eval(::Module, ::Any) at .\boot.jl:235
 [4] eval_user_input(::Any, ::Base.REPL.REPLBackend) at .\REPL.jl:66
 [5] macro expansion at .\REPL.jl:97 [inlined]
 [6] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at .\event.jl:73
while loading no file, in expression starting on line 0
DataStructures.DefaultDict{String,Int64,Int64} with 0 entries

julia> DefaultDict{String, Int}(0)
DataStructures.DefaultDict{String,Int64,Int64} with 0 entries
```

This issue is comparable to #340

Kind regards

scls19fr:
with setup/repl macro

Closes #364
Helps #360 
Replaces #365

### DIFF
--- a/docs/src/default_dict.md
+++ b/docs/src/default_dict.md
@@ -14,13 +14,13 @@ DefaultDict(default, kv)    # create a DefaultDict with a default value or funct
                             # optionally wrapping an existing dictionary
                             # or array of key-value pairs
 
-DefaultDict(KeyType, ValueType, default)   # create a DefaultDict with Dict type (KeyType,ValueType)
+DefaultDict{KeyType, ValueType}(default)   # create a DefaultDict with Dict type (KeyType,ValueType)
 
 DefaultOrderedDict(default, kv)     # create a DefaultOrderedDict with a default value or function,
                                     # optionally wrapping an existing dictionary
                                     # or array of key-value pairs
 
-DefaultOrderedDict(KeyType, ValueType, default) # create a DefaultOrderedDict with Dict type (KeyType,ValueType)
+DefaultOrderedDict{KeyType, ValueType}(default) # create a DefaultOrderedDict with Dict type (KeyType,ValueType)
 ```
 
 All constructors also take a `passkey::Bool=false` keyword argument which determines whether to pass along the `key`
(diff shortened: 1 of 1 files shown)


scls19fr:
This PR shouldn't be merged for now because Documenter.jl repl/setup macros seems to be buggy when using CRLF EOL.

After saving this `default_dict.md` file with LF as EOL, local build of doc is correctly rendered but using such EOL is not a correct workaround as all other files are using CRLF and git complains when adding it

    warning: LF will be replaced by CRLF in docs/src/default_dict.md.
    The file will have its original line endings in your working directory.

See https://github.com/JuliaDocs/Documenter.jl/issues/790 and https://github.com/JuliaDocs/Documenter.jl/issues/749

Waiting for @mortenpi / @fredrikekre (or others [Documenter.jl contributors](https://github.com/JuliaDocs/Documenter.jl/graphs/contributors)) to fix it

kmsquire:
Can you add a commit before this one, which changes all doc files to use LF instead of CRLF?

scls19fr:
Sorry @kmsquire but I don't understand why I should do that.

Under Windows, git is replacing automatically LF with CRLF
See https://stackoverflow.com/questions/1967370/git-replacing-lf-with-crlf

Do you know a command line to do it quickly?

fredrikekre:
Documenter v19.5 is released with the bugfix included so if that was the problem here you are good to go.

PR #435, opened 2018-08-16
Create Makefile for doc building

scls19fr:


scls19fr:
I often forget  `--color=yes` optional parameter, but that's very convenient when building doc.
So this `Makefile` is a kind of reminder

PR #440, opened 2018-08-23
Solve issue #434

### LINKED ISSUE #434: Priorityqueue does not allow to have arbitrary priorities
Hi

Following minimal example demonstrates the problem:

```Julia
julia> using DataStructures

julia> pq = PriorityQueue("a" => "A")
PriorityQueue{String,String,Base.Order.ForwardOrdering} with 1 entry:
  "a" => 1
```

Kind regards

Ben

BenLauwens:
Closes #434

Fix is quite trivial
Iterate without second argument also has to return a key-value pair.

### DIFF
--- a/test/test_priority_queue.jl
+++ b/test/test_priority_queue.jl
@@ -210,6 +210,15 @@ import Base.Order.Reverse
             @test Set(collect(pq)) == Set(["a"=>2, "c"=>1])
         end
     end
+    
+    @testset "Iteration" begin
+        io = IOBuffer()
+        pq = PriorityQueue('a'=>'A', 'b'=>'B', 'c'=>'C')
+        for (k, v) in pq
+            println(io, k, " => ", v)
+        end
+        @test String(take!(copy(io))) == "a => A\nc => C\nb => B\n"
+    end
 
     @testset "LowLevelHeapOperations" begin
         pmax = 1000
--- a/test/test_priority_queue.jl
+++ b/test/test_priority_queue.jl
@@ -217,7 +217,11 @@ import Base.Order.Reverse
         for (k, v) in pq
             println(io, k, " => ", v)
         end
-        @test String(take!(copy(io))) == "a => A\nc => C\nb => B\n"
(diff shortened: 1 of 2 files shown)


scls19fr:
Thanks @BenLauwens for your contribution but could you add a test (which was failing previously
but which is now passing with your fix)

BenLauwens:
Test added;)

test/test_priority_queue.jl
```diff
@@ -210,6 +210,19 @@ import Base.Order.Reverse
...
+        pq = PriorityQueue('a'=>'A', 'b'=>'B', 'c'=>'C')
+        for (k, v) in pq
+            println(io, k, " => ", v)
+        end
+        if Sys.WORD_SIZE == 32
+            @test String(take!(copy(io))) == "b => B\nc => C\na => A\n"
+        else
+            @test String(take!(copy(io))) == "a => A\nc => C\nb => B\n"
+        end
+    end
```
kmsquire: It's not really a good idea to depend on the sort order of an unordered collection.  The actual order is an implementation detail that might change.  (For example, there was talk previously of changing `Dict` to maintain insertion order of keys when iterating.  It didn't happen, but it could.)

Instead, a simpler test would probably suffice in this case.  How about just doing 

```julia
pq = PriorityQueue('a'=>'A')
@test string(dequeue_pair!(pq)) == "'a' => 'A'"
```

From your own examples in #434, this would have failed previously.

BenLauwens:
Changes implemented as asked in review;)

BenLauwens:
I don't understand the coverage/coveralls output ...

BenLauwens:
Can this be merged? It is blocking migration of SimJulia to Julia v0.7 - v1.0
Thanks!

kmsquire:
(Whoops--I should have squashed this.)

kmsquire:
New release will be available when https://github.com/JuliaLang/METADATA.jl/pull/17383 is merged.

PR #442, opened 2018-08-24
Make the LinkedList implementation immutable

sjoelund:
Nothing in the linked list uses mutation and the type can simply be
changed to immutable. This assumes that no code in the wild uses this
linked list as a mutable type.

### DIFF
--- a/src/list.jl
+++ b/src/list.jl
@@ -1,9 +1,9 @@
 abstract type LinkedList{T} end
 
-mutable struct Nil{T} <: LinkedList{T}
+struct Nil{T} <: LinkedList{T}
 end
 
-mutable struct Cons{T} <: LinkedList{T}
+struct Cons{T} <: LinkedList{T}
     head::T
     tail::LinkedList{T}
 end
--- a/test/test_list.jl
+++ b/test/test_list.jl
@@ -103,4 +103,9 @@
         @test collect(l10) == [2; 4; 5.6; 10.5]
     end
 
+    @testset "immutable" begin
+        @test isimmutable(list(1))
+        @test isimmutable(nil())
+    end
+
 end # @testset LinkedList


oxinabox:
Thanks for the suggestion.
While at first this seems like a sensible optimization,
on further thought I really don't think we can rely  on no one using the linked list as immutable.

A mutable linked list and an immutable linked list are rather different beasties.

In a mutable linked list,  mutation of elements is _O(1)_.
In an immutable linked list it is _O(n-i)_ where _n_ is the length of the list, and _i_ is the index

a mutable linked list can have all kinds of stitching done to it to slice out a range of elements.
Can even do things, like attach its head to its tail so it is infinitely long.

There is a serviceable immutable linked list in 
https://github.com/JuliaCollections/FunctionalCollections.jl/blob/master/src/PersistentList.jl 

Which we should perhaps mention in the docs

sjoelund:
Thanks for the pointer to FunctionalCollections; it seems to be a package with more immutable types that seems to be more what I'd like.

Some notes though:
> While at first this seems like a sensible optimization,
I don't consider having an immutable list an optimization, but rather having the ability to pass a data structure without fear of it being mutated. Copying data around is both annoying and error-prone.

> In a mutable linked list, mutation of elements is O(1).
> In an immutable linked list it is O(n-i) where n is the length of the list, and i is the index

Actually, mutation of a mutable linked list is `O(i)` where `i` is the index (you need to walk `i` elements to find the position to mutate). It is `O(i)` also for immutable linked lists (where you walk `i` elements and create `i` new nodes in a partially new list).

> a mutable linked list can have all kinds of stitching done to it to slice out a range of elements.
> Can even do things, like attach its head to its tail so it is infinitely long.
I know. I don't necessarily consider it a feature (see above). Haskell people like this pattern, and manage to do it on immutable lists...

oxinabox:
> Actually, mutation of a mutable linked list is O(i) where i is the index (you need to walk i elements to find the position to mutate). It is O(i) also for immutable linked lists (where you walk i elements and create i new nodes in a partially new list).

Sure.
I was thinking mutation once already in position. e.g. if you were already iterating over the list.
If not already in position yes, immutable is _O(i+i) = O(2i) = O(i)_

Anyway, are you happy to close this? or is there still things to be done?
Might be worth actionizing adding a link to the docs.
Though that would be easier in a fresh PR

sjoelund:
Yes, I guess I'm happy with closing the PR. I think it would be good to make the description of the package state that it contains mutable types and then link to FunctionalCollections for an alternative with immutable types.

PR #448, opened 2018-09-19
Rebase on OrderedCollections

timholy:
This removes the OrderedDict/OrderedSet functionality in favor of the implementation in [OrderedCollections](https://github.com/JuliaCollections/OrderedCollections.jl). Sorry this took so long.


src/dict_support.jl
```diff
@@ -1,12 +0,0 @@
-# support functions
-
-# _tablesz and hashindex are defined in Base, but are not exported,
-# so they are redefined here.
-_tablesz(x::Integer) = x < 16 ? 16 : one(x)<<((sizeof(x)<<3)-leading_zeros(x-1))
-hashindex(key, sz) = (reinterpret(Int,(hash(key))) & (sz-1)) + 1
-
-function not_iterator_of_pairs(kv)
```
oxinabox: I am not please at depending on importing this from 
*OrderedCollections*,
because out other dicts need it, 
and ideally we'ld be able to cut them out into their own packages, 
which would also not depend on `OrderedCollections`

I see two alternative options.
1. just duplicate the code. I don't mind this, while it is not DRY, it  does let us evolve the remainind dicts independently of **OrderedCollections**
2. make a micro package **DataStructuresUtil.jl** or some such, and have both import this.

timholy: Interestingly, I think OrderedCollections doesn't even use that function. So I can put it back here and delete it from there.

oxinabox:
Checking others
`is_ordered` is a trait, (exists reL #237)
and it conceptually belongs to  the OrderedCollection namespace

src/DataStructures.jl
```diff
@@ -16,7 +16,9 @@ module DataStructures
                  eachindex, keytype, valtype
     import Base: iterate
 
-    using InteractiveUtils: methodswith
+    using OrderedCollections
+    using OrderedCollections: not_iterator_of_pairs, rehash!
+    import OrderedCollections: filter, filter!, isordered
```
oxinabox: `filter` and `filter!` both come from `Base`
Why are they being imported?

src/DataStructures.jl
```diff
@@ -16,7 +16,9 @@ module DataStructures
                  eachindex, keytype, valtype
     import Base: iterate
 
-    using InteractiveUtils: methodswith
+    using OrderedCollections
+    using OrderedCollections: not_iterator_of_pairs, rehash!
```
oxinabox: `rehash!` AFAICT is only used by *OrderedCollections*
so I don't think we need it to be used.
(or did I miss an usage?)

oxinabox:
Ok, I am done with my look over.
Thanks for doing this.

(Complete sidenote; thanks for the recommendation of Gopen and Swan's "The Science of Scientific Writing". I wish I'ld read it at the start of my PhD, but reading it before writing my thesis intro/conclusion is definitely better than never.)

src/DataStructures.jl
timholy: Well, they have to be imported because DataStructures extends them. I imported them from OrderedCollections instead of Base due to the `Iterators.filter` separation---I basically thought "this should be the same `filter` that OrderedCollections uses, whatever version of `filter` that is."

src/DataStructures.jl
timholy: https://github.com/JuliaCollections/DataStructures.jl/blob/bdd7aa034f569fbea7fef5d1998caf1863937528/src/dict_sorting.jl#L6 and https://github.com/JuliaCollections/DataStructures.jl/blob/bdd7aa034f569fbea7fef5d1998caf1863937528/src/dict_sorting.jl#L16

Project.toml
```diff
@@ -4,6 +4,7 @@ version = "0.12.0"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
+OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
```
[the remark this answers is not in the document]
timholy: I don't think it's necessary, since the latest tagged version of OrderedCollections works here.

src/DataStructures.jl
oxinabox: Ah, 
I thought that file had been moved over.
I think it should be moved over.
It uses only Base and OrderedCollections types, no?

timholy:
Good to go?

src/DataStructures.jl
timholy: Yes, you're absolutely right.

Oddly, these comments only show up if I access this page in certain ways. Weird GitHub bug?

Project.toml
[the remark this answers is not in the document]
timholy: Since I was making other changes, I added the compat

REQUIRE
```diff
@@ -1 +1,2 @@
 julia 0.7
+OrderedCollections 1.0.1
```
timholy: Can we just get rid of this file now?

oxinabox: Nope, I checked this morning.
Attobot won't talk to you if you don't have a REQUIRE file.

timholy:
Good to go? The appveyor failure seems independent.

timholy:
Unless there are objections I'm going to merge this soon because in the current situation we are causing problems: see https://github.com/JuliaCollections/OrderedCollections.jl/issues/10

timholy:
Thanks! Sorry this took more review from you than it should have.

PR #449, opened 2018-09-21
Define empty! for PriorityQueue

timholy:
In addition to defining `empty!`, this makes `PriorityQueue` immutable. I don't see a reason it should be mutable, and indeed modifying its, e.g., ordering after filling it with items seems like a bad idea.


kmsquire:
Appveyor failure is https://github.com/JuliaLang/julia/issues/29300

Issue #443, opened 2018-08-24
Add features to LinkedList


oxinabox:
When looking at #442 
I noticed just how featureless our  `LinkedList` is.
In particular, as pointed out by #442 we have no mutating operations.

I think adding some of the missing methods would be great for someone new to julia.
Since implementing a linked list is like DataStructures101.

Here is a list of functions I think should be implemented.
Feel free to suggest more.

 - `getindex` 
 - `setindex!`  
 - `delete!` with overloads both for single indexes and for ranges.
 - `append!`
 - The iterator traits (or are the defaults right?)

c-p-murphy:
I think it might be worthwhile to move the current `LinkedList` to something like `FunctionalList` and then build a new mutable `LinkedList`. Or make `LinkedList` immutable and build something like `MutableList`. I've been looking at the current code and thinking about how to add on the mutable functionality, but I am reluctant to mess around too much with the current functional style.

c-p-murphy:
Also, I just found this, which seems to be a pretty comprehensive implementation of mutable linked lists:
https://github.com/ChrisRackauckas/LinkedLists.jl

c-p-murphy:
I don't really see a way to modify the existing code to add the mutable functionality without reworking the whole thing. It might make more sense to make the current `LinkedList` immutable and move it to something like `FunctionalList` and then write a new mutable `LinkedList` from scratch. 

Another option would be to incorporate [this package](https://github.com/ChrisRackauckas/LinkedLists.jl) into `DataStructures`. It still needs to be updated to work with julia v1.0, but it seems like a fairly comprehensive implementation.

oxinabox:
Making the current `LinkedList` immutable is a breaking change.

c-p-murphy:
I might be wrong, but as far as I can see, with the existing implementation, there isn't any way to take advantage of the mutability. So, presumably, no one is relying on it then, right?

c-p-murphy:
I guess this works with the existing implementation:
```julia
function setindex!(l::LinkedList{T}, data, idx::Int) where T
    node = l
    for i = 1:idx
        node = node.tail
    end
    node.head = data
    return l
end
```

c-p-murphy:
There also seems to be an issue with the current implementation when trying to delete the first element in a list (it works for every other element). For example:
```julia
function tail!(l::LinkedList)
    l = l.tail
    return l
end

julia> l = list(1:10...)
[...]

Issue #310, opened 2017-07-22
Split up the package?


ararslan:
This package has become quite large and is something of a heavyweight dependency for packages that just need a single type. An example of that is StatsBase, which only uses DataStructures for the arrays as heaps stuff. Indeed, just going through and trying to fix the deprecation warnings in this package for more recent Julia versions has proved quite cumbersome. It seems to me that for the sake of maintainability, the easily separable pieces of this package could be split off into separate packages, say like Heaps.jl, Deques.jl, etc. That way packages can just pull in the functionality they need. Thoughts?

yuyichao:
> just going through and trying to fix the deprecation warnings in this package for more recent Julia versions has proved quite cumbersome

What about it is cumbersome? Splitting part that doesn't depend on anything else might be ok though I don't think depwarn fix is a good reason for it and I don't see how splitting will help there. The depwarn fix now needs to be down on multiple packages rather than one.

ararslan:
That wasn't really an argument for splitting up so much as just a comment on the fact that it's big enough that fixing depwarns is annoying to do all at once.

TotalVerb:
I also think this package is unnecessarily large. It can reasonably be changed to simply re-export the exports of smaller packages which become dependencies, to maintain the current convenience while allowing more modular development

ararslan:
Exactly what I was thinking, @TotalVerb.

oxinabox:
Yes, please lets split it up.
It is so large.
I keep wanting to fix things, or refactor them, etc.
But the weight of all the files gets me down.
There is a lot of cruft in the code.
It has good test coverage, so refactoring and breaking down should be safe.
(Can keep a copy of all the tests in DataStructures.jl and then run them using the reexported ones until happy)

Breaking it into separate repos,  will make it easier to decruft it.
One thing that can be noted is that it really does beak up.
Some parts were clearly written by different people,
who (reasonably enough) did not touch the other parts.

Like All the Sorted stuff (see breakdown in next post) was written by @StephenVavasis .
All the Heap stuff was extracted from Base.

There are basically 5 packages living in here, all with there own core author list.
(and then the good work of later people maintaining them all)

I spent an hour or so checking and breaking down a potential split of the package.
I'ma put the list in a separate post.

oxinabox:
This is just the files in /src/

## Heaps.jl
Anything to do with heaps.
Lightweight package for people who just want the old priorityqueue from base,
or other heap related functions.

- heaps
[...]

PR #450, opened 2018-09-23
Add MutableLinkedList

c-p-murphy:
I put this together in response to issue [#443](https://github.com/JuliaCollections/DataStructures.jl/issues/443). This is my first real contribution julia, so feel free to share any advice and/or feedback about how I could improve this implementation.


oxinabox:
Thanks for doing this.

@ararslan, @kmsquire  I am really busy right now.
could either of you take a look at this?

Cross-ref
#442 and #443 


Generally, speaking I am kinda opposed to adding more DataStructures to Data Structures.jl.
as I still think we really need to be doing: #310 

But with that said, #310 is a fair bit of work and I've failed at it once, and don't see myself having free time to run with it anytime soon.
As such it really shouldn't block other things going forward, I guess.

src/mutable_list.jl
```diff
@@ -0,0 +1,217 @@
...
+    l.len -= 1
+    return data
+end
+
+function show(io::IO, node::ListNode)
+    x = node.data
+    print(io, "$(typeof(node))($x)")
+end
+
+function show(io::IO, l::MutableLinkedList)
```
ararslan: You can simplify this a bit using existing Base functions:
```julia
function show(io::IO, l::MutableLinkedList)
    print(io, typeof(l), '(')
    join(io, l, ", ")
    print(io, ')')
end
```

src/mutable_list.jl
```diff
@@ -0,0 +1,217 @@
...
+function copy(l::MutableLinkedList{T}) where T
+    l2 = MutableLinkedList{T}()
+    for h in l
+        push!(l2, h)
+    end
+    return l2
+end
+
+function getindex(l::MutableLinkedList, idx::Int)
+    0 < idx <= l.len || throw(ArgumentError("Index out of bounds"))
```
ararslan: Typically for bounds errors we should use Base's `@boundscheck` so that `@inbounds` knows how to skip bounds checking for `MutableLinkedList`s. So for example, I think this should be
```julia
@boundscheck 0 < idx <= l.len || throw(BoundsError(l, idx))
```

ararslan: (Same applies to other places you've implemented bounds checking)

src/mutable_list.jl
```diff
@@ -0,0 +1,217 @@
...
+
+isempty(l::MutableLinkedList) = l.len == 0
+length(l::MutableLinkedList) = l.len
+collect(l::MutableLinkedList{T}) where T = T[x for x in l]
+eltype(l::MutableLinkedList{T}) where T = T
+first(l::MutableLinkedList) = l.front.next.data
+last(l::MutableLinkedList) = l.back.prev.data
+
+function ==(l1::MutableLinkedList, l2::MutableLinkedList)
+    typeof(l1) == typeof(l2) || return false
```
ararslan: This is actually a conditional on the element type, since we know that the wrappers are the same. So you could do
```julia
==(l1::MutableLinkedList{T}, l2::MutableLinkedList{S}) where {T,S} = false

function ==(l1::MutableLinkedList{T}, l2::MutableLinkedList{T}) where T
    # the rest of the checking
end
```

src/mutable_list.jl
```diff
@@ -0,0 +1,217 @@
...
+        l.back = ListNode{T}()
+        l.front.next = l.back
+        l.back.prev = l.front
+        return l
+    end
+end
+
+MutableLinkedList() = MutableLinkedList{Any}()
+
+function MutableLinkedList{T}(elts...) where T
```
ararslan: I'm actually surprised this doesn't overwrite the method defined for the inner constructor, with an infinite loop when `elts` is empty. But it seems to work as expected, so nice. TIL.

ararslan:
This is looking good to me, though I'd be nice to have another set of eyes on it as well.

src/mutable_list.jl
```diff
@@ -0,0 +1,213 @@
...
+function ==(l1::MutableLinkedList{T}, l2::MutableLinkedList{T}) where T
+    length(l1) == length(l2) || return false
+    for (i, j) in zip(l1, l2)
+        i == j || return false
+    end
+    return true
+end
+
+function map(f::Function, l::MutableLinkedList{T}) where T
+    l2 = MutableLinkedList{T}()
```
kmsquire: This will break if the return type of `f(h)` is not `T`.

You probably want something like (untested):
```julia
    S = promote_type(Base.returntypes(f, (T,))...)
    l2 = MutableLinkedList{S}()
```

Alternatively (or in addition), you could allow the user to specify the desired eltype.

c-p-murphy: I couldn't find anything about a `returntypes` function, but what about something like this?
 ```julia
    S = typeof(f(first(l)))
    l2 = MutableLinkedList{S <: T ? T : S}()
```
It doesn't seem like the most elegant solution, but I'm not sure what a better approach would be (unless it makes more sense to just have the user specify the type).

Comment (kmsquire) [reply src/mutable_list.jl]: Actually, it was supposed to be `return_types` (with an underscore).  But reading a little further, it's use is supposed to be more targeted.

See https://discourse.julialang.org/t/obtaining-a-functions-output-type/11313/5 for a full discussion on how `map` is recommended to be implemented.

It's a little involved, though I don't think it's too complicated.  If you decide not to implement the full version here, I think what you specified above is fine (it's similar to the first step in the post)--just leave a breadcrumb to the full implementation.

Comment (c-p-murphy) [reply src/mutable_list.jl]: So, I think this is basically the approach that is suggested in that discussion:
```julia
function map(f::Base.Callable, l::MutableLinkedList{T}) where T
    if isempty(l) && f isa Function
        S = Core.Compiler.return_type(f, (T,))
        return MutableLinkedList{S}()
    elseif isempty(l) && f isa Type
        return MutableLinkedList{f}()
    else
        S = typeof(f(first(l)))
        l2 = MutableLinkedList{S}()
        for h in l
            el = f(h)
            if el isa S
                push!(l2, el)
            else
                R = typejoin(S, typeof(el))
                l2 = MutableLinkedList{R}(collect(l2)...)
                push!(l2, el)
            end
        end
        return l2
    end
end
```
It seems to work as expected for most inputs, but for something like this:
```julia
function f(x)
    if x % 2 == 0
        return convert(Int8, x)
    else
        return convert(Float16, x)
    end
end
```
`Core.Compiler.return_type`  returns the `Union` of the types rather than the closest common ancestor (as `typejoin` does). For example:
```julia
julia> map(f, MutableLinkedList{Float64}())
MutableLinkedList{Union{Float16, Int8}}()

julia> map(f, MutableLinkedList{Float64}(1:10...))
MutableLinkedList{Real}(1.0, 2, 3.0, 4, 5.0, 6, 7.0, 8, 9.0, 10)
```
Is that something worth worrying about or not really?

kmsquire:
Nice implementation!  I think that's fine.    I'm curious if there's a type
instability in there common case where the function return type doesn't
change?

c-p-murphy:
Thanks 😄

Unfortunately, it looks like it is isn't (I forgot to check). `@code_warntype` highlights these lines red:
```julia
julia> @code_warntype map(f, l)
Body::MutableLinkedList{_1} where _1
.
.
.
5 ── . . .
│    %9  = (Core.Compiler.return_type)(f, %8)::Type
│    %10 = (Core.apply_type)(DataStructures.MutableLinkedList, %9)::Type{MutableLinkedList{_1}} where _1
│    %11 = (%10)()::MutableLinkedList{_1} where _1
.
.
.
```
Not sure what the best way to fix it is though. Is there a recommended approach?

kmsquire:
Not that I'm aware of. In this case I think we just live with it. 

Thanks for the contribution!

PR #451, opened 2018-09-24
Allow negative indices <= 0 and indices > 2n in CircularBuffers

jbrea:
Please ignore this proposition, if you prefer not to allow negative indices and indices > 2n.

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -27,7 +27,7 @@ function Base.empty!(cb::CircularBuffer)
 end
 
 Base.@propagate_inbounds function _buffer_index_checked(cb::CircularBuffer, i::Int)
-    @boundscheck if i < 1 || i > cb.length
+    @boundscheck if !isfull(cb) && (i < 1 || i > cb.length)
         throw(BoundsError(cb, i))
     end
     _buffer_index(cb, i)
@@ -36,9 +36,12 @@ end
 @inline function _buffer_index(cb::CircularBuffer, i::Int)
     n = cb.capacity
     idx = cb.first + i - 1
-    0 < idx <= n && return idx
-    idx <= 0 && return n + idx % n
-    (idx - 1) % n + 1
+    if idx > 0
+        idx <= n && return idx
+        idx <= 2n && return idx - n
+        return (idx - 1) % n + 1
+    end
+    n + idx % n
 end
 
 """
--- a/test/test_circular_buffer.jl
+++ b/test/test_circular_buffer.jl
@@ -12,6 +12,7 @@
         
         @testset "With 1 element" begin
             push!(cb, 1)
+            @test_throws BoundsError cb[2]
             @test length(cb) == 1
             @test capacity(cb) == 5
             @test isfull(cb) == false
@@ -32,8 +33,10 @@
             @test cb[3] == 6
             @test cb[4] == 7
             @test cb[5] == 8
-            @test_throws BoundsError cb[6]
-            @test_throws BoundsError cb[3:6]
+            @test cb[20] == 8
+            @test cb[0] == 8
+            @test cb[-1] == 7
+            @test cb[-11] == 7
             @test cb[3:4] == Int[6,7]
             @test cb[[1,5]] == Int[4,8]
         end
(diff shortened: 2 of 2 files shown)


oxinabox:
Thanks for this.
I am not sure if it should be allowed or not.
for circular buffers.
I'ld have to think some more.
(or perhaps someone with experience with these in another language could chime in).

But in anycase,
if this is to be merged it needs to have some tests added.
to prove the functionality works.

Let us know if you have any trouble.

src/circular_buffer.jl
```diff
@@ -36,11 +36,9 @@ end
 @inline function _buffer_index(cb::CircularBuffer, i::Int)
     n = cb.capacity
     idx = cb.first + i - 1
-    if idx > n
-        idx - n
-    else
-        idx
-    end
+    idx > 0 && idx <= n && return idx
```
timholy: `0 < idx <= n && return idx` is a little easier to parse

jbrea:
Before
```julia
julia> using DataStructures, BenchmarkTools

julia> c = CircularBuffer{Int64}(100);

julia> for i in 1:100 push!(c, i) end

julia> import DataStructures: _buffer_index

julia> @benchmark _buffer_index($c, 40)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.953 ns (0.00% GC)
  median time:      1.966 ns (0.00% GC)
  mean time:        1.970 ns (0.00% GC)
  maximum time:     11.326 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark _buffer_index($c, 140)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.634 ns (0.00% GC)
  median time:      1.642 ns (0.00% GC)
  mean time:        1.647 ns (0.00% GC)
  maximum time:     10.826 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```
After:
```julia
julia> @benchmark _buffer_index($c, 40)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.954 ns (0.00% GC)
  median time:      1.969 ns (0.00% GC)
  mean time:        2.246 ns (0.00% GC)
  maximum time:     15.625 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark _buffer_index($c, 140)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     5.520 ns (0.00% GC)
  median time:      5.539 ns (0.00% GC)
  mean time:        5.602 ns (0.00% GC)
  maximum time:     17.713 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark _buffer_index($c, -140)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     11.299 ns (0.00% GC)
  median time:      11.521 ns (0.00% GC)
  mean time:        11.802 ns (0.00% GC)
  maximum time:     48.642 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     999
```

test/test_circular_buffer.jl
```diff
@@ -121,4 +121,17 @@
...
+        cb = CircularBuffer{Int}(5)
+        for i in 1:5
+            push!(cb, i)
+        end
+        k = 1
+        for j in -19:20
+            @test DataStructures._buffer_index(cb, j) == k
+            k += 1
+            if k > 5 k = 1 end
+        end
```
oxinabox: This is a little too complex.
Testing the right things,
but a good test set is trivially obvious.

Like 
```
@test cb[1]=1
@test cb[5]=5
@test cb[20]=5
@test cb[0]=5
@test cb[-1]=4
@test cb[-11]=6
```

Your `for` loop does the same thing,
and more comprehensively.
But it is not obvious what it is doing.

Test code that is non-obvious might have bugs.
(and because tests don't have tests ...)

jbrea:
@oxinabox Thanks for your feedback.
Originally I thought just to change the `_buffer_index` function, but I guess it makes sense to propagate the changes to `getindex` and `setindex`.

Issue #399, opened 2018-07-17
Push should promote?


ChrisRackauckas:
We had an issue where users wanted to push integers into a mutable binary heap that holds floats. It seems like this would work but instead it has a missing method error. 

https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/313

oxinabox:
Closed in #452

PR #452, opened 2018-09-24
fix push! and update! for heaps to convert the input value

mcognetta:
This fixes issue #399. That issue did not consider the case of updating a value in a mutable heap, but this PR contains a fix for that as well.


oxinabox:
Yes, this is good, 
and correct.
Thank you

Issue #469, opened 2018-12-11
Get rid of heap pseudoconstructors ?


oxinabox:
https://github.com/JuliaCollections/DataStructures.jl/blob/31044efe5130b8d1980e2b6650337da08f27274e/src/heaps/binary_heap.jl#L119-L125

The heap code has these pseudoconstructors, written in `snake_case`
This code is 6 years old and was written just a month after the julia 0.1 release.
I am feeling like they were a work around for some thing 0.1 was missing.
(In general a lot of the code had workarounds for things like the inability to just put type-parameters on functions without using them in the arguments. That is moistly gone now).

I think we can move these pseudoconstructors into overloads of the real constructors (in ProperCase),
and deprecate the old ones.

dillondaudert:
I'm willing to go ahead and give this a shot. How do you think users should instantiate the min/max versions of the heap? The first thing I thought of would be `BinaryHeap{T, Comp}() where {T, Comp} = new{T, Comp}(Vector{T}())` however that would require the user passing `GreaterThan()`/`LessThan()` themselves. 

The `GreaterThan`/`LessThan` types look like a holdover from when `Base.Order` didn't exist, and that seems OK as long as the users aren't forced to interact with them directly. If they are, I would much rather it be `BinaryHeap{T, O}() where {T, O<:Ordering} = new{T, O}(Vector{T}())`. Of course, that would require a more extensive rewrite of the heaps...

oxinabox:
I was thinking more along the lines of creating a type alias style const.

``` 
const BinaryMinHeap{T} = BinaryHeap{T,LessThan}
const BinaryMaxHeap{T} = BinaryHeap{T,GreaterThan}
```
I think that works.


This is not to say that the more extensive rewrite to use `Base.Order` is not needed.

dillondaudert:
Ok. Once #468 is merged, I'll make a PR for this.

dillondaudert:
This is a little tricky due to the ordering of the parametric arguments and how I'm thinking the constructors should work.
I'm going off of desiring two ways to instantiate a heap:
- `Heap{T}()`: initialize empty heap with element types `T`
- `Heap(xs::AbstractVector{T})`: initialize a heap with array of elements of type `T`.

For the second to be possible with `BinarMinHeap`/`BinaryMaxHeap`, the type aliases need to be defined without type parameters. However, if we do `const BinaryMinHeap = BinaryHeap{LessThan}` then

[...]

PR #472, opened 2018-12-27
Update heap constructors

dillondaudert:
This PR addresses #469 by deprecating the old pseudoconstructors `(mutable_)binary_{min|max}heap` and replacing them with `(Mutable)Binary{Min|Max}Heap` constructors.

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -33,7 +33,7 @@ module DataStructures
     export IntDisjointSets, DisjointSets, num_groups, find_root, in_same_set, root_union!
 
     export AbstractHeap, compare, extract_all!
-    export BinaryHeap, binary_minheap, binary_maxheap, nlargest, nsmallest
+    export BinaryHeap, BinaryMinHeap, BinaryMaxHeap, nlargest, nsmallest
     export MutableBinaryHeap, mutable_binary_minheap, mutable_binary_maxheap
     export heapify!, heapify, heappop!, heappush!, isheap
     export BinaryMinMaxHeap, popmin!, popmax!, popall!
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -34,7 +34,7 @@ module DataStructures
 
     export AbstractHeap, compare, extract_all!
     export BinaryHeap, BinaryMinHeap, BinaryMaxHeap, nlargest, nsmallest
-    export MutableBinaryHeap, mutable_binary_minheap, mutable_binary_maxheap
+    export MutableBinaryHeap, MutableBinaryMinHeap, MutableBinaryMaxHeap
     export heapify!, heapify, heappop!, heappush!, isheap
     export BinaryMinMaxHeap, popmin!, popmax!, popall!
 
--- a/src/heaps/binary_heap.jl
+++ b/src/heaps/binary_heap.jl
@@ -106,15 +106,19 @@ mutable struct BinaryHeap{T,Comp} <: AbstractHeap{T}
     comparer::Comp
     valtree::Vector{T}
 
-    function BinaryHeap{T,Comp}(comp::Comp) where {T,Comp}
-        new{T,Comp}(comp, Vector{T}())
-    end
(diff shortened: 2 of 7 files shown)


src/DataStructures.jl
```diff
@@ -33,8 +33,8 @@ module DataStructures
     export IntDisjointSets, DisjointSets, num_groups, find_root, in_same_set, root_union!
 
     export AbstractHeap, compare, extract_all!
-    export BinaryHeap, binary_minheap, binary_maxheap, nlargest, nsmallest
-    export MutableBinaryHeap, mutable_binary_minheap, mutable_binary_maxheap
+    export BinaryHeap, BinaryMinHeap, BinaryMaxHeap, nlargest, nsmallest
+    export MutableBinaryHeap, MutableBinaryMinHeap, MutableBinaryMaxHeap
```
oxinabox: still need to export the old deprecated names.
Til removed.

but can do so on own line, making it easy to remove them later.

src/heaps/binary_heap.jl
```diff
@@ -106,23 +106,31 @@ mutable struct BinaryHeap{T,Comp} <: AbstractHeap{T}
...
+BinaryMinHeap(xs::AbstractVector{T}) where T = BinaryMinHeap{T}(xs)
+BinaryMaxHeap(xs::AbstractVector{T}) where T = BinaryMaxHeap{T}(xs)
+
+# deprecated constructors
+                            
+binary_minheap(::Type{T}) where T = BinaryMinHeap{T}()
+binary_maxheap(::Type{T}) where T = BinaryMaxHeap{T}()
+binary_minheap(xs::AbstractVector{T}) where T = BinaryMinHeap(xs)
+binary_maxheap(xs::AbstractVector{T}) where T = BinaryMaxHeap(xs)
+
```
oxinabox: You shouldn't need these ones above.
The deprecased ones below should do it.

src/heaps/mutable_binary_heap.jl
```diff
@@ -155,23 +155,35 @@ mutable struct MutableBinaryHeap{VT, Comp} <: AbstractMutableHeap{VT,Int}
...
+                            
+MutableBinaryMinHeap(xs::AbstractVector{T}) where T = MutableBinaryMinHeap{T}(xs)
+MutableBinaryMaxHeap(xs::AbstractVector{T}) where T = MutableBinaryMaxHeap{T}(xs)
+
+# deprecated constructors
+mutable_binary_minheap(::Type{T}) where T = MutableBinaryMinHeap{T}()
+mutable_binary_maxheap(::Type{T}) where T = MutableBinaryMaxHeap{T}()
+mutable_binary_minheap(xs::AbstractVector{T}) where T = MutableBinaryMinHeap(xs)
+mutable_binary_maxheap(xs::AbstractVector{T}) where T = MutableBinaryMaxHeap(xs)
+
```
oxinabox: As above, should only have the deprecations. (which you have below)

src/DataStructures.jl
dillondaudert: The `@deprecate` macro exports the functions, so this should be OK.

oxinabox: Cool. I learn something new every day.

dillondaudert:
One thing that's maybe missing are adding `@test_deprecated` to the old constructors, which I can do if others feel that is a good idea (and it might address the very minor loss in coverage)

oxinabox:
Yes, I was wondering about that.
I'm note sure how well any of the deprecation stuff is working, in 1.0.

dillondaudert:
Any further thoughts on this? @ararslan @oxinabox 

I'm also curious if the new heap constructors (along with the minmax heap) will be tagged in a new release?

oxinabox:
I was waiting for the tests to be added to the deprecated methods.
Yes, we should tag a release, it does seem nice to have both this and the `MinMaxHeaps`  tagged together

oxinabox:
LGTM

Issue #474, opened 2019-01-03
Support Set operations on counters (because they are bags)


oxinabox:
related to https://github.com/JuliaCollections/DataStructures.jl/issues/133

We should support the operation ons on

`Accumulator`s that are supported on Sets.

https://en.wikipedia.org/wiki/Multiset#Basic_properties_and_operations

oxinabox:
Closed by #475

Issue #133, opened 2015-10-05
It should be mentioned in the Readme that a Counter is a Multiset/Bag


oxinabox:
Today I needed a [Multiset](https://en.wikipedia.org/wiki/Multiset)
I noted before that there was not on in Core or in this package.

I started to think about how I would implement one quickly/simply.
One implementation would be a dictionary, storing what is in the multiset and how many times it occurs.
But of-course that is just a Counter.

It should be in the readme so this package comes up when people google "Julia multiset"

oxinabox:
Though before making that change,
it makes sense for Accumulators (and thus counters) to implement `(==)` and `hash` since Sets do.

I can make a pull request for these changes.
I have the code for that -- it is just delgating it to the Accumulator.map since Dicts implement that.

rawls238:
@oxinabox feel free to make a PR for these changes

PR #475, opened 2019-01-03
Define multiset operations

oxinabox:
needs tests
needs docs

Solves #474 

when documented will solve #133 

https://en.wikipedia.org/wiki/Multiset#Basic_properties_and_operations

It is tempting to define `zero(A::Type(<:Accumulator)=A()`
which would give us a group (in the mathematical sense).
If we are willing to think of `merge` and `setdiff` as the same operation, but one of which creating the inverse element.

### DIFF
--- a/docs/src/accumulators.md
+++ b/docs/src/accumulators.md
@@ -52,7 +52,7 @@ reset!(a, x)     # remove a key x from a, and return its current value
 merge!(a, a2)    # add all counts from a2 to a1
 merge(a, a2)     # return a new accumulator/counter that combines the
                  # values/counts in both a and a2
-                 # `a1[v] + a2[v]` over all `v` in the universe
+                 # `a[v] + a2[v]` over all `v` in the universe
 ```
 
 `merge` is the multiset sum operation (sometimes written ⊎).
--- a/docs/src/accumulators.md
+++ b/docs/src/accumulators.md
@@ -70,7 +70,7 @@ Note that these operations will throw an error if the accumulator has negative o
 ```julia
 
 setdiff(a1, a2)          # The opposite of `merge` (i.e. multiset sum),
-                         # Returns `a1` with the count of items in `a2` removed, down to a minimum of zero
+                         # returns a new multiset with the count of items in `a2` removed from `a1`, down to a minimum of zero
                          # `max(a1[v] - a2[v], 0)` over all `v` in the universe
 
 
--- a/docs/src/accumulators.md
+++ b/docs/src/accumulators.md
@@ -62,7 +62,7 @@ merge(a, a2)     # return a new accumulator/counter that combines the
 An `Accumulator{T, <:Integer} where T` such as is returned by `counter`,
 is a [multiset](https://en.wikipedia.org/wiki/Multiset) or Bag, of objects of type `T`.
 If the count type is not an integer but a more general real number, then this is a form of fuzzy multiset.
-We support a number of operations supporting the use of `Accumulators` as multisets.
+We support a number of operations supporting the use of `Accumulator`s as multisets.
 
 
 Note that these operations will throw an error if the accumulator has negative or zero counts for any items.
--- a/docs/src/accumulators.md
+++ b/docs/src/accumulators.md
@@ -55,7 +55,7 @@ merge(a, a2)     # return a new accumulator/counter that combines the
                  # `a1[v] + a2[v]` over all `v` in the universe
 ```
 
-`merge` is the multiset sum (Sometimes written ⊎) operation.
+`merge` is the multiset sum operation (sometimes written ⊎).
 
 ## Use as a multiset
 
--- a/docs/src/accumulators.md
+++ b/docs/src/accumulators.md
@@ -55,4 +55,31 @@ merge(a, a2)     # return a new accumulator/counter that combines (sums) the
                  # values/counts in both a and a2.
 ```
 
+merge is the multiset sum (Sometimes written ⊎) operation.
 
+## Use as a MultiSet
+
+An `Accumulator{T, <:Integer}` such as is returned by `counter`, is a [multiset](https://en.wikipedia.org/wiki/Multiset) or Bag, of objects of type `T`.
+If the count type is not an integer but a more general real number,
+then this is a form of fuzzy multiset.
+We support a number of operations to support the use of `Accumulators` as multisets.
+
+
+Note that these multiset operations will throw an error if the accumulator has negative or zero counts for any items,.
+
+```julia
+
+setdiff(a1, a2)          # The opposite of `merge` (i.e. multiset sum),
+                         # Returns `a1` with the count of items in `a2` removed, down to a minimum of zero
+                         # max(a1[v] - a2[v], 0) over all `v` in universe
+
+
+union(a1, a2)            # multiset union (sometimes called maximum, or lowest common multiple)
+                         # returns a new multiset with the counts being the higher of those in `a1` or `a2`.
+                         # max(a1[v], a2[v]) over all `v` in universe
+
+intersection(a1, a2)     # multiset intersection (sometimes called infimum or greatest common divisor)
+                         # returns a new multiset with the counts being the lowest of those in `a1` or `a2`.
+                         # Note that this means things not occurring in both with be removed (count zero).
+                         # min(a1[v], a2[v]) over all `v` in universe
+```
--- a/docs/src/accumulators.md
+++ b/docs/src/accumulators.md
@@ -41,7 +41,6 @@ a[x]             # get the current value/count for x,
 
 a[x] = v         # sets the current value/count for `x` to `v`
 
-
 inc!(a, x)       # increment the value/count for x by 1
 inc!(a, x, v)    # increment the value/count for x by v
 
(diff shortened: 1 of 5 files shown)


oxinabox:
No longer needs tests.
Still needs docs.

oxinabox:
I want merge 
https://github.com/JuliaCollections/DataStructures.jl/pull/476
 first before I write the docs.
to decrease chance of conflicts

test/test_accumulator.jl
```diff
@@ -175,6 +175,32 @@
...
+            @test ∪(counter([1,3]), counter([2,2])) == counter([1,2,2,3])
+            @test ∪(counter([1,2,3]), counter(Int[])) == counter([1,2,3])
+        end
+ 
+        @testset "union" begin
+            @test ∩(counter([1,2,3]), counter([1,2,3])) == counter([1,2,3])
+            @test ∩(counter([1,2,3]), counter([1,2,2,3])) == counter([1,2,3])
+            @test ∩(counter([1,3]), counter([2,2])) == counter(Int[])
+            @test ∩(counter([1,2,3]), counter(Int[])) == counter(Int[])
+        end
```
nickrobinson251: change testset name "union" -> "intersect"

src/accumulator.jl
```diff
@@ -182,4 +181,55 @@ nsmallest(acc::Accumulator, n) = partialsort!(collect(acc), 1:n, by=last, rev=fa
...
+function Base.intersect!(a::Accumulator, b::Accumulator)
+    for (kb, vb) in b
+        vb > 0 || throw(MultiplicyException(kb, vb))
+        a[kb] = min(a[kb], vb)
+    end
+    # Need to do this bidirectionally, as anything not in both needs to be removed
+    for (ka,va) in a
+        a[ka] = min(b[ka], va)
+    
+        drop_nonpositive!(a, ka) # Drop any that ended up zero
```
nickrobinson251: Do we throw an Exception if any value `vb` in `b` is `<= 0`, but just silently drop any value `va` in `a` which is `<= 0`? That would make the operation non-commutative. Shouldn't we also throw an Exception in the second case, where a value `va` in `a` is`<= 0`?

docs/src/accumulators.md
```diff
@@ -45,3 +45,36 @@ pop!(a, x)       # remove a key x from a, and return its current value
...
+
+
+union(a1, a2)            # multiset union (sometimes called maximum, or lowest common multiple)
+                         # returns a new multiset with the counts being the higher of those in `a1` or `a2`.
+                         # max(a1[v], a2[v]) over all `v` in universe
+
+intersection(a1, a2)     # multiset intersection (sometimes called infimum or greatest common divisor)
+                         # returns a new multiset with the counts being the lowest of those in `a1` or `a2`.
+                         # Note that this means things not occurring in both with be removed (count zero).
+                         # min(a1[v], a2[v]) over all `v` in universe
```
nickrobinson251: "intersection" -> "intersect"

src/accumulator.jl
oxinabox: yes, it should.
I was mistaken.
The drop at the end is just to clean out any zeros i added the the first loop

docs/src/accumulators.md
```diff
@@ -45,3 +45,36 @@ pop!(a, x)       # remove a key x from a, and return its current value
 merge(a, a2)     # return a new accumulator/counter that combines the
                  # values/counts in both a and a2
 ```
+
+
+merge is the multiset sum (Sometimes written ⊎) operation.
```
omus: `merge` is the multiset sum (sometimes written ⊎) operation.

docs/src/accumulators.md
```diff
@@ -45,3 +45,36 @@ pop!(a, x)       # remove a key x from a, and return its current value
...
+
+## Use as a MultiSet
```
omus: You seem to be inconsistent in your casing of multiset.

docs/src/accumulators.md
```diff
@@ -45,3 +45,36 @@ pop!(a, x)       # remove a key x from a, and return its current value
...
+
+An `Accumulator{T, <:Integer}` such as is returned by `counter`, is a [multiset](https://en.wikipedia.org/wiki/Multiset) or Bag, of objects of type `T`.
```
omus: If you don't care about the second parametric type then maybe write as: `Accumulator{T}`

docs/src/accumulators.md
```diff
@@ -45,3 +45,36 @@ pop!(a, x)       # remove a key x from a, and return its current value
...
+If the count type is not an integer but a more general real number,
+then this is a form of fuzzy multiset.
+We support a number of operations to support the use of `Accumulators` as multisets.
+
+
+Note that these multiset operations will throw an error if the accumulator has negative or zero counts for any items,.
```
omus: `,.`

docs/src/accumulators.md
```diff
@@ -45,3 +45,36 @@ pop!(a, x)       # remove a key x from a, and return its current value
...
 ```
+
+
+merge is the multiset sum (Sometimes written ⊎) operation.
+
+## Use as a MultiSet
+
+An `Accumulator{T, <:Integer}` such as is returned by `counter`, is a [multiset](https://en.wikipedia.org/wiki/Multiset) or Bag, of objects of type `T`.
+If the count type is not an integer but a more general real number,
+then this is a form of fuzzy multiset.
```
omus: Why is this word wrapped so early while other lines are not wrapped

docs/src/accumulators.md
```diff
@@ -45,3 +45,36 @@ pop!(a, x)       # remove a key x from a, and return its current value
...
+We support a number of operations to support the use of `Accumulators` as multisets.
```
omus: `Accumulator`s

docs/src/accumulators.md
```diff
@@ -45,3 +45,36 @@ pop!(a, x)       # remove a key x from a, and return its current value
...
+
+
+Note that these multiset operations will throw an error if the accumulator has negative or zero counts for any items,.
+
+```julia
+
+setdiff(a1, a2)          # The opposite of `merge` (i.e. multiset sum),
+                         # Returns `a1` with the count of items in `a2` removed, down to a minimum of zero
+                         # max(a1[v] - a2[v], 0) over all `v` in universe
```
omus: max should be in backticks

docs/src/accumulators.md
```diff
@@ -45,3 +45,36 @@ pop!(a, x)       # remove a key x from a, and return its current value
...
+                         # max(a1[v], a2[v]) over all `v` in universe
+
+intersect(a1, a2)        # multiset intersection (sometimes called infimum or greatest common divisor)
+                         # returns a new multiset with the counts being the lowest of those in `a1` or `a2`.
+                         # Note that this means things not occurring in both with be removed (count zero).
+                         # min(a1[v], a2[v]) over all `v` in universe
+```
+
+
+
```
omus: Too much extra whitespace

src/accumulator.jl
```diff
@@ -182,4 +181,57 @@ nsmallest(acc::Accumulator, n) = partialsort!(collect(acc), 1:n, by=last, rev=fa
...
+## Multiset operations
+
+struct MultiplicyException{K,V} <: Exception
+    k::K
+    v::V
+end
+
+function Base.showerror(io::IO, err::MultiplicyException)
+    print(io, "When using an Accumulator as a Multiset, all elements must have positive multiplicy")
+    print(io, "element `$(err.k)` has multiplicy $(err.v)")
```
omus: No space between the two lines

src/accumulator.jl
```diff
@@ -182,4 +181,57 @@ nsmallest(acc::Accumulator, n) = partialsort!(collect(acc), 1:n, by=last, rev=fa
...
+        v > 0 || throw(MultiplicyException(k, v))
+        ret[k] -= v # defaults to zero if not found
+        drop_nonpositive!(ret, k)
+    end
+    return ret
+end
+
+Base.issubset(a::Accumulator, b::Accumulator) = all(b[k] >= v for (k, v) in a)
+
+Base.union(a::Accumulator, b::Accumulator) = Base.union!(copy(a), b)
```
omus: I'll note typically `union` methods are written as `union(::Accumulator, ::Accumulator...)`

src/accumulator.jl
```diff
@@ -182,4 +181,57 @@ nsmallest(acc::Accumulator, n) = partialsort!(collect(acc), 1:n, by=last, rev=fa
 @deprecate pop!(ct::Accumulator, x) reset!(ct, x)
 @deprecate push!(ct1::Accumulator, ct2::Accumulator) merge!(ct1,ct2)
 
+###########################################################
+## Multiset operations
+
+struct MultiplicyException{K,V} <: Exception
```
omus: Multiplicy -> Multiplicity?

test/runtests.jl
```diff
@@ -5,7 +5,7 @@ using Serialization
 
 import DataStructures: IntSet
 
-@test isempty(detect_ambiguities(Base, Core, DataStructures))
+@test [] == detect_ambiguities(Base, Core, DataStructures)
```
omus: I'm curious about this change

test/test_accumulator.jl
```diff
@@ -175,6 +175,32 @@
...
+
+
+
+
+    end
+
```
omus: Add test for `reset!` return value

omus: Add tests where `MultiplicyException` are expected

omus: Add `issubset` tests

test/runtests.jl
oxinabox: Now, when you have ambigurities in there,
the `@test` macro prints what they are.

Realyl should have been its own commit (if it wasn't already)

docs/src/accumulators.md
omus: Ok, I think maybe you should write this as `Accumulator{T, V<:Integer}` because at first I thought you had a typo and meant `Accumulator{T<:Integer}`

omus: Alternatively `Accumulator{T, <:Integer} where T` would also indicate it was done on purpose to me

test/runtests.jl
omus: Can you open an issue for this against Julia? I expanded the `@test` macro to also show `isapprox` and `isequal` and I think `isempty` is common enough we can show the contents there too.

oxinabox: https://github.com/JuliaLang/julia/issues/30651

docs/src/accumulators.md
```diff
@@ -51,8 +50,36 @@ dec!(a, x, v)    # decrement the value/count for x by v
...
 ```
 
+`merge` is the multiset sum (Sometimes written ⊎) operation.
+
+## Use as a multiset
+
+An `Accumulator{T, <:Integer} where T` such as is returned by `counter`,
+is a [multiset](https://en.wikipedia.org/wiki/Multiset) or Bag, of objects of type `T`.
+If the count type is not an integer but a more general real number, then this is a form of fuzzy multiset.
+We support a number of operations supporting the use of `Accumulators` as multisets.
```
omus:
```suggestion
We support a number of operations supporting the use of `Accumulator`s as multisets.
```

src/accumulator.jl
```diff
@@ -182,4 +181,59 @@ nsmallest(acc::Accumulator, n) = partialsort!(collect(acc), 1:n, by=last, rev=fa
...
+###########################################################
+## Multiset operations
+
+struct MultiplicityException{K,V} <: Exception
+    k::K
+    v::V
+end
+
+function Base.showerror(io::IO, err::MultiplicityException)
+    print(io, "When using an Accumulator as a Multiset, all elements must have positive multiplicity")
```
omus:
```suggestion
    print(io, "When using an `Accumulator` as a multiset, all elements must have positive multiplicity")
```

oxinabox:
Apparently this has been ready to merge for 10 days.
@omus  you want to give it a glance, see that everything was addressed?

Good thing you told me to add those tests, some of the behavour was wrong (or at least weird) before

test/runtests.jl
omus: I'll note this has now been addressed and you can use `isempty`

PR #481, opened 2019-01-22
At getindex for BinaryHeap

dillondaudert:
Both `MutableBinaryHeap` and `BinaryMinMaxHeap` define `getindex`. This PR adds the same to `BinaryHeap`.

### DIFF
--- a/src/heaps/binary_heap.jl
+++ b/src/heaps/binary_heap.jl
@@ -152,3 +152,5 @@ Returns the element at the top of the heap `h`.
 @inline top(h::BinaryHeap) = h.valtree[1]
 
 pop!(h::BinaryHeap{T}) where {T} = _binary_heap_pop!(h.comparer, h.valtree)
+                            
+getindex(h::BinaryHeap, i::Int) = h.valtree[i]


oxinabox:
What does it mean to index into a heap?

dillondaudert:
That's a good question. I didn't think about the semantics for heaps too closely, my interest was just in making accessing the underlying storage unified across all existing heaps. Right now, it's possible to write h[i] for a mutable heap but one must write h.valtree[i] for other types.

oxinabox:
@CarloLucibello added this in https://github.com/JuliaCollections/DataStructures.jl/pull/269

Maybe he can speak to the semantics.


To me this feels alot like exposing `getindex` on a `Set`.
Which we do not do.
Heaps are not indexed structures.

dillondaudert:
I agree the indexing isn't truly on the heap itself, but the underlying (implementation-dependent) data storage. It just so happens in this case all the heaps have indexable storage.

It would be a breaking change, but unifying that across heaps is something I think would be nice, maybe as an indexable property (e.g. `heap.storage[i]`).

PR #483, opened 2019-02-27
Update docs/make.jl to include mutable_linked_list.md

c-p-murphy:
Missed that this had to be updated in #450.


oxinabox:
Thanks

PR #484, opened 2019-03-05
Add support for range indexing into MutableLinkedLists

c-p-murphy:
Thought this might be useful to have.

### DIFF
--- a/test/test_mutable_list.jl
+++ b/test/test_mutable_list.jl
@@ -41,14 +41,7 @@
                         l1 = MutableLinkedList{Int32}(1:i...)
                         io = IOBuffer()
                         @test sprint(io -> show(io, iterate(l1))) == "(1, DataStructures.ListNode{Int32}(2))"
-<<<<<<< HEAD
-<<<<<<< HEAD
-                        @test sprint(io -> show(io, iterate(l, l.node.next.next))) == "(2, DataStructures.ListNode{Int64}(3))"
-=======
->>>>>>> 5767f1a... Fix iterate test
-=======
-                        @test sprint(io -> show(io, iterate(l, l.node.next.next))) == "(2, DataStructures.ListNode{Int32}(3))"
->>>>>>> f7806fd... fix test error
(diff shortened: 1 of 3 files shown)


test/test_mutable_list.jl
```diff
@@ -1,13 +1,21 @@
...
+    @testset "empty list / listnode" begin
         l1 = MutableLinkedList{Int}()
         @test MutableLinkedList() == MutableLinkedList{Any}()
+        @test iterate(l1) === nothing
         @test isempty(l1)
         @test length(l1) == 0
         @test lastindex(l1) == 0
         @test collect(l1) == Int[]
         @test eltype(l1) == Int
+        n1 = DataStructures.ListNode{Int}()
```
c-p-murphy: This last part is changed again in the next commit, so feel free to ignore.

c-p-murphy:
Tweaked a few things, but I think it's good to go now.

c-p-murphy:
Tweaked a few things, but I think it's good to go now.

c-p-murphy:
I tweaked a few things, but I think this is all set now.

test/test_mutable_list.jl
```diff
@@ -110,12 +131,32 @@
...
+                    l = MutableLinkedList{Int32}(1:n...)
+                    let
+                        io = IOBuffer()
+                        @test sprint(io -> show(io, l.node.next)) == "$(typeof(l.node.next))($(l.node.next.data))"
+                        begin io1 = IOBuffer()
+                              write(io1, "MutableLinkedList{Int32}(");
+                              write(io1, join(l, ", "));
+                              write(io1, ")")
+                              io1.ptr = 1
+                              @test sprint(io -> show(io, l)) == read(io1, String)
```
kmsquire: Thanks for adding the tests. 

To seek back to the beginning of an IOBuffee, it's generally better to use `seekstart()`, rather than seeing `ptr`directly.

Otherwise, looks go to me!

PR #485, opened 2019-03-08
Add Fenwick Tree

eulerkochy:
This PR adds the feature of a fairly simple, but quite powerful DS.


src/fenwick.jl
```diff
@@ -0,0 +1,37 @@
+mutable struct FenwickTree{T}
+    BIT::Vector{T}
```
oxinabox: Should be lower-case

src/fenwick.jl
```diff
@@ -0,0 +1,37 @@
...
+    n::Integer
```
oxinabox: For speed this should either be a concrete type (e.g. `Int`) or a type param.
I am leaning towards `Int`, there is afact no reason someone would want to use e.g. a `BigInt`.

src/fenwick.jl
```diff
@@ -0,0 +1,37 @@
...
+end
+
+FenwickTree{T}() where T = FenwickTree{T}(0)
+FenwickTree{T}(n::Integer) where T = FenwickTree{T}(fill(zero(T), n), n)
```
oxinabox: Use `zeros(T, n)` rather than `fill`

src/fenwick.jl
```diff
@@ -0,0 +1,37 @@
+mutable struct FenwickTree{T}
+    BIT::Vector{T}
+    n::Integer
+end
+
+FenwickTree{T}() where T = FenwickTree{T}(0)
```
oxinabox: Should there be a constructor that takes a iterator if items and fills the tree?

src/fenwick.jl
```diff
@@ -0,0 +1,37 @@
...
+FenwickTree{T}(n::Integer) where T = FenwickTree{T}(fill(zero(T), n), n)
+
+bit(F::FenwickTree{T}) where T = F.BIT
+size(F::FenwickTree{T}) where T = F.n
+
+function update(F::FenwickTree{T}, ind::Int, val::T) where T
```
oxinabox: This should be called `update!` as it is mutating

src/fenwick.jl
```diff
@@ -0,0 +1,37 @@
+mutable struct FenwickTree{T}
+    BIT::Vector{T}
+    n::Integer
+end
+
+FenwickTree{T}() where T = FenwickTree{T}(0)
+FenwickTree{T}(n::Integer) where T = FenwickTree{T}(fill(zero(T), n), n)
+
+bit(F::FenwickTree{T}) where T = F.BIT
+size(F::FenwickTree{T}) where T = F.n
```
oxinabox: I think this should be `length` not `size`.

src/fenwick.jl
```diff
@@ -0,0 +1,37 @@
...
+
+function update(F::FenwickTree{T}, ind::Int, val::T) where T
+    i = ind
+    N = F.n
```
oxinabox: Casing

src/fenwick.jl
```diff
@@ -0,0 +1,37 @@
...
+        i += i&(-i)
+    end
+end
+
+function update(F::FenwickTree{T}, left::Int, right::Int, val::T) where T
+    update(F, left, +val)
+    update(F, right + 1, -val)
+end
+
+function getsum(F::FenwickTree{T}, ind::Int) where T
```
oxinabox: Why not just call this `sum`

src/fenwick.jl
eulerkochy: Umm, no. I thought about it, but the tree is actually is the representation of an array, you can't just have a random array be a BIT and then expect some proper functions out of it. That being said, there is always a scope for its inclusion. Maybe, I'll include it to make it look more generic.

src/fenwick.jl
eulerkochy: Ok!

src/fenwick.jl
eulerkochy: cool!

src/fenwick.jl
eulerkochy: corrected!

src/fenwick.jl
eulerkochy: ok!

src/fenwick.jl
eulerkochy: :+1:

src/fenwick.jl
eulerkochy: True

eulerkochy:
Bump

oxinabox:
@ararslan , @kmsquire 
could one of you take a look at this?
I am going to be busy for the next few days

ararslan:
I don't know that I'd be the best one to review this, as I've never even heard of this data structure.

SyxP:
I am very familiar with this structure. This is used in Competitive Programming, but not utilized much elsewhere. I feel that this very stand alone structure should just split itself into its own library and be very self-contained. I have a looked through the code and here are my thoughts:

1) The initialise `FenwickTree(::Vector{T}) where T` method directly creates an array. This might be counterintuitive. For example, you initialise with `A = FenwickTree([1,2,3])`, you might expect `sum(A, 2)` to return `3` but it returns `2`.

2) The method sum should be called `subarraysum` or something. As is, it seems a little weird to have it extending a method in main doing something entirely different.

3) You should annotate the `@boundscheck` so that `@inbounds` optimisation will propogate through.

4) There is no `getindex` method. Hence, `A[1]` as per above will fail. 

5) There is an off-by-one error in the range updating. `update!(A,1,3,1)` as per above will fail. 

Ill iterate that I think its best for this to be on its own self-contained library. Perhaps even a Competitive Programming library.

eulerkochy:
Thanks for the review. Ok, I'll look into those cases. Even I feel these things should be split out for a `CP` library in Julia. What's the general opinion towards these cause?

ararslan:
> Ill iterate that I think its best for this to be on its own self-contained library.

Why? Seems reasonable to me for it to live alongside other data structures. Surely _someone_ uses this outside of competative programming.

SyxP:
1) As in, there is an increasing complaint that `DataStructure.jl` is a big dependency and perhaps it is best for individual parts to be packaged separately. Something as disjoint as this would be a good start.

2) Well, it is possible that it is used outside of Competitive Programming. But to my knowledge, the common applications cited (Arithmetic Coding/Compression) uses it over a different semiring. Namely, instead of asking for the sum of `A[i:j]` it is the exclusive disjunction. The code as such would be too inflexible for such a case.

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+    FenwickTree(arr::AbstractArray) 
+    
+Constructs a [`FenwickTree`](https://en.wikipedia.org/wiki/Fenwick_tree) from `arr`.
+ 
+"""
+function FenwickTree(a::T) where T <: AbstractVector
+    U = eltype(a)
+    n = size(a)[1]
+    bit = FenwickTree{U}(n)
+    @boundscheck for i = 1:n
```
ararslan: This should still be `@inbounds`. `@boundscheck` is used for error checking, e.g.
```julia
@boundscheck condition || error
@inbounds for ...
```

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+FenwickTree{T}() where T = FenwickTree{T}(0)
+FenwickTree{T}(n::Integer) where T = FenwickTree{T}(zeros(T, n), n)
+
+"""
+    FenwickTree(arr::AbstractArray) 
+    
+Constructs a [`FenwickTree`](https://en.wikipedia.org/wiki/Fenwick_tree) from `arr`.
+ 
+"""
+function FenwickTree(a::T) where T <: AbstractVector
```
ararslan:
```suggestion
function FenwickTree(a::AbstractVector{U}) where U
```

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+    U = eltype(a)
```
ararslan:
```suggestion
```

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+    n = size(a)[1]
```
ararslan:
```suggestion
    n = length(a)
```

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+    bit = FenwickTree{U}(n)
+    @boundscheck for i = 1:n
+        update!(bit, i, a[i])
+    end
+    bit
+end
+
+bit(F::FenwickTree{T}) where T = F.bit
```
ararslan:
```suggestion
bit(F::FenwickTree) = F.bit
```
Type parameter isn't used

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+length(F::FenwickTree{T}) where T = F.n
```
ararslan:
```suggestion
length(F::FenwickTree) = F.n
```

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+"""
+    update!(F::FenwickTree{T}, ind, val::T)
+
+Update the value of the [`FenwickTree`] by `val` from the index `ind` upto the length of the Fenwick Tree.
+
+""" 
+function update!(F::FenwickTree{T}, ind::Int, val::T) where T
+    i = ind
+    n = F.n
+    (i in 1:n) || throw(DomainError(i, "$i should be in between 1 and $n"))
```
ararslan:
```suggestion
    @boundscheck 1 <= i <= n || throw(ArgumentError("$i should be in between 1 and $n"))
```

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+ 8
+julia> sum(f, 6)
+ 5
+```
+"""
+function sum(F::FenwickTree{T}, ind::Int) where T
+    sum = zero(T)
+    i = ind
+    n = F.n
+    (i in 1:n) || throw(DomainError(i, "$i should be in between 1 and $n"))
```
ararslan:
```suggestion
    @boundscheck 1 <= i <= n || throw(ArgumentError("$i should be in between 1 and $n"))
```

eulerkochy:
Bump

eulerkochy:
Can anyone review this? It's has been inactive for quite sometime!

kmsquire:
Will try to take a look Monday morning PST.

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
+"""
+    FenwickTree{T}(n)
+    
+Constructs a [`FenwickTree`](https://en.wikipedia.org/wiki/Fenwick_tree) of length `n`.
+ 
+"""
+mutable struct FenwickTree{T}
+    bit::Vector{T}
```
kmsquire: I found that I was confused by this name being used everywhere.  Googling around, it seems this might be an abbreviation of "binary index tree" (which is another name for a Fenwick tree).

In this package, most other data structures use more verbose names, I would suggest renaming this `binary_index_tree`, or possibly `bi_tree` (but add a comment describing what it means).

`store` is another option (e.g., as used by the `Deque` data structure).

It is true that other data structures here sometimes use short names--e.g., [here](https://github.com/JuliaCollections/DataStructures.jl/blob/69c867fcccc8a8210b4d841ba0ea0b4e9487a660/src/sorted_dict.jl#L5).  
But the type is also specified right after that, making it easy to understand the acronym.

The term, `bit`, though, is also a common computing term, and when I was looking at it, all I could think of is, what do "bits" have to do with Fenwick trees (other than the fact that it's implemented on a computer).

src/fenwick.jl
kmsquire: Doesn't seem to have been addressed

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+
+"""
+    FenwickTree(arr::AbstractArray) 
+    
+Constructs a [`FenwickTree`](https://en.wikipedia.org/wiki/Fenwick_tree) from `arr`.
+ 
+"""
+function FenwickTree(a::AbstractVector{U}) where U
+    n = length(a)
+    bit = FenwickTree{U}(n)
```
kmsquire: I would also rename `bit` here.  E.g., `ft` would be fine, or `tree`.

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+    @inbounds for i = 1:n
+        update!(bit, i, a[i])
+    end
+    bit
+end
+
+bit(F::FenwickTree{T}) where T = F.bit
```
kmsquire: How frequent is it that a user would want to access the underlying tree?  It seems that the prefix sum is the most important aspect of the tree.

Unless you expect this to be accessed quite frequently by users, I would suggest removing it.  The underlying data store can be accessed using `f.binary_index_tree` (or whatever it ends up being called).

kmsquire: Also, in julia, types are usually CamelCase, and variables are lower case. So `F` should be `f` or `ft` (here and in most of the function definitions below).

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+length(F::FenwickTree{T}) where T = F.n
+
+"""
+    update!(F::FenwickTree{T}, ind, val::T)
+
+Update the value of the [`FenwickTree`] by `val` from the index `ind` upto the length of the Fenwick Tree.
+
+""" 
+function update!(F::FenwickTree{T}, ind::Int, val::T) where T
```
kmsquire: I would suggest renaming this to `inc!`, for consistency with accumulators [here](https://github.com/JuliaCollections/DataStructures.jl/blob/69c867fcccc8a8210b4d841ba0ea0b4e9487a660/src/accumulator.jl#L64-L70).

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+    end
+end
+
+"""
+    update!(F::FenwickTree{T}, left, right, val::T)
+
+Update the value of the [`FenwickTree`] by `val` from the index `left` upto the index `right`.
+
+"""    
+function update!(F::FenwickTree{T}, left::Int, right::Int, val::T) where T
```
kmsquire: If updated to `inc!`, consider changing this to 

```julia
function inc!(ft::FenwickTree{T}, range::AbstractUnitRange, val::T) where T
```

(with appropriate changes to the function below.)

You also might consider dropping the `T` and instead converting val to type `T` below, to give a user more flexibility.

See, e.g., [here](https://github.com/JuliaLang/julia/blob/master/base/dict.jl#L371-L372) for an example of this usage.

src/fenwick.jl
kmsquire: If you do so, consider also adding a `dec!` function.

src/fenwick.jl
```diff
@@ -0,0 +1,96 @@
...
+julia> update!(f, 1, 4, 3)
+julia> sum(f, 1)
+ 3
+julia> sum(f, 3)
+ 8
+julia> sum(f, 6)
+ 5
+```
+"""
+function sum(F::FenwickTree{T}, ind::Int) where T
```
kmsquire: I agree with @SyxP's comment that `sum` is not the best name for this function.  

`cumsum` is a Julia function which calculates the cumulative sum, but it returns an array of all values up to the end of an array. 

I would suggest `prefixsum()`, which is used in the description you linked to on Wikipedia.

eulerkochy:
@kmsquire all fine now?

src/fenwick.jl
```diff
@@ -0,0 +1,114 @@
...
+    end
+end
+
+"""
+    dec!(ft::FenwickTree{T}, ind, val)
+
+Decreases the value of the [`FenwickTree`] by `val` from the index `ind` upto the length of the Fenwick Tree.
+
+""" 
+dec!(ft::FenwickTree{T}, ind::Int, val) where T = inc!(ft, ind, -val)
```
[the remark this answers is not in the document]
kmsquire: Minor: also, here, `ind` should probably be typed as an `Integer`.

(Rule of thumb: in `structs`, it's best to be as specific (and concrete) as possible with types.  In function definitions, types should be as loose as possible.)

src/fenwick.jl
```diff
@@ -0,0 +1,114 @@
...
+
+"""
+    inc!(ft::FenwickTree{T}, range::AbstractUnitRange, val)
+
+Increases the value of the [`FenwickTree`] by `val` from the indices in `range`.
+
+"""    
+function inc!(ft::FenwickTree{T}, range::AbstractUnitRange, val) where T
```
kmsquire: I suggested this change, but I'm wondering if updating a range of values in a Fenwick tree is a common operation?

Looking at its use in the tests, my intuition says that this should be equivalent to
```
for idx in range
    inc!(ft, idx, val)
end
```
which is clearly not the case.

Given this is actually equivalent to an `inc!` followed by a `dec!`, how about renaming this function `incdec!` (and not using a range).  I.e.,

```suggestion
function incdec!(ft::FenwickTree{T}, left::Integer, right::Integer, val)
```

The intuition here is that the value at `left` is being increased by `val`, which increases the prefix sum for everything, but then the value at `right` is decreased by `val`, leaving the remaining elements unchanged.

If you follow (and agree with) this intuition, then that actually changes the implementation below slightly (i.e., the value at `right` would be decreased / decremented instead of `right +1`).

(I'm open to other ideas...)

src/fenwick.jl
```diff
@@ -0,0 +1,114 @@
...
+
+length(ft::FenwickTree{T}) where T = ft.n
+
+"""
+    inc!(ft::FenwickTree{T}, ind, val)
+
+Increases the value of the [`FenwickTree`] by `val` from the index `ind` upto the length of the Fenwick Tree.
+
+""" 
+function inc!(ft::FenwickTree{T}, ind::Int, val) where T
```
kmsquire: Is it reasonable for `val` to default to `1` here?  Or is it more typical to increment an index by an arbitrary amount?

test/test_fenwick.jl
```diff
@@ -0,0 +1,46 @@
+@testset "Fenwick Tree" begin
+    @testset "initialisation" begin
+        f1 = FenwickTree{Int}()
+        @test f1.bi_tree == Int[]
+        @test length(f1) == 0
+        
+        f2 = FenwickTree{Float64}(5)
+        z = zero(Float64)
+        @test f2.bi_tree == [z, z, z, z, z]
```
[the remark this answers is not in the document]
ararslan: Or even
```suggestion
        @test f2.bi_tree == zeros(5)
```

src/fenwick.jl
```diff
@@ -0,0 +1,114 @@
...
+    
+Constructs a [`FenwickTree`](https://en.wikipedia.org/wiki/Fenwick_tree) of length `n`.
+ 
+"""
+mutable struct FenwickTree{T}
+    bi_tree::Vector{T} #bi_tree is shorthand for Binary Indexed Tree, an alternative name for Fenwick Tree
+    n::Integer
+end
+
+FenwickTree{T}() where T = FenwickTree{T}(0)
```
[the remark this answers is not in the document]
eulerkochy: 1.  Cool
2. This seems to be a great addition to the power of Fenwick Tree.
3. Sure!

src/fenwick.jl
eulerkochy: Umm, the range update above as you've mentioned(`for idx in range`) is an `O((r - l + 1) * log n)` operation, whereas what I implemented here is `O(logn)` which is much better. The intuition is that of a [Heaviside step function](https://en.wikipedia.org/wiki/Heaviside_step_function), you add `val` from `left` to `Inf`, and subtract `val` from `right + 1` to `Inf`, what remains after this is `+val` from `left` to `right`.

As for `incdec!`, this is just opposite of what I stated above.

Read [this](https://ocw.mit.edu/courses/mathematics/18-03sc-differential-equations-fall-2011/unit-iii-fourier-series-and-laplace-transform/step-and-delta-functions-integrals-and-generalized-derivatives/MIT18_03SCF11_s24_1text.pdf). The exact thing I'm referring to is the `Box function`.

src/fenwick.jl
eulerkochy: Sure, any default value would suffice, `1` seems pretty reasonable.

src/fenwick.jl
kmsquire: I think I'm using a different mental model of Fenwick trees than you, so please bear with me.  

From reading about Fenwick trees, it seems that their purpose is to keep counts of numbers in such a way that it is easy to get the cumulative sum (or prefix sum) (though not necessarily the original count).  So if I'm keeping track counts of 5 numbers, I might have 

```julia
x = [1, 0, 0, 3, 2]   # 1 one, 3 fours, and 2 fives
cumsum(x) # = [1, 1, 1, 4, 6]
```

I can then increment the third element:
```julia
x[3] += 1  # x = [1, 0, 1, 3, 2]
cumsum(x) # = [1, 1, 2, 5, 7]
```

So what I see a Fenwick tree doing is exactly the same, except that it calculates/updates individual values of the cumulative sum more efficiently (and doesn't keep track of the actual counts).

```julia
ft = FenwickTree([1, 0, 0, 3, 2])
[ft[i] for i = 1:5]  # [1, 1, 1, 4, 6]
inc!(ft, 3, 1);
[ft[i] for i = 1:5]  # [1, 1, 2, 5, 7]
```

This is why `inc!` and `dec!` make sense to me (they update the underlying counts), and why I proposed `incdec!` as a way to updating the cumulative sum (or prefix sum) for a range in the middle of the array.

From your response, it seems that in your mental model, you're trying to manipulate the cumulative sums directly.  Can you explain why viewing the data structure in this way is useful?

eulerkochy: First of all, I'm definitely in for adding the functionality of `incdec!`, but what I'm suggesting is just a operational benefit, nothing more nothing less. If range updates are not there as per my model, complexity of the operation, i.e update each index individually is an expensive operation, additionally what I'm suggesting is making a sort of `arr[l : r] += val` , i know this thing doesn't exist in Julia, but there's no harm in introducing it. Actually the use case becomes more evident, when we have range queries, but for that we need another manipulation. I suggest we scrap this functionality, 'cuz we are introducing too much complications here. Instead, we add the `incdec!`

kmsquire: I think my main concern with the current implementation is this:

```julia
ft = FenwickTree([1, 0, 0, 3, 2])
inc!(ft, 3, 1);
[ft[i] for i = 1:5]  # [1, 1, 2, 5, 7]
```
vs.

```julia
ft = FenwickTree([1, 0, 0, 3, 2])
inc!(ft, 3:5, 1)
[ft[i] for i = 1:5]  # [1, 1, 2, 5, 7]
```

These two operations do the same thing with the current code, but I don't think they should.

(BTW, adding a value to a range of a Vector is easy in Julia (though you need to use `.+`:

```julia
julia> x = [1, 0, 0, 3, 2]
5-element Array{Int64,1}:
 1
 0
 0
 3
 2

julia> x[2:4] .+= 1; x
5-element Array{Int64,1}:
 1
 1
 1
 4
 2
```
)

eulerkochy: Umm, I can't seem to explain the behavior right now. Like, I've spent more than 1 hr trying to understand why it's happening, but now I've understood. The output will be same, 'cuz you're doing the same operation. Please read the docstring. 
```
ft = FenwickTree([1, 0, 0, 3, 2])
inc!(ft, 3:4, 1);
[ft[i] for i = 1:5] #[1, 1, 2, 5, 6]
```

kmsquire: Right.  I do understand why they're doing the same thing in the current code.  But I don't think they should do the same thing. :-)

For consistency,

```julia
inc!(ft, 3:4, 1);
```
should be equivalent to
```julia
inc!(ft, 3, 1);
inc!(ft, 4, 1);
```
And it obviously isn't.  But, it is equivalent to 
```julia
inc!(ft, 3, 1);
dec!(ft, 5, 1);
```
Or, if you were to add `incdec!`, 
```julia
incdec!(ft, 3, 5, 1)
```

test/test_fenwick.jl
```diff
@@ -0,0 +1,46 @@
+@testset "Fenwick Tree" begin
+    @testset "initialisation" begin
+        f1 = FenwickTree{Int}()
+        @test f1.bi_tree == Int[]
+        @test length(f1) == 0
```
kmsquire: You've removed this constructor, so this set of tests should be removed.

test/test_fenwick.jl
```diff
@@ -0,0 +1,46 @@
...
+        
+        f2 = FenwickTree{Float64}(5)
+        z = zero(Float64)
```
kmsquire:
```suggestion
```

kmsquire:
Hi @eulerkochy, just to summarize where I think we are right now:

1. There are 4 outstanding "suggestions" that you should apply to your code (either here or elsewhere), which update the code to make it more idiomatic (and correct one type issue)
2. With regard to `inc!` and `dec!` on ranges, you can either remove those functions, or add `incdec!`, as discussed earlier.

WIth those two changes, I believe this can go in.  Thanks!

eulerkochy:
Ok, working on this right away!

eulerkochy:
@kmsquire since we have a monotonic `prefixsum` array, we can also add a `lower_bound` function for finding the lowest index having the `prefixsum` greater than or equal to a particular value. What do you think about this?

src/fenwick.jl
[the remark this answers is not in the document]
eulerkochy: @kmsquire as for point 3, we should replace the `ArgumentError` with this modified implementation, right?

eulerkochy:
Well, I've been wondering about the implementation of `resize!` function for `FenwickTree`, it seems that I can't seem to keep the monotonicity property of `FenwickTree`. Just incrementing the size for the sake of increment destroys the beauty of this DS. @kmsquire advice?

eulerkochy:
Additionally, imo the best way out is to allow this implementation and then make a separate PR for the implementation with the `suffixsum` that allows for dynamic update in size, as mentioned in the [article](https://stackoverflow.com/questions/21995930/dynamic-i-e-variable-size-fenwick-tree).

src/fenwick.jl
```diff
@@ -0,0 +1,95 @@
+"""
+    FenwickTree{T}(n)
+    
+Constructs a [`FenwickTree`](https://en.wikipedia.org/wiki/Fenwick_tree) of length `n`.
+ 
+"""
+mutable struct FenwickTree{T}
```
kmsquire:
```suggestion
struct FenwickTree{T}
```

One more suggested change: let's make the struct immutable.  This still allows the `bi_tree` to be modified, but it won't allow `n` to change.

src/fenwick.jl
```diff
@@ -0,0 +1,95 @@
+"""
+    FenwickTree{T}(n)
+    
+Constructs a [`FenwickTree`](https://en.wikipedia.org/wiki/Fenwick_tree) of length `n`.
+ 
+"""
+struct FenwickTree{T}
+    bi_tree::Vector{T} #bi_tree is shorthand for Binary Indexed Tree, an alternative name for Fenwick Tree
+    n::Integer
```
kmsquire:
```suggestion
    n::Int
```

kmsquire:
> Additionally, imo the best way out is to allow this implementation and then make a separate PR for the implementation with the suffixsum that allows for dynamic update in size, as mentioned in the article.

I think that sounds fine.  

Almost there with this one: Can you add a test for `dec!` and `incdec!`?  After that, this can be merged.

eulerkochy:
Cool.

src/DataStructures.jl
```diff
@@ -28,6 +28,8 @@ module DataStructures
     export classified_lists, classified_sets, classified_counters
 
     export IntDisjointSets, DisjointSets, num_groups, find_root, in_same_set, root_union!
+    
+    export FenwickTree, length, inc!, dec!, prefixsum
```
kmsquire: Tests failed--need to add `incdec!` here.

kmsquire:
@eulerkochy, thanks for the contribution, and for your patience and perseverance!

Issue #482, opened 2019-02-23
`IntPriorityQueue` for efficient priority queues with integer keys 


SohamTamba:
I will be creating a datastructure `IntPriorityQueue` which is a `PriorityQueue` which excepts keys from `1:n` (`n` is the input) similar to [IntDisjointSets](https://github.com/JuliaCollections/DataStructures.jl/blob/master/src/disjoint_set.jl).
This is required to increase the performance in [this PR](https://github.com/JuliaGraphs/LightGraphs.jl/pull/1077#issuecomment-460020615) by replacing the [index hash table](https://github.com/JuliaCollections/DataStructures.jl/blob/b11a10fba4d8a8e0e7eba569bd36c816fa948450/src/priorityqueue.jl#L28) with an array.

Should I PR the code to `DataStructures.jl`?

kmsquire:
I think that would be fine. If you post and don't get a response within a day or two, send a gentle bump.🙂

SyxP:
By the way, I think this would only be efficient on dense graphs. I'm not entirely sure whether this would actually be an improvement on sparse graphs.

kmsquire:
If the interface is similar to the existing interface, it's usually nice to have different data structures and algorithms available for comparison for any given problem.

SohamTamba:
@SyxP is right and integer priority queue did not improve performance as much as I thought it would.
[See Comment](https://github.com/JuliaGraphs/LightGraphs.jl/pull/1077#issuecomment-468908467)

In case anyone is interested, I have a rough implementation of `int_PriorityQueue` which can be found [here](https://github.com/SohamTamba/LightGraphs.jl/blob/Int_PQ/src/DataStructures/priorityqueue.jl).

SyxP:
I feel that this should still be merged in. (Even better, allow the user a control over the type of dictionary used to hash the entries in the PQ, rather than just a Dict as in the current implementation.) This could have uses outside of Dijkstra's algorithm.

oxinabox:
am leaving this open for a bit.

The Int_PQ had noticably less alocations than other methods.

eulerkochy:
Anyone working on this?

SohamTamba:
> am leaving this open for a bit.
> 
> The Int_PQ had noticably less alocations than other methods.
> So worth thinking about.

The allocations are fewer because the `index` array is allocated statically, instead of a dynamic hash table.

Sorry for the late response, didn't see the notification until recently

[...]

PR #487, opened 2019-03-13
IntPQ

SohamTamba:
PR for Issue #482 

I used Inheritance to maximise code reuse.
Currently, Julia allows inheritance from Abstract type only so there is still more duplicate code than ideal.

I did not add alot of tests for IntPQ because there might be a way to reuse the code in the tests.
Something like `for PQ in [PriorityQueue, IntPriorityQueue]`


src/priorityqueue.jl
```diff
@@ -48,26 +47,121 @@ struct PriorityQueue{K,V,O<:Ordering} <: AbstractDict{K,V}
...
         pq
     end
 end
 
+"""
+    IntPriorityQueue{K, V}([ord])
+
+PriorityQueue that accepts integer in 1:n as keys.
+"""
+struct IntPriorityQueue{K,V,O<:Ordering} <: AbstractPriorityQueue{K,V,O}
```
oxinabox: I feel like:

```suggestion
struct IntPriorityQueue{V,O<:Ordering} <: AbstractPriorityQueue{Int,V,O}
```

src/priorityqueue.jl
```diff
@@ -48,26 +47,121 @@ struct PriorityQueue{K,V,O<:Ordering} <: AbstractDict{K,V}
...
+    xs::Array{Pair{K,V}, 1}
```
oxinabox:
```suggestion
    xs::Array{Pair{Int,V}, 1}
```

oxinabox:
I have not yet done a full review.

src/priorityqueue.jl
SohamTamba: That's a good point since keys must be in `1:n`.
A better way might be to enforce it might be `K<:Integer`.

Or `K<:UnsignedInteger`

SohamTamba:
Coverage is fine when I run locally

Issue #387, opened 2018-05-25
delete! in sorted containers incompatible with Dict and Set


StephenVavasis:
The statement `delete!(d,k)`, where `d` is a `Dict` or `Set` and `k` is a key,  does not throw a key error even when `k` is missing from `d`.  (In this case,  it is a no-op.)  On the other hand, SortedSet and SortedDict, throw a key error in this circumstance.  Because these packages have been around for awhile, possibly someone is counting on this incompatible behavior, so I don't know whether to modify these containers to bring them in line with Dict.  One subtlety: `Dict` and `Set` are not necessarily typed, whereas SortedDict and SortedSet are almost always typed.  So one corner case is `delete!(s,k)` where `s`  has keytype `K`, and `k` is not convertible to `K`.

oxinabox:
From: https://github.com/JuliaCollections/DataStructures.jl/pull/571#issuecomment-594146622
> Notice that there is an inconsistency with delete! that I didn't fix because I was afraid it would be a breaking change: #387

It is the general concensus that
turning errors into non-errors is not a breaking  change.

NHDaly:
> One subtlety: `Dict` and `Set` are not necessarily typed, whereas SortedDict and SortedSet are almost always typed. So one corner case is `delete!(s,k)` where `s` has keytype `K`, and `k` is not convertible to `K`.


I would agree that this "strongly typed" key functions is a feature of SortedDict/SortedSet, and so I think we can keep this behavior while still claiming to implement Base.AbstractDict / AbstractSet. It seems reasonable to me to allow `delete!` nonexistent keys _only when_ they're a subtype of `K`. 👍

NHDaly:
Okay, i've opened https://github.com/JuliaCollections/DataStructures.jl/pull/607, which does exactly this.

PR #490, opened 2019-04-04
Added `delete!(node_handle)` functionality for mutable binary heaps

### LINKED ISSUE #489: Is there any `delete!()` functionality for mutable binary heaps?
Shouldn't there be a `delete!(key_index)` function for mutable binary heaps ? We can implement this by swapping the node to be deleted with the last node, and then remove the last node by reducing heap_size and finally calling `heapify` to restore the binary heap property.

Sumegh-git:
Ping @oxinabox 
Fixes #489 .

### DIFF
--- a/docs/src/heaps.md
+++ b/docs/src/heaps.md
@@ -30,7 +30,7 @@ i = push!(h, v)              # adds a value to the heap and and returns a handle
 
 update!(h, i, v)             # updates the value of an element (referred to by the handle i)
 
-delete!(h, i)               # deletes the node present at index i from the heap
+delete!(h, i)               # deletes the node with handle i from the heap
 
 v, i = top_with_handle(h)    # returns the top value of a heap and its handle
 ```
--- a/docs/src/heaps.md
+++ b/docs/src/heaps.md
@@ -7,7 +7,7 @@ All heaps in this package are derived from `AbstractHeap`, and provide
 the following interface:
 
 ```julia
-# Let h be a heap, i be a handle, id be a node_id, and v be a value.
+# Let h be a heap, i be a handle, and v be a value.
 
 length(h)         # returns the number of elements
 
@@ -30,7 +30,7 @@ i = push!(h, v)              # adds a value to the heap and and returns a handle
 
 update!(h, i, v)             # updates the value of an element (referred to by the handle i)
 
-delete!(h, id)               # deletes the node present at node_index id from the heap
+delete!(h, i)               # deletes the node present at index i from the heap
 
 v, i = top_with_handle(h)    # returns the top value of a heap and its handle
 ```
--- a/docs/src/heaps.md
+++ b/docs/src/heaps.md
@@ -7,7 +7,7 @@ All heaps in this package are derived from `AbstractHeap`, and provide
 the following interface:
 
 ```julia
-# Let h be a heap, i be a handle, and v be a value.
+# Let h be a heap, i be a handle, id be a node_id, and v be a value.
 
 length(h)         # returns the number of elements
 
@@ -18,6 +18,9 @@ push!(h, v)       # add a value to the heap
 top(h)            # return the top value of a heap
 
 pop!(h)           # removes the top value, and returns it
+
+delete!(h,id)     # removes the node having node_id as id from heap 
+
 ```
 
 Mutable heaps (values can be changed after being pushed to a heap) are
--- a/docs/src/heaps.md
+++ b/docs/src/heaps.md
@@ -19,8 +19,6 @@ top(h)            # return the top value of a heap
 
 pop!(h)           # removes the top value, and returns it
 
-delete!(h,id)     # removes the node having node_id as id from heap 
-
 ```
 
 Mutable heaps (values can be changed after being pushed to a heap) are
@@ -31,6 +29,8 @@ provides the following interface:
 i = push!(h, v)              # adds a value to the heap and and returns a handle to v
 
 update!(h, i, v)             # updates the value of an element (referred to by the handle i)
+
+delete!(h, id)               # deletes the node present at node_index id from the heap
 
 v, i = top_with_handle(h)    # returns the top value of a heap and its handle
 ```
--- a/src/heaps.jl
+++ b/src/heaps.jl
@@ -37,8 +37,8 @@
 #   - update!(h, i, v)    updates the value of an element
 #                         (referred to by the handle i)
 #
-#   - delete!(h, id)      deletes the node present at
-#                         node_index id from the heap
+#   - delete!(h, i)      deletes the node present at
+#                         index i from the heap
 #
 #   - top_with_handle(h)  return the top value of a heap
 #                         and its handle
--- a/src/heaps.jl
+++ b/src/heaps.jl
@@ -37,6 +37,9 @@
 #   - update!(h, i, v)    updates the value of an element
 #                         (referred to by the handle i)
 #
+#   - delete!(h, id)      deletes the node present at
+#                         node_index id from the heap
+#
 #   - top_with_handle(h)  return the top value of a heap
 #                         and its handle
 #
--- a/src/heaps.jl
+++ b/src/heaps.jl
@@ -37,8 +37,8 @@
 #   - update!(h, i, v)    updates the value of an element
 #                         (referred to by the handle i)
 #
-#   - delete!(h, i)      deletes the node present at
-#                         index i from the heap
+#   - delete!(h, i)      deletes the node with
+#                         handle i from the heap
 #
 #   - top_with_handle(h)  return the top value of a heap
 #                         and its handle
--- a/src/heaps/mutable_binary_heap.jl
+++ b/src/heaps/mutable_binary_heap.jl
@@ -251,6 +251,36 @@ function update!(h::MutableBinaryHeap{T}, i::Int, v) where T
         _heap_bubble_down!(comp, nodes, nodemap, nd_id)
     end
 end
+"""
+    delete!{T}(h::MutableBinaryHeap{T}, i::Int)
+
+Delete the element present at index `i` from heap `h` .    
+"""
+function delete!(h::MutableBinaryHeap{T}, i::Int) where T
+    nodes = h.nodes
+    nodemap = h.node_map
+    comp = h.comparer
+
+    nd_id = nodemap[i]
+    rt = nodes[nd_id]
+    v0 = rt.value 
+    @inbounds nodemap[rt.handle] = 0
+
+    if length(nodes) == 1
+        #clear
+        empty!(nodes)
+    else
+        #place last node to removed one
+        @inbounds nodes[nd_id] = new_rt = pop!(nodes)
+        @inbounds nodemap[new_rt.handle] = 1
+
+        if length(nodes) > 1
+            _heap_bubble_down!(comp, nodes, nodemap, nd_id)
+        end
+    end
+    v0
+end
+        
 
 setindex!(h::MutableBinaryHeap, v, i::Int) = update!(h, i, v)
 getindex(h::MutableBinaryHeap, i::Int) = h.nodes[h.node_map[i]].value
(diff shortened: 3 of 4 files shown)


test/test_mutable_binheap.jl
```diff
@@ -157,6 +157,16 @@ end
         @test pop!(h) == 2
         @test isequal(heap_values(h), [7, 10])
         @test isequal(list_values(h), [10, 7])
+        empty!(h.nodes)
```
oxinabox: This should be its own separate testset.
And rather than `empty!(h.nodes)` it should just declare a new heap `h`.


Run-on tests become increasingly hard to read/edit.
And we have too many of them as it is.

src/heaps/mutable_binary_heap.jl
```diff
@@ -251,6 +251,36 @@ function update!(h::MutableBinaryHeap{T}, i::Int, v) where T
...
+        #place last node to removed one
+        @inbounds nodes[nd_id] = new_rt = pop!(nodes)
+        @inbounds nodemap[new_rt.handle] = 1
+
+        if length(nodes) > 1
+            _heap_bubble_down!(comp, nodes, nodemap, nd_id)
+        end
+    end
+    v0
+end
```
oxinabox: I believe, and you should confirm this,
The function that deleted an element at index `i`, and returns that element,
is called 
`pop!(collection, i)`.
for arrays and dicts.

`delete!(collection, i)` is the same but it always returns `nothing`.

Do check this though

oxinabox:
Tests are needed for
 - index = -5
 - index = 0
 - index larger  than the size of the heap.

src/heaps/mutable_binary_heap.jl
Sumegh-git: The `delete!(Dict, key)` and `deleteat!(Vector, i)` returns the remaining collection after deleting the element specified - https://github.com/JuliaLang/julia/blob/903644385b91ed8d95e5e3a5716c089dd1f1b08a/base/array.jl#L819-L834

Sumegh-git:
Should i throw `BoundsError` separately for such indexes by putting a check, or else anyways the function will throw since it will try to access `node_map[index]`.

oxinabox:
You should let the `node_map[index]` throw

StephenVavasis:
One issue regarding throwing an error for a bad key in `delete!`: #387

src/heaps/mutable_binary_heap.jl
Sumegh-git: So finally what should it be returning? The final container? or `nothing`? Any comments?

Sumegh-git:
> Thanks, I've only done a quick look over so far.
> But my notes after that first pass are here/above
> 
> Tests are needed for
> 
> * index = -5
> * index = 0
> * index larger  than the size of the heap.

Done!

src/heaps/mutable_binary_heap.jl
oxinabox: You should return the final container.
Sorry, i misremembered

oxinabox:
I suggest implement
`pop(heap, ii)` to return the value
then have something like

```
function delete!(heap, ii)
    pop!(heap, ii)
    return heap
```

Sumegh-git:
> I suggest implement
> `pop(heap, ii)` to return the value
> then have something like
> 
> ```
> function delete!(heap, ii)
>     pop!(heap, ii)
>     return heap
> ```

Pushed a commit with the changes.

Sumegh-git:
@oxinabox any more suggestions with the code structure or tests?

src/heaps/mutable_binary_heap.jl
```diff
@@ -251,6 +251,44 @@ function update!(h::MutableBinaryHeap{T}, i::Int, v) where T
         _heap_bubble_down!(comp, nodes, nodemap, nd_id)
     end
 end
+"""
+    pop!{T}(h::MutableBinaryHeap{T}, i::Int)
+
+Pops the element present at index `i` from heap `h` .    
+"""
+function pop!(h::MutableBinaryHeap{T}, i::Int) where T
```
kmsquire: This function copies most of its code from `_binary_heap_pop!`, which isn't very DRY.  

You should modify that function to accept `nd_id` as a parameter, defaulting to 1, and then call that function from here.

Sumegh-git: Yes, true!
Pushed a commit addressing your requested change.

Sumegh-git:
Ping @kmsquire . Is it okay now? Any other suggestions?

Sumegh-git:
Bump

Sumegh-git:
Bump @kmsquire .

src/heaps/mutable_binary_heap.jl
```diff
@@ -101,23 +101,23 @@ function _heap_bubble_down!(comp::Comp,
 end
 
 function _binary_heap_pop!(comp::Comp,
-    nodes::Vector{MutableBinaryHeapNode{T}}, nodemap::Vector{Int}) where {Comp,T}
+    nodes::Vector{MutableBinaryHeapNode{T}}, nodemap::Vector{Int}, nd_id::Int) where {Comp,T}
 
-    # extract root node
-    rt = nodes[1]
+    # extract  node
```
kmsquire: Nitpick: 2 spaces

Sumegh-git: ok. Will remove that .

Sumegh-git: Done!

test/test_mutable_binheap.jl
```diff
@@ -230,4 +249,4 @@ end
     end
     
 
-end # @testset MutableBinheap
+end # @testset MutableBinheap
```
kmsquire: This was probably your editor removing the last line of the file?  If so, can you set it not do do so, so we don't get extraneous changes like this?  Thanks.

docs/src/heaps.md
```diff
@@ -18,6 +18,9 @@ push!(h, v)       # add a value to the heap
 top(h)            # return the top value of a heap
 
 pop!(h)           # removes the top value, and returns it
+
+delete!(h,id)     # removes the node having node_id as id from heap 
+
```
kmsquire: This should be moved to the mutable heap description below, since it's not a method of all heaps.  It should also be added to comment at the top of `heaps.jl`.

src/heaps/mutable_binary_heap.jl
```diff
@@ -101,23 +101,23 @@ function _heap_bubble_down!(comp::Comp,
...
-        # place last node to root
-        @inbounds nodes[1] = new_rt = pop!(nodes)
-        @inbounds nodemap[new_rt.handle] = 1
+        # place last node to removed one
+        @inbounds nodes[nd_id] = new_rt = pop!(nodes)
+        @inbounds nodemap[new_rt.handle] = nd_id
 
         if length(nodes) > 1
-            _heap_bubble_down!(comp, nodes, nodemap, 1)
+            _heap_bubble_down!(comp, nodes, nodemap, nd_id)
```
kmsquire: Since we're not always replacing the root node any more, we'll need to call `_heap_bubble_up` here if the new node value is less than its parent (or alternatively, less than the node it replaces, since that's easier to determine).

kmsquire: Should also add a test for this situation.

Sumegh-git: Ya! Sure! Sorry missed it.

src/heaps/mutable_binary_heap.jl
```diff
@@ -101,23 +101,23 @@ function _heap_bubble_down!(comp::Comp,
 end
 
 function _binary_heap_pop!(comp::Comp,
-    nodes::Vector{MutableBinaryHeapNode{T}}, nodemap::Vector{Int}) where {Comp,T}
+    nodes::Vector{MutableBinaryHeapNode{T}}, nodemap::Vector{Int}, nd_id::Int) where {Comp,T}
```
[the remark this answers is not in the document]
Sumegh-git: I actually didn't get what you wanted in a separate PR. Could you elaborate a bit please?

Sumegh-git:
@kmsquire  made the changes you asked.

Sumegh-git:
Sorry for pinging so often. Reminding you once in case you forgot @kmsquire .

docs/src/heaps.md
```diff
@@ -7,7 +7,7 @@ All heaps in this package are derived from `AbstractHeap`, and provide
 the following interface:
 
 ```julia
-# Let h be a heap, i be a handle, and v be a value.
+# Let h be a heap, i be a handle, id be a node_id, and v be a value.
```
kmsquire: I believe that `i` and `id` actually refer to the same thing here and below.

Sumegh-git: No, actually `i` is the index of the element in heap, and `id` is the actual node index. 
It's like `node_map[i] == nd_id` is `true`.

kmsquire: Then is the documentation below incorrect?  It says

```julia
delete!(h, id)               # deletes the node present at node_index id from the heap
```

kmsquire: Looking at the code closer, the `handle` of a node is always set to its index within the heap.  Its purpose seems to be to have a convenient way for a node to know its index.

Sumegh-git: Actually, I had added that `delete!(h, id)` documentation, that's why it mentions `id`. 
The `update!(h, i , v)` has this line `nd_id = nodemap[i]`

Sumegh-git: I checked this also. After we `pop!` from the heap, that particular `nodemap[i]` is set to 0.

Sumegh-git: So @kmsquire this is fine then?

src/heaps.jl
```diff
@@ -37,6 +37,9 @@
 #   - update!(h, i, v)    updates the value of an element
 #                         (referred to by the handle i)
 #
+#   - delete!(h, id)      deletes the node present at
+#                         node_index id from the heap
+#
 #   - top_with_handle(h)  return the top value of a heap
 #                         and its handle
 #
```
kmsquire: For consistency with the other mutable heap functions, I think we need to change `delete!` to work on a `handle` instead of a `node_index`.

Sumegh-git: In a minute!

Sumegh-git: Done @kmsquire .

src/heaps.jl
```diff
@@ -37,8 +37,8 @@
 #   - update!(h, i, v)    updates the value of an element
 #                         (referred to by the handle i)
 #
-#   - delete!(h, id)      deletes the node present at
-#                         node_index id from the heap
+#   - delete!(h, i)      deletes the node present at
+#                         index i from the heap
```
kmsquire: Since the other functions in this group (`push!`, `update!`, and `top_with_handle`) refer to `i` as a `handle`, let's do it here as well.

docs/src/heaps.md
```diff
@@ -30,7 +30,7 @@ i = push!(h, v)              # adds a value to the heap and and returns a handle
 
 update!(h, i, v)             # updates the value of an element (referred to by the handle i)
 
-delete!(h, id)               # deletes the node present at node_index id from the heap
+delete!(h, i)               # deletes the node present at index i from the heap
```
kmsquire: Here too.

src/heaps/mutable_binary_heap.jl
```diff
@@ -259,10 +259,11 @@ end
 """
     delete!{T}(h::MutableBinaryHeap{T}, i::Int)
 
-Deletes the element present at node_id `i` from heap `h` . 
+Deletes the element present at index `i` from heap `h` .
```
kmsquire:
```suggestion
Deletes the element with handle `i` from heap `h` .
```

docs/src/heaps.md
Sumegh-git: Done!

Sumegh-git:
Anything more @kmsquire ?

kmsquire:
Merged-thanks @Sumegh-git!

PR #494, opened 2019-04-10
Adding the get! method for the PriorityQueue very similar to Dict.

sambitdash:


test/test_priority_queue.jl
```diff
@@ -110,12 +110,14 @@ import Base.Order.Reverse
...
 
-        @testset "peek/get/dequeue!" begin
+        @testset "peek/get/dequeue!/get!" begin
             @test peek(pq1) == ('a'=>1)
             @test get(pq1, 'a', 0) == 1
             @test get(pq1, 'c', 0) == 0
             @test dequeue!(pq1) == 'a'
             @test dequeue!(pq1) == 'b'
+            @test get!(pq1, 'c', 0) == 0
+            @test peek(pq1) == ('c'=>0)
```
oxinabox: This also needs a test that when the iterm is present, `get!` returns the correct result

src/priorityqueue.jl
```diff
@@ -175,6 +175,17 @@ function get(pq::PriorityQueue{K,V}, key, deflt) where {K,V}
     i == 0 ? deflt : pq.xs[i].second
 end
 
+function get!(pq::PriorityQueue{K,V}, key, deflt) where {K,V}
```
oxinabox: please spell out `deflt` fully,
it isn't much longer, and it makes it more readable.
(I spent a few seconds thinking `deflate`, wat?)

I am guessing you just copies the one above, which is fair.
It would be really great if you could fix that one on the way past.
(Some of this code is older than the idea of coding standard in julia)
Also nio need to mention `K` or `V`.

sambitdash: @oxinabox Agreed! I just did a cut and paste to match the old code. I will fix both places.

oxinabox:
LGTM, thanks!

PR #497, opened 2019-04-16
Add suffix form of Fenwick Tree

eulerkochy:
This is a form of Fenwick Tree(#485 ) which supports dynamic update in size. This PR is a result of [this](https://github.com/JuliaCollections/DataStructures.jl/pull/485#issuecomment-480479329) conversation with @kmsquire . 

- [x] Added tests
- [x] Added documentation

### DIFF
--- a/src/suffix_fenwick.jl
+++ b/src/suffix_fenwick.jl
@@ -6,7 +6,7 @@ end
 """
     SuffixFenwickTree{T}(n)
     
-Constructs a [`SuffixFenwickTree`](https://en.wikipedia.org/wiki/SuffixFenwick_tree) of length `n`.
+Constructs a [`SuffixFenwickTree`](https://stackoverflow.com/questions/21995930/dynamic-i-e-variable-size-fenwick-tree) of length `n`.
  
 """
 SuffixFenwickTree{T}(n::Integer) where T = SuffixFenwickTree{T}(zeros(T, n), n)
--- a/src/suffix_fenwick.jl
+++ b/src/suffix_fenwick.jl
@@ -1,0 +1,95 @@
+mutable struct SuffixFenwickTree{T}
+    bi_tree::Vector{T} #bi_tree is shorthand for Binary Indexed Tree, an alternative name for SuffixFenwick Tree
+    n::Int
+end
+
+"""
+    SuffixFenwickTree{T}(n)
+    
+Constructs a [`SuffixFenwickTree`](https://en.wikipedia.org/wiki/SuffixFenwick_tree) of length `n`.
+ 
+"""
+SuffixFenwickTree{T}(n::Integer) where T = SuffixFenwickTree{T}(zeros(T, n), n)
+
+"""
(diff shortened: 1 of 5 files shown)


eulerkochy:
@kmsquire review?

eulerkochy:
Bump!

src/suffix_fenwick.jl
```diff
@@ -0,0 +1,95 @@
+mutable struct SuffixFenwickTree{T}
+    bi_tree::Vector{T} #bi_tree is shorthand for Binary Indexed Tree, an alternative name for SuffixFenwick Tree
+    n::Int
```
kmsquire: Because the structure of the suffix version of the tree is identical to `FenwickTree`, and some of the code below is identical, I would consider modifying `FenwickTree` itself to be parameterized on whether it is a prefix or suffix tree.

Something like (untested):
```julia
mutable struct GenericFenwickTree{F,T}
    bi_tree::Vector{T}
    n::Int

GenericFenwickTree{F,T}(n::Integer) where {F,T} = GenericFenwickTree{F,T}(zeros(T, n), n)
const FenwickTree{T} = GenericFenwickTree{:prefix, T}
const SuffixFenwickTree{T} = GenericFenwickTree{:suffix, T}

# etc.
```

Then shared code can dispatch on `GenericFenwickTree`, and code specific to prefix or suffix trees can dispatch on `FenwickTree` and `SuffixFenwickTree`, respectively.
You could also override `show` to display `FenwickTree` or `SuffixFenwickTree`, instead of `GenericFenwickTree{:prefix}` and `GenericFenwickTree{:suffix}`, etc.

What do you think?

kmsquire: I'll also add that I think you can remove `n` from the struct, since it's always the length of `bi_tree`, and getting the length of a vector is `O(1)` (the length is stored).

eulerkochy: > What do you think?

It would be great! So, instead of creating a new file, like as I did here, I suggest that I make changes in `fenwick.jl` and add the generic version of the tree there. Opinion?

eulerkochy: > 

> I'll also add that I think you can remove `n` from the struct, since it's always the length of `bi_tree`, and getting the length of a vector is `O(1)` (the length is stored).

Great idea, making the change in `FenwickTree` too, would be appropiate.

kmsquire: Just in case you missed the 👍 , yes, this is the idea--thanks!

eulerkochy: @kmsquire i was working on this PR today, and this is what I think I think about it. Although the generic structure of `FenwickTree` and `SuffixFenwickTree` is the same, but their purposes and quite different. Why I am able to justify this ? The documentation for the functions aren't similar. So, there will be lot of code duplication and confusion if both of the code end up in a same file, `fenwick.jl`.
As for the other change, the PR has been made, and you have already approved it !!

eulerkochy:
can be closed in favour of #589

PR #492, opened 2019-04-08
Add documentation for FenwickTree

eulerkochy:
Fix #491 and a positional change in docstring in `fenwick.jl`

PR #501, opened 2019-05-14
RobinDict 

eulerkochy:
I call this `RobinDict`, which implements [Robinhood Hashing technique](https://cs.uwaterloo.ca/research/tr/1986/CS-86-14.pdf) .

- [x] Write `delete`/`erase` operation for `RobinDict`
- [x] Add tests
- [x] Write docstrings
- [x] Benchmarking the code
- [x] Documentation


oxinabox:
Some early points since i know this is not done yet.
 - Please use 4 spaces for intending. Not tabs.
 - IIRC: The AbstractDict interface requires `get(dict, key, default)`  to be implemented, and if you do that (along with some other things you've already done) a bunch of things (like equality) will work for free.

eulerkochy:
I didn't quite get the first point, should `Tab size : 4` do the same thing?
PS: Look at the bottom right-hand corner of the attached screenshot
Screenshot (5)_LI

src/robin_dict.jl
```diff
@@ -0,0 +1,189 @@
...
+function rh_search(h::RobinDict{K, V}, key::K) where {K, V}
+	sz = length(h.keys)
+	index = hashindex(key, sz)
+	cdibs = 0
+	while true
+		if h.slots[index] == 0x0
+			return -1
+		elseif cdibs > h.dibs[index]
+			return -1
+		elseif h.keys[index] == key
```
oxinabox: I'm not sure how to configure you editor,
but I can tell you there are definately tabs on this line.

eulerkochy:
My bad. That block of code was written in Jupyter Notebook. Thanks for
pointing out. I'll take care of it from now on.

src/robin_dict.jl
```diff
@@ -1,6 +1,6 @@
 import Base: setindex!, sizehint!, empty!, isempty, length, getindex, getkey, haskey, iterate, @propagate_inbounds, pop!, delete!, get, isbitstype
 
-const LOAD_FACTOR = 0.90
+const LOAD_FACTOR = 0.80
```
oxinabox: Can this have a comment,
also should be `ROBIN_DICT_LOAD_FACTOR`

eulerkochy: Sure! That's the load factor I'm trying to achieve, but I can't seem to get close to that as of now!

src/robin_dict.jl
```diff
@@ -0,0 +1,679 @@
...
+RobinDict(ps::Pair{K,V}...) where {K,V} = RobinDict{K,V}(ps)
+RobinDict(ps::Pair{K}...,) where {K}             = RobinDict{K,Any}(ps)
+RobinDict(ps::(Pair{K,V} where K)...,) where {V} = RobinDict{Any,V}(ps)
+RobinDict(ps::Pair...) = RobinDict{Any,Any}(ps)
+
+function RobinDict(kv)
+    try
+        dict_with_eltype(kv, eltype(kv))
+    catch e
+        if isempty(methods(iterate, (typeof(kv),))) ||
```
vtjnash:
```suggestion
    if !isiterable(typeof(kv)) || !all(x -> isa(x, Union{Tuple,Pair}), kv)
```
(from Base)

src/robin_dict.jl
```diff
@@ -0,0 +1,679 @@
...
+            throw(ArgumentError("RobinDict(kv): kv needs to be an iterator of tuples or pairs"))
+        else
+            rethrow(e)
+        end
+    end
+end
+
+dict_with_eltype(kv, ::Type{Tuple{K,V}}) where {K,V} = RobinDict{K,V}(kv)
+dict_with_eltype(kv, ::Type{Pair{K,V}}) where {K,V} = RobinDict{K,V}(kv)
+dict_with_eltype(kv, t) = RobinDict{Any,Any}(kv)
```
vtjnash: These seem to be from an old implementation of `dict_with_eltype` and should be dropped

src/robin_dict.jl
```diff
@@ -0,0 +1,679 @@
...
+RobinDict(kv::AbstractDict{K,V}) where {K,V} = RobinDict{K,V}(kv)
+
+RobinDict(ps::Pair{K,V}...) where {K,V} = RobinDict{K,V}(ps)
+RobinDict(ps::Pair{K}...,) where {K}             = RobinDict{K,Any}(ps)
+RobinDict(ps::(Pair{K,V} where K)...,) where {V} = RobinDict{Any,V}(ps)
+RobinDict(ps::Pair...) = RobinDict{Any,Any}(ps)
+
+function RobinDict(kv)
+    try
+        dict_with_eltype(kv, eltype(kv))
```
vtjnash:
```suggestion
        dict_with_eltype((K, V) -> RobinDict{K, V}, kv, eltype(kv))
```

src/robin_dict.jl
```diff
@@ -0,0 +1,679 @@
...
+RobinDict(kv::Tuple{Vararg{Pair}})                        = RobinDict{Any,Any}(kv)
+
+RobinDict(kv::AbstractArray{Tuple{K,V}}) where {K,V} = RobinDict{K,V}(kv)
+RobinDict(kv::AbstractArray{Pair{K,V}}) where {K,V}  = RobinDict{K,V}(kv)
+RobinDict(kv::AbstractDict{K,V}) where {K,V} = RobinDict{K,V}(kv)
+
+RobinDict(ps::Pair{K,V}...) where {K,V} = RobinDict{K,V}(ps)
+RobinDict(ps::Pair{K}...,) where {K}             = RobinDict{K,Any}(ps)
+RobinDict(ps::(Pair{K,V} where K)...,) where {V} = RobinDict{Any,V}(ps)
+RobinDict(ps::Pair...) = RobinDict{Any,Any}(ps)
```
vtjnash: This seems to be a much larger list of constructors than Base has:
```
julia> methods(Dict)
# 5 methods for type constructor:
[1] (::Type{Dict})() in Base at dict.jl:117
[2] (::Type{Dict})(kv::Tuple{}) in Base at dict.jl:118
[3] (::Type{Dict})(ps::Pair{K,V}...) where {K, V} in Base at dict.jl:123
[4] (::Type{Dict})(ps::Pair...) in Base at dict.jl:124
[5] (::Type{Dict})(kv) in Base at dict.jl:127
```

src/robin_dict.jl
```diff
@@ -0,0 +1,679 @@
...
+end
+
+dict_with_eltype(kv, ::Type{Tuple{K,V}}) where {K,V} = RobinDict{K,V}(kv)
+dict_with_eltype(kv, ::Type{Pair{K,V}}) where {K,V} = RobinDict{K,V}(kv)
+dict_with_eltype(kv, t) = RobinDict{Any,Any}(kv)
+
+# default hashing scheme used by Julia
+hashindex(key, sz) = (((hash(key)%Int) & (sz-1)) + 1)::Int
+
+_tablesz(x::Integer) = x < 16 ? 16 : one(x)<<((sizeof(x)<<3)-leading_zeros(x-1))
```
vtjnash: Seems like these are just copies of Base, so you can do `using Base: _tablesz, hashindex` instead

src/robin_dict.jl
```diff
@@ -0,0 +1,682 @@
...
+    h.idxfloor = 0
+
+    for i = 1:sz
+        @inbounds if olds[i] == 0x1
+            k = oldk[i]
+            v = oldv[i]
+            rh_insert_for_rehash!(h, k, v)
+            if h.totalcost != totalcost0
+                # if `h` is changed by a finalizer, retry
+                return rehash!(h, newsz)
```
vtjnash: you've already destroyed `h` by assigning the fields at the top of the function, you need to assign those at the end

src/robin_dict.jl
```diff
@@ -0,0 +1,682 @@
...
+```
+"""
+mutable struct RobinDict{K,V} <: AbstractDict{K,V}
+    #there is no need to maintain an table_size as an additional variable
+    slots::Array{UInt8,1} # indicator, to be used later on
+    keys::Array{K,1}
+    vals::Array{V,1}
+    dibs::Array{Int,1} # distance to initial bucket - critical for implementation
+    count::Int
+    totalcost::Int
```
vtjnash: what is a `totalcost`?

src/robin_dict.jl
```diff
@@ -0,0 +1,682 @@
...
+    maxprobe::Int    # length of longest probe
+    idxfloor::Int
+    max_lf :: Float32
```
vtjnash: Seems to be unused (except for perhaps statistics)?
```suggestion
```

src/robin_dict.jl
```diff
@@ -0,0 +1,682 @@
...
+# default hashing scheme used by Julia
+hashindex(key, sz) = (((hash(key)%Int) & (sz-1)) + 1)::Int
+
+_tablesz(x::Integer) = x < 16 ? 16 : one(x)<<((sizeof(x)<<3)-leading_zeros(x-1))
+
+# insert algorithm
+function rh_insert!(h::RobinDict{K, V}, key::K, val::V) where {K, V}
+    # table full
+    if h.count == length(h.keys)
+        return -1
```
vtjnash:
```suggestion
        rehash!(h, h.count > 64000 ? sz*2 : sz*4)
```

But since you handle this in the caller, an `@assert` seems more applicable here to me

src/robin_dict.jl
eulerkochy: Kind of like `age` parameter in `Dict`

src/robin_dict.jl
```diff
@@ -0,0 +1,679 @@
...
+        h.slots[curr] = 0x0
+        ccall(:jl_arrayunset, Cvoid, (Any, UInt), h.keys, curr-1)
+        ccall(:jl_arrayunset, Cvoid, (Any, UInt), h.vals, curr-1)
+        h.dibs[curr] = 0
+    end
+
+    h.count -= 1
+    h.totalcost += 1
+    # this is necessary because key at idxfloor might get deleted 
+    h.idxfloor = get_idxfloor(h)
```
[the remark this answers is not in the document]
eulerkochy: Thinking intuitively, it should be equivalent! Let's commit it and see!

src/robin_dict.jl
```diff
@@ -0,0 +1,679 @@
...
+    if !isequal(key, key0)
+        throw(ArgumentError("$(limitrepr(key0)) is not a valid key for type $K"))
+    end
+    index = rh_search(h, key)
+    @inbounds return (index < 0) ? default : h.keys[index]::K
+end
+
+# backward shift deletion by not keeping any tombstones
+function rh_delete!(h::RobinDict{K, V}, index) where {K, V}
+    #forceful
```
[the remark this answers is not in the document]
eulerkochy: `rh_delete!` calls happen only when there is a `key/value` pair in the table. This check is sort-of an overkill for checking that `index` should really exist.

src/robin_dict.jl
eulerkochy: Okay!

src/robin_dict.jl
eulerkochy: That's the main purpose of it! Should I remove this now?

src/robin_dict.jl
eulerkochy: Alright!

src/robin_dict.jl
```diff
@@ -0,0 +1,679 @@
...
+get!(dict, key) do
+    # default value calculated here
+    time()
+end
+```
+"""
+get!(f::Function, collection, key)
+
+function get!(default::Callable, h::RobinDict{K,V}, key0) where {K, V}
+    isa(convert(K, key0), MethodError) && throw(MethodError("Cannot `convert` an object of type $(typeof(key0)) to an object of type $K"))
```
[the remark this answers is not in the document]
eulerkochy: Okay, wait! I am already checking for that!

src/robin_dict.jl
vtjnash: Yeah, I think that would make more sense to exist in a wrapper as part of an analysis harness

src/robin_dict.jl
vtjnash: It just seems unfortunate that this is causing there to be two copies of the insert function

src/robin_dict.jl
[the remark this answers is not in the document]
vtjnash: yeah, that's not what this code would do

src/robin_dict.jl
eulerkochy: I can cut that down to one function. That was a pretty lazy implementation on my part.

src/robin_dict.jl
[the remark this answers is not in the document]
eulerkochy: removed!

eulerkochy:
Ready for review @oxinabox

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+    
+    ckey, cval, chash = key, val, hash_key(key)
+    sz = length(h.keys)
+    index_init = desired_index(chash, sz)
+
+    index_curr = index_init
+    probe_distance = 0
+    probe_current = 0
+    @inbounds while true
+        if (isslotempty(h, index_curr)) || (isslotfilled(h, index_curr) && h.keys[index_curr] == ckey)
```
vtjnash:
```suggestion
        if (isslotempty(h, index_curr)) || (isslotfilled(h, index_curr) && isequal(h.keys[index_curr], ckey))
```

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+            h.vals[index_curr], cval = cval, h.vals[index_curr]
+            h.keys[index_curr], ckey = ckey, h.keys[index_curr]
+            h.hashes[index_curr], chash = chash, h.hashes[index_curr]
+            probe_current = probe_distance
+        end
+        probe_current += 1
+        index_curr = (index_curr & (sz - 1)) + 1
+    end
+    
+    @inbounds if isslotfilled(h, index_curr) && h.keys[index_curr] == ckey
```
vtjnash:
```suggestion
    @inbounds if isslotfilled(h, index_curr) && isequal(h.keys[index_curr], ckey)
```

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+    sz = length(h.keys)
+    chash = hash_key(key)
+    index = desired_index(chash, sz)
+    cdibs = 0
+    @inbounds while true
+        if isslotempty(h, index)
+            return -1
+        elseif cdibs > calculate_distance(h, index)
+            return -1
+        elseif h.keys[index] == key && h.hashes[index] == chash
```
vtjnash:
```suggestion
        elseif h.hashes[index] == chash && (h.keys[index] === key || isequal(h.keys[index], key))
```

I assume you meant to do the hash-check first as a fast path, then the slower equality comparison?

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+function RobinDict(kv)
+    try
+        d = dict_with_eltype((K, V) -> RobinDict{K, V}, kv, eltype(kv))
+        if isa(d, Dict)
+            return RobinDict(d)
+        else
+            return d
+        end
+    catch e
+    if !isiterable(typeof(kv)) || !all(x -> isa(x, Union{Tuple,Pair}), kv)
```
oxinabox:
```suggestion
        if !isiterable(typeof(kv)) || !all(x -> isa(x, Union{Tuple,Pair}), kv)
```

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+```
+"""
+get!(f::Function, collection, key)
+
+function get!(default::Callable, h::RobinDict{K,V}, key0::K) where {K, V}
+    key = convert(K, key0)
+    return _get!(default, h, key)
+end
+
+function _get!(default::Callable, h::RobinDict{K,V}, key::K) where V where K
```
oxinabox:
```suggestion
function _get!(default, h::RobinDict{K,V}, key::K) where V where K
```

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+```julia
+get!(dict, key) do
+    # default value calculated here
+    time()
+end
+```
+"""
+get!(f::Function, collection, key)
+
+function get!(default::Callable, h::RobinDict{K,V}, key0::K) where {K, V}
```
oxinabox: No need to ban functors

```suggestion
function get!(default, h::RobinDict{K,V}, key0::K) where {K, V}
```

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+```julia
+get(dict, key) do
+    # default value calculated here
+    time()
+end
+```
+"""
+get(::Function, collection, key)
+
+function get(default::Callable, h::RobinDict{K,V}, key0) where {K, V}
```
oxinabox:
```suggestion
function get(default, h::RobinDict{K,V}, key0) where {K, V}
```

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+        h.vals[curr] = h.vals[next]
+        h.keys[curr] = h.keys[next]
+        h.hashes[curr] = h.hashes[next]
+        curr = next
+        next = (next & (sz-1)) + 1
+    end
+    
+    #curr is at the last position, reset back to normal
+    ccall(:jl_arrayunset, Cvoid, (Any, UInt), h.keys, curr-1)
+    ccall(:jl_arrayunset, Cvoid, (Any, UInt), h.vals, curr-1)
```
oxinabox: What is thsi about?
Why do we need to unset elements in an array?
Are we later checking them with `isdefined`?

eulerkochy: This way the way deletions were happening in `base/dict.jl`, so I kept this part of the code. No, aren't checking for `isdefined`, checking whether `hashes` is `0` or not, seems sufficient

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+```jldoctest
+julia> RobinDict("A"=>1, "B"=>2)
+RobinDict{String,Int64} with 2 entries:
+  "B" => 2
+  "A" => 1
+```
+"""
+mutable struct RobinDict{K,V} <: AbstractDict{K,V}
+    #there is no need to maintain an table_size as an additional variable
+    # dibs::Array{Int8,1} # distance to initial bucket - critical for implementation
```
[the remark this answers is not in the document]
eulerkochy: ohh, i'll scrap that out! that was my initial implementation. somehow forgot to delete the commented out line!

src/robin_dict.jl
oxinabox: @vtjnash maybe can comment.
If there is a good reason to do this we should explain why,
if not we should just not do this.

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+end
+
+function RobinDict(kv)
+    try
+        d = dict_with_eltype((K, V) -> RobinDict{K, V}, kv, eltype(kv))
+        if isa(d, Dict)
+            return RobinDict(d)
+        else
+            return d
+        end
```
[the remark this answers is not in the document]
eulerkochy: there were few tests for L93 that were returning `Dict` instead of `RobinDict`, so the whole code segment checks for the return type of `dict_with_eltype` and acts accordingly

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+    if !isiterable(typeof(kv)) || !all(x -> isa(x, Union{Tuple,Pair}), kv)
+            !all(x->isa(x,Union{Tuple,Pair}),kv)
+            throw(ArgumentError("RobinDict(kv): kv needs to be an iterator of tuples or pairs"))
+        else
+            rethrow(e)
+        end
+    end
+end
+
+hash_key(key) = ((hash(key)%UInt32) & 0xffffffff)|1
```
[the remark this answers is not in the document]
eulerkochy: `hash(key)` behaves according to architecture, it returns `UInt64` for 64 bit machines, `UInt32` otherwise. I am storing 32 bits hash, so I guess `(hash(key)%UInt32)|1` will be correct

kmsquire:
@chethega, would you be interested in reviewing this?

src/robin_dict.jl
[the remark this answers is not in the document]
oxinabox: IIRC
The literal `1` depends on architecture also.
That function thus always returns that sized `Int`.

I recommend adding a test for this.

src/robin_dict.jl
chethega: We need to unset the array indices for key/value types that are stored by garbage collected reference. Otherwise, the garbage collector will see the reference and keep the object alive. 

The unset-call is unneeded for bitstypes and bits unions; it should look exactly like Base here:
```
isbitstype(K) || isbitsunion(K) || ccall(:jl_arrayunset, Cvoid, (Any, UInt), h.keys, curr-1)
isbitstype(V) || isbitsunion(V) || ccall(:jl_arrayunset, Cvoid, (Any, UInt), h.vals, curr-1)
```

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+    h.totalcost = 0
+    h.idxfloor = 0
+    return h
+end
+ 
+function rh_search(h::RobinDict{K, V}, key::K) where {K, V}
+    sz = length(h.keys)
+    chash = hash_key(key)
+    index = desired_index(chash, sz)
+    cdibs = 0
```
chethega: This would be the place for prefetching `h.keys[index]` and `h.vals[index]`, via
`prefetch_r(pointer(h.keys, index))`, using e.g.:
```
@inline function _prefetchr(p::Ptr)
    p = reinterpret(UInt, p)
    if UInt === UInt64
        Core.Intrinsics.llvmcall(("declare void @llvm.prefetch(i8*, i32, i32, i32)", """
    %ptr = inttoptr i64 %0 to i8*
    call void @llvm.prefetch(i8* %ptr, i32 0, i32 3, i32 1)
    ret void
    """), Nothing, Tuple{UInt}, p)
    else
        Core.Intrinsics.llvmcall(("declare void @llvm.prefetch(i8*, i32, i32, i32)", """
    %ptr = inttoptr i32 %0 to i8*
    call void @llvm.prefetch(i8* %ptr, i32 0, i32 3, i32 1)
    ret void
    """), Nothing, Tuple{UInt}, p)
    end
end
```
For a dict with low load factor, this is unneeded: The branch predictor will speculatively execute the loop and prefetch the relevant addresses, spectre-style, while waiting for the `h.hashes[index]` cache miss to resolve.

I am unsure about high loads: When the branch predictor guesses that we won't have a match, then the transient execution might run in a look that fails to fire off the memory load.

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+    totalcost0 = h.totalcost
+    h.count = 0
+    h.maxprobe = 0
+    h.idxfloor = 0
+
+    for i = 1:sz
+        @inbounds if oldh[i] != 0
+            k = oldk[i]
+            v = oldv[i]
+            rh_insert!(h, k, v)
```
chethega: `rehash!` should not compute a single hash; instead we should reuse the stored hashes.

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+            !all(x->isa(x,Union{Tuple,Pair}),kv)
+            throw(ArgumentError("RobinDict(kv): kv needs to be an iterator of tuples or pairs"))
+        else
+            rethrow(e)
+        end
+    end
+end
+
+hash_key(key) = ((hash(key)%UInt32) & 0xffffffff)|1
+desired_index(hash, sz) = ((hash) & (sz -1)) + 1
```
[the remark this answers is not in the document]
chethega: The LSB of `hash_key` is always set and is always used in the `desired_index`. In other words, it looks like every entry wants to be in an even slot.

Is this intended?

Otherwise, one could try something like storing some `ashift` in the RobinDict such that `sz = 1<<(32 - ashift)`, and `desired_index(hash, ashift) = 1+(hash >>ashift)`. 

This would have the advantage that when we iterate over one RobinDict and use that to access another RobinDict, we get very favorable memory access patterns (almost linear), regardless of whether sizes match. This especially applies for `rehash!`. With the "reduction-by-masking", we get an interleaving of `new_size/old_size` many linear patterns.

If we stay with masking, one could store `hash_key(key) = (hash(key)%UInt32) | 0x80000000`

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+    # table full
+    @assert h.count != length(h.keys)
+    
+    ckey, cval, chash = key, val, hash_key(key)
+    sz = length(h.keys)
+    index_init = desired_index(chash, sz)
+
+    index_curr = index_init
+    probe_distance = 0
+    probe_current = 0
```
chethega: As above, here would be the moment to prefetch using e.g.
```
@inline function _prefetchw(p::Ptr)
    p = reinterpret(UInt, p)
    if UInt === UInt64
        Core.Intrinsics.llvmcall(("declare void @llvm.prefetch(i8*, i32, i32, i32)", """
    %ptr = inttoptr i64 %0 to i8*
    call void @llvm.prefetch(i8* %ptr, i32 1, i32 3, i32 1)
    ret void
    """), Nothing, Tuple{UInt}, p)
    else
        Core.Intrinsics.llvmcall(("declare void @llvm.prefetch(i8*, i32, i32, i32)", """
    %ptr = inttoptr i32 %0 to i8*
    call void @llvm.prefetch(i8* %ptr, i32 1, i32 3, i32 1)
    ret void
    """), Nothing, Tuple{UInt}, p)
    end
end
```

chethega:
After taking a quick look, I really like this.

I think the max load factor of 0.7, combined with power-of-2 sizes is problematic: It wastes a lot of space. Since rehashing can be very cheap (no hash calculations), it could be advantageous to use `pos_of_hash(h::UInt32, sz) = widemul(h, sz%UInt32)>>32`, cf eg [here](https://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/).

That would create the issue that computation of displacement/distance is not as cheap anymore. 

I further see that using 32 bit slots/hashes make fast simd iteration hard. The 8 bit slots in base allow for a mostly branch-free iteration, cf [here](https://github.com/JuliaLang/julia/pull/31243). 

There is a possibility of also storing a bitvector-like slots, just for iteration. This costs us one additional cache line fill on inserts and deletions. The latency of this would be unimportant (nothing depends on that). Thoughts?

At last, I'd like to see some `check_invariants` function for the tests (verify and thereby document all maintained invariants of the RobinDict).

src/robin_dict.jl
eulerkochy: So, that would require an alternate `insert_for_rehash`, right?

eulerkochy:
@chethega the computation of `distance` has to be cheap, for this hashing scheme to perform well. And, regarding the implementation using BitVector, I had a discussion with @vtjnash , before zeroing in down onto this implementation. The reason for 70% load factor is summarised here https://nextjournal.com/eulerkochy/blog-two-jsoc/ . 

> At last, I'd like to see some check_invariants function for the tests 

Can you elaborate? Or better, provide an example!

chethega:
Re distance computation: `widemul(h, N)>>32 % UInt32` should not be that much more expensive than what is done currently (still < 5 cycle latency). We'd have to benchmark that.

Regarding `check_invariants`:  The essential invariant is that if an element has displacement `k`, then every item between the desired and actual position must be filled and must have displacement `>= k`; otherwise, we might report a present key as non-present. I would feel better if we had code and tests that check this invariant directly. While doing that, one can also check that hashes are correct (to directly defend against bugs where `h.hashes` and `h.keys` go out-of-sync during all the swaps).

Regarding load factor, 70% is certainly a good target. With power-of-2 sizes, this often means 35% load factor or even 17% in your blogpost, which is pretty bad with respect to memory consumption. If we reused the stored hashes during `rehash!` (i.e. not compute any hashes), and also improved the cache locality for `rehash!` (use shift/widemul instead of masking to turn hash into desired index), then we might get away with smaller and cheaper resizes, and could also "freeze" dictionaries (rehash! them to optimal load factor, useful when batch-constructing a large dictionary that will later need very few insertion/deletions).

The `BitVector` would only be used for iteration, not for lookups. That gave a 10x speedup for moderate load factors [here](https://github.com/JuliaLang/julia/pull/29746) and 2x-4x in Base dict [here](https://github.com/JuliaLang/julia/pull/31243). Whether that is worth it depends on how much we care for iteration speed, as opposed to lookup and insert/delete.

Btw, I really like how you unapologetically restrict sizes to `1<<31` items.

src/robin_dict.jl
chethega: Exactly. `Base` also has an alternate `insert_for_rehash` that never calls `isequal` on keys; here, we could additionally save the `hash` calls.

src/robin_dict.jl
vtjnash: This seems likely far too late for prefetch to make a difference. You would have needed to call it around a hundred cycles ago for it to be expected to reduce latency. Putting it here, mere cycles away from the expected load, seems likely to just slow down execution (it'll waste one of the a load ports and an instruction micro-op). https://stackoverflow.com/a/3123012

We might be able to move the actual load here, but currently the intent is to follow the implementation in base reasonably closely, for more fair benchmarking.

For this PR, I'd like to see it get merged fairly quickly (after a thorough review for correctness). Then we can work on additional performance evaluation and other trade-offs, relative to the initial version as a reference point.

src/robin_dict.jl
vtjnash: Perhaps, but that's the current API for Dict (the proposed change would introduce a method ambiguity)

src/robin_dict.jl
vtjnash: Correct. The checks are non-essential (jl_arrayunset must execute them also), but generally it's good to match Base anyways.

src/robin_dict.jl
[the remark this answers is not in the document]
vtjnash: Are you able to reproduce those? It sounds like a bug in `dict_with_eltype`, unless you perhaps just had old state confusing the test?

src/robin_dict.jl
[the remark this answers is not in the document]
vtjnash: From discussions with eulerkochy, I think our goal was to drop the high bit (ensuring the hash would never be exactly zero—our sentinel value for an empty cell, and easy to mask out later)
```suggestion
hash_key(key) = (hash(key) % UInt32) | 0x80000000
```
or we could just filter that value directly:
```suggestion
hash_key(key) = (h = hash(key); h == 0 ? UInt32(1) : h % UInt32)
```

src/robin_dict.jl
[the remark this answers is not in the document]
eulerkochy: > Is this intended?

No, it wasn't. Just did it for some reason

src/robin_dict.jl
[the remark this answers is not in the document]
eulerkochy: Test at L132, 140 and 148 takes a hit, if you just return what `dict_with_eltype` returns.
```julia
typeof: Test Failed at C:\Users\ww\Desktop\Julia\DataStructures.jl\test\test_robin_dict.jl:132
  Expression: typeof(d) == typeof(d2) == typeof(d3) == RobinDict{Int, Any}
   Evaluated: RobinDict{Int64,Any} == Dict{Int64,Any} == Dict{Int64,Any} == RobinDict{Int64,Any}
Stacktrace:
 [1] top-level scope at C:\Users\ww\Desktop\Julia\DataStructures.jl\test\test_robin_dict.jl:132
 [2] top-level scope at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.1\Test\src\Test.jl:1083
 [3] top-level scope at C:\Users\ww\Desktop\Julia\DataStructures.jl\test\test_robin_dict.jl:118
typeof: Test Failed at C:\Users\ww\Desktop\Julia\DataStructures.jl\test\test_robin_dict.jl:140
  Expression: typeof(d) == typeof(d2) == typeof(d3) == RobinDict{Any, Int}
   Evaluated: RobinDict{Any,Int64} == Dict{Any,Int64} == Dict{Any,Int64} == RobinDict{Any,Int64}
Stacktrace:
 [1] top-level scope at C:\Users\ww\Desktop\Julia\DataStructures.jl\test\test_robin_dict.jl:140
 [2] top-level scope at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.1\Test\src\Test.jl:1083
 [3] top-level scope at C:\Users\ww\Desktop\Julia\DataStructures.jl\test\test_robin_dict.jl:118
typeof: Test Failed at C:\Users\ww\Desktop\Julia\DataStructures.jl\test\test_robin_dict.jl:148
  Expression: typeof(d) == typeof(d2) == typeof(d3) == RobinDict{Any, Any}
   Evaluated: RobinDict{Any,Any} == Dict{Any,Any} == Dict{Any,Any} == RobinDict{Any,Any}
Stacktrace:
 [1] top-level scope at C:\Users\ww\Desktop\Julia\DataStructures.jl\test\test_robin_dict.jl:148
 [2] top-level scope at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.1\Test\src\Test.jl:1083
 [3] top-level scope at C:\Users\ww\Desktop\Julia\DataStructures.jl\test\test_robin_dict.jl:118
Test Summary: | Pass  Fail  Total
typeof        |   13     3     16
ERROR: LoadError: Some tests did not pass: 13 passed, 3 failed, 0 errored, 0 broken.
```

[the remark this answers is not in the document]
eulerkochy: Probably some bug in the code of `dict_with_eltype`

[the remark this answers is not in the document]
vtjnash: Looks like you're missing:
```julia
empty(d::RobinDict, ::Type{K}, ::Type{V}) where {K, V} = RobinDict{K, V}()
```

src/robin_dict.jl
```diff
@@ -0,0 +1,593 @@
...
+        end
+    end
+    return 0
+end
+
+@propagate_inbounds _iterate(t::RobinDict{K,V}, i) where {K,V} = i == 0 ? nothing : (Pair{K,V}(t.keys[i],t.vals[i]), i == typemax(Int) ? 0 : get_next_filled(t, i+1))
+@propagate_inbounds function iterate(t::RobinDict)
+    _iterate(t, t.idxfloor)
+end
+@propagate_inbounds iterate(t::RobinDict, i) = _iterate(t, get_next_filled(t, i))
```
vtjnash:
```suggestion
@propagate_inbounds iterate(t::RobinDict, i) = _iterate(t, get_next_filled(t, i))

filter!(f, d::RobinDict) = Base.filter_in_one_pass!(f, d)
```

src/robin_dict.jl
chethega: +1 to getting the PR merged quickly and leaving performance optimizations for later, and only merging them if we have an example where they make a difference.

Re "too late for prefetch": We want to fire off the memory load as early as possible. This means firing it off as soon as we have computed the hash, i.e. simultaneously with the memory load for `h.hashes[index]`, instead of waiting until `h.hashes[index]` has arrived and we have compared it to the target `chash`. As I said, this should not matter for low load factors, because speculative execution should fire off the same memory load anyway while waiting for the hash comparison: Spectre-style prefetch. However, if the branch predictor guesses that none of the hashes match, then we risk waiting for a second memory round-trip. Or am I misunderstanding something here?

eulerkochy:
I thought I'll just post the benchmarking results here and keep updating it as I refactor the code.
Here's the result after adding `rh_insert_for_rehash!`:
```julia
julia> include("test/bench_robin_dict.jl")
.
Sample #1 Key => Integer , Size => 10^6 entries
.
        add_entries for RobinDict()
  847.246 ms (5914106 allocations: 196.91 MiB)
        add_entries for Dict()
  1.122 s (5985262 allocations: 156.50 MiB)
        add_entries for RobinDict{Int, Int}()
  182.534 ms (48 allocations: 106.67 MiB)
        add_entries for Dict{Int, Int}()
  112.181 ms (54 allocations: 65.17 MiB)
.
Sample #2 Key => Float32 , Size => 10^6 entries
.
        add_entries for RobinDict()
  821.209 ms (5914028 allocations: 196.91 MiB)
        add_entries for Dict()
  1.107 s (5986513 allocations: 156.52 MiB)
        add_entries for RobinDict{Float32, Float32}()
  149.796 ms (46 allocations: 64.00 MiB)
        add_entries for Dict{Float32, Float32}()
  99.266 ms (52 allocations: 34.50 MiB)
.
Sample #3 Key => String , Size => 10^6 entries
.
        add_entries for RobinDict()
  589.279 ms (1956745 allocations: 136.53 MiB)
        add_entries for Dict()
  568.355 ms (2562043 allocations: 104.26 MiB)
        add_entries for RobinDict{String, String}()
  365.851 ms (48 allocations: 106.67 MiB)
        add_entries for Dict{String, String}()
  439.566 ms (54 allocations: 65.17 MiB)
.
Sample #4 Key => Integer , Size => 10^7 entries
.
        add_entries for RobinDict()
  6.338 s (35658047 allocations: 970.77 MiB)
        add_entries for Dict()
  12.688 s (57823447 allocations: 1.39 GiB)
        add_entries for RobinDict{Int, Int}()
  1.656 s (54 allocations: 426.67 MiB)
        add_entries for Dict{Int, Int}()
  1.544 s (72 allocations: 541.17 MiB)
.
Sample #5 Key => Integer , Size => 10^5 entries
.
        add_entries for RobinDict()
  19.470 ms (444164 allocations: 13.45 MiB)
        add_entries for Dict()
  17.759 ms (396983 allocations: 11.73 MiB)
        add_entries for RobinDict{Int, Int}()
  8.130 ms (36 allocations: 6.67 MiB)
        add_entries for Dict{Int, Int}()
  5.527 ms (36 allocations: 5.67 MiB)
.
Sample #6 Key => Float32 , Size => 10^5 entries
.
        add_entries for RobinDict()
  19.748 ms (444207 allocations: 13.45 MiB)
        add_entries for Dict()
  17.693 ms (399228 allocations: 11.76 MiB)
        add_entries for RobinDict{Float32, Float32}()
  7.281 ms (34 allocations: 4.00 MiB)
        add_entries for Dict{Float32, Float32}()
  4.871 ms (34 allocations: 3.00 MiB)
.
Sample #7 Key => String , Size => 10^5 entries
.
        add_entries for RobinDict()
  22.418 ms (121883 allocations: 8.53 MiB)
        add_entries for Dict()
  18.473 ms (115882 allocations: 7.44 MiB)
        add_entries for RobinDict{String, String}()
  15.422 ms (36 allocations: 6.67 MiB)
        add_entries for Dict{String, String}()
  15.398 ms (36 allocations: 5.67 MiB)
```

Here are the plots, which can provide an insight into the working of `RobinDict` at load factor of 70%
dibs_1000K  @0 7 L F
lf_and_max_probe_1000K @0 7 L F

vtjnash:
> which is pretty bad with respect to memory consumption

FWIW, the growth factor use usually use for AbstractDict is:
`rehash!(h, length(h) > 64000 ? sz*2 : sz*4)`

test/test_robin_dict.jl
```diff
@@ -0,0 +1,413 @@
...
+    h["a"] = 1
+    h[2] = "b"
+    @test h.idxfloor == get_idxfloor(h)
+    pop!(h)
+    @test h.idxfloor == get_idxfloor(h)
+    pop!(h)
+    @test h.idxfloor == get_idxfloor(h) == 0 
+end
+
+@testset "invariants" begin
```
chethega: Very good. So, what I had in mind was something like
```
function check_invariants(h::RobinDict)
cnt = 0
sz = length(h.keys)
for i=1:length(h.keys)
isslotfilled(h, i) || continue
@assert hash_key(h.keys[i]) == h.hashes[i]
@assert (h.hashes[i] & 0x80000000) != 0
cnt += 1
des_ind = desired_index(h.hashes[i], sz)
dist = calculate_distance(h, i)
@assert dist <= h.maxprobe
if dist == des_ind - i
for j = des_ind : (i-1)
@assert  calculate_distance(h, j)>= dist
else
@assert  dist == des_ind - i + sz
for j=des_ind:sz
@assert  calculate_distance(h, j)>= dist
end
for j = 1 : (i-1)
@assert  calculate_distance(h, j)>= dist
end
end
end
@assert cnt == length(h)
end
```
This would be to make the test run compiled and hence fast enough to hopefully catch corner cases.

test/test_robin_dict.jl
```diff
@@ -0,0 +1,413 @@
...
+    for i = 1:500
+        h[i] = i+1
+    end
+    sz = length(h.keys)
+    for i in 1:sz
+        if isslotfilled(h, i)
+            hash = h.hashes[i]
+            des_ind = desired_index(hash, sz)
+            dist = calculate_distance(h, i)
+            @test dist == (i - des_ind)
```
chethega: Shouldn't this test fail if the desired index is near the end and the actual index wraps around?

eulerkochy:
@chethega @vtjnash @oxinabox review ? I'll `rebase` after that

eulerkochy:
Bump

docs/src/robin_dict.md
```diff
@@ -0,0 +1,36 @@
+# RobinDict
+
+`RobinDict` is simply a dictionary which uses Robin Hood hashing technique in the background for performing standard `Dict` operations.
+
+The function-calls of `RobinDict` replicates that of `Dict`.
+
+Examples:
+
+```julia
```
vtjnash: Does this repo support and run doctest? If not, should probably at least use `julia-repl` here.

eulerkochy: this repo supports doctests, i had previously committed to the repo [documentation](https://github.com/JuliaCollections/DataStructures.jl/pull/492)  and used `julia` there, so I guess it should work fine.

eulerkochy:
Bump

src/robin_dict.jl
```diff
@@ -0,0 +1,611 @@
...
+            !all(x->isa(x,Union{Tuple,Pair}),kv)
+            throw(ArgumentError("RobinDict(kv): kv needs to be an iterator of tuples or pairs"))
+        else
+            rethrow(e)
+        end
+    end
+end
+
+hash_key(key) = (hash(key)%UInt32) | 0x80000000
+desired_index(hash, sz) = ((hash) & (sz - 1)) + 1
```
chethega: `desired_index(hash, sz) = (hash & (sz - 1)) + 1`

src/robin_dict.jl
```diff
@@ -0,0 +1,611 @@
...
+function setindex!(h::RobinDict{K,V}, v0, key0) where {K, V}
+    key = convert(K, key0)
+    isequal(key, key0) || throw(ArgumentError("$key0 is not a valid key for type $K"))
+    _setindex!(h, key, v0)
+end
+
+function _setindex!(h::RobinDict{K,V}, key::K, v0) where {K, V}
+    v = convert(V, v0)
+    sz = length(h.keys)
+    (h.count > ROBIN_DICT_LOAD_FACTOR * sz) && rehash!(h, sz<<2)
```
chethega: Always resizing by a factor of 4 looks sketchy (leads to very low load factors and hence memory waste for long-lived dicts). RobinDict should now be much cheaper to rehash than Base dict.

Maybe just copy what Base does? (i.e. use threshold to decide whether to grow by factor 2 or 4)

src/robin_dict.jl
```diff
@@ -0,0 +1,611 @@
...
+function get!(default::Callable, h::RobinDict{K,V}, key0::K) where {K, V}
+    key = convert(K, key0)
+    return _get!(default, h, key)
+end
+
+function _get!(default::Callable, h::RobinDict{K,V}, key::K) where V where K
+    index = rh_search(h, key)
+    
+    index > 0 && return h.vals[index]
+
```
chethega: In theory, we could use the partial traversal from `rh_search` and continue insertion from there, instead of starting from `desired_index` again: `rh_search` aborts at the first entry that would be relocated on insertion. 

Just leaving that here for posterity (can be separate PR to improve perf).

src/robin_dict.jl
eulerkochy: I did that on purpose as this change showed some improvement in benchmarks . In details described [here](https://nextjournal.com/eulerkochy/blog-two-jsoc/)

oxinabox:
Given @chethega  has approved, I am going to merge this.
Nice work @eulerkochy

PR #505, opened 2019-05-22
Same behaviour for get and getkey as with julias base Dict.

altre:
SortedDict get and getkey methods behave differently then the similar methods in base julias Dict implementation in trying to convert the default to the dicts datatype. julias base dict simply returns the default. 

One could argue against this decision in julias base dict, but I tripped over this when trying to use SortedDict as a drop-in replacement for dict.


oxinabox:
Seems like a good idea to me.
You should also change
`function get!(default_::Union{Function,Type}, m::SortedDict{K,D}, k_) where {K,D}`

altre:
I think in the case of `get!` the convert is necessary and with equal behavior in SortedDict and Dict, since get! saves the default value if the key is not present.

oxinabox:
Oh, yeah. Of-course.

LGTM
thanks

PR #509, opened 2019-06-12
CircularBuffer(n)

### LINKED ISSUE #506: CircularBuffer(n) should be the same as CircularBuffer{Any}(n)

scls19fr:
Closes #506


src/circular_buffer.jl
```diff
@@ -16,6 +16,8 @@ mutable struct CircularBuffer{T} <: AbstractVector{T}
     CircularBuffer{T}(capacity::Int) where {T} = new{T}(capacity, 1, 0, Vector{T}(undef, capacity))
 end
 
+CircularBuffer(capacity::Int) = CircularBuffer{Any}(capacity)
```
oxinabox: Do you think we should relax this type-constraint?

PR #512, opened 2019-07-05
Prepare to release v0.15.0

### LINKED ISSUE #477: Clean out deprecations that we declared
We have a lot of old deprecations, some many versions old.
I have just tagged a v0.15.0,
which means next release we are allowed to delete all of them.

Someone just needs to go through  and do so.

https://github.com/JuliaCollections/DataStructures.jl/search?p=1&q=deprecate&unscoped_q=deprecate

oxinabox:
I have just registered for  release 0.14.2
we keep making releases without dropping deprecations.
So release 0.15 will accumulate several years of deprecations,
and just get it done.

Closes #477

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,13 +1,13 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.14.2"
+version = "0.15.0"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 
 [compat]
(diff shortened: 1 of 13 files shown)


Project.toml
```diff
@@ -1,13 +1,13 @@
...
-version = "0.14.2"
+version = "0.15.0"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 
 [compat]
-OrderedCollections = "1.0.1"
+OrderedCollections = "1.1.0"
```
nickrobinson251: should have `compat` for Julia?

oxinabox: yes, and my 0.14.2 tag is about to be failed because of that.

PR #513, opened 2019-07-10
Adding `eltype` for containers

mcognetta:
Most of the containers did not have a defined `eltype` operation. I added it (or corrected the current implementation) for:
- `CircularBuffer`
- `Deque`
- `DisjointSet`
- `FenwickTree`
- `AbstractHeap`
- `IntSet`
- `LinkedList`
- `MutableList`
- `Queue`
- `SortedSet`
- `Stack`

Please let me know if I missed any. I purposefully avoided associative containers in this PR, though it should be defined for them as well. It makes sense to have a Key-Value associative container return `Pair{K, V}` from `eltype` similar to how `Dict` does it in Julia core.

### DIFF
--- a/src/sorted_set.jl
+++ b/src/sorted_set.jl
@@ -180,8 +180,7 @@ end
 Returns the key type for SortedSet.
 This function may also be applied to the type itself. Time: O(1)
 """
-@inline eltype(m::SortedSet{K,Ord}) where {K,Ord <: Ordering} = K
-@inline eltype(::Type{SortedSet{K,Ord}}) where {K,Ord <: Ordering} = K
+@inline Base.eltype(::Type{SortedSet{K,Ord}}) where {K,Ord <: Ordering} = K
 
 """
     keytype(sc)
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -162,6 +162,8 @@ Return the number of elements currently in the buffer.
 """
 Base.length(cb::CircularBuffer) = cb.length
 
+Base.eltype(::Type{CircularBuffer{T}}) where T = T
+
 """
     size(cb)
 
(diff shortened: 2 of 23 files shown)


src/sorted_set.jl
```diff
@@ -180,8 +180,7 @@ end
 Returns the key type for SortedSet.
 This function may also be applied to the type itself. Time: O(1)
 """
-@inline eltype(m::SortedSet{K,Ord}) where {K,Ord <: Ordering} = K
-@inline eltype(::Type{SortedSet{K,Ord}}) where {K,Ord <: Ordering} = K
+@inline Base.eltype(::Type{SortedSet{K,Ord}}) where {K,Ord <: Ordering} = K
```
StephenVavasis: Why are you proposing to remove the method of  `eltype`  that accepts a sorted set instead of its type?  Here is what the documentation (1.1) says about `eltype`:

***Determine the type of the elements generated by iterating a collection of the given type. For dictionary types, this will be a Pair{KeyType,ValType}. The definition eltype(x) = eltype(typeof(x)) is provided for convenience so that instances can be passed instead of types. However the form that accepts a type argument should be defined for new types.***


So according to the docs, it is OK to define a version of `eltype` that takes the container itself.  By removing this statement, you may break people's code that uses this functionality.  Furthermore, `SortedSet` is intended to a mostly drop-in replacement for `Set`, and `Set` supports both `eltype` calls.

mcognetta: `eltype` of an instance falls back to `eltype(typeof(x))` so this won't break any code.

StephenVavasis: Got it, thanks!

mcognetta: I actually included tests for this as well cause I was worried about the same thing.

PR #514, opened 2019-07-10
Adding constructors for Circular(Deque | Buffer) | Deque from iterators

mcognetta:
This adds constructors from iterators for `CircularDeque`, `CircularBuffer`, and `Deque`. I did not include `LinkedList` since the interface does not have a generic append method.

6 tests are skipped as they depend on #513.

This PR also removed an inner constructor for `CircularBuffer` and made it an outer constructor instead. 

Example:

```julia
julia> c = CircularDeque(1:3)
CircularDeque{Int64}([1,2,3])

julia> capacity(c)
3
```

### DIFF
--- a/src/deque.jl
+++ b/src/deque.jl
@@ -80,6 +80,16 @@ Create a deque of type `T`.
 """
 deque(::Type{T}) where {T} = Deque{T}()
 
+# constructor for iterator
+function Deque(iter)
+    n = length(iter)
+    d = deque(eltype(iter))
+    for e in iter
+        push!(d, e)
+    end
+    d
+end
+
 isempty(q::Deque) = q.len == 0
 length(q::Deque) = q.len
 num_blocks(q::Deque) = q.nblocks
--- a/test/test_circ_deque.jl
+++ b/test/test_circ_deque.jl
@@ -80,6 +80,35 @@
         for i in 1:5 push!(D, i) end
         @test collect([i for i in D]) == collect(1:5)
     end
+
+    @testset "constructing from iterator" begin
+        # when eltype is known
+        D = CircularDeque(1:3)
+        @test_skip eltype(D) == Int
+        @test capacity(D) == 3
+        @test length(D) == 3
+
+        # when eltype is not known
+        struct MyIter
+            n::Int
+        end
+        Base.iterate(M::MyIter, state=1) = state > M.n ? nothing : (iseven(state) ? ('a', state+1) : (1, state+1))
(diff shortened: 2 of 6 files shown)


mcognetta:
CI Errors stemmed from https://discourse.julialang.org/t/type-definition-not-allowed-inside-a-local-scope/3907/20. I am attempting to resolve it, though it leads to a bit of redundant code that I will clean up later.

mcognetta:
Bump (I need this for the JuliaCon hackathon actually).

src/deque.jl
```diff
@@ -80,6 +80,16 @@ Create a deque of type `T`.
 """
 deque(::Type{T}) where {T} = Deque{T}()
 
+# constructor for iterator
+function Deque(iter)
+    n = length(iter)
+    d = deque(eltype(iter))
```
kmsquire: The `deque(::Type{T})` call should be deprecated, since we don't construct containers that way anymore. Can you change the call here:
```suggestion
    d = Deque{eltype(iter)}()
```

src/deque.jl
```diff
@@ -80,6 +80,16 @@ Create a deque of type `T`.
 """
 deque(::Type{T}) where {T} = Deque{T}()
 
+# constructor for iterator
+function Deque(iter)
+    n = length(iter)
```
kmsquire: `n` isn't used here.  Did you mean to use it as a size parameter when constructing the `Deque`?  

Obviously, if you use `deque`, that isn't possible, but if you call the main constructor, you can.

That said, since iterators don't necessarily have a known length, I would suggest making it an optional parameter to the function.

```
function Deque(iter, blksize::Int=DEFAULT_DEQUEUE_BLOCKSIZE)
```
and then modify the deque construction below accordingly.

src/deque.jl
kmsquire: Or, if you make `blksize` a parameter:
```suggestion
    d = Deque{eltype(iter)}(blksize)
```

test/test_circ_deque.jl
```diff
@@ -80,6 +92,23 @@
...
+
+    @testset "constructing from iterator" begin
+        # when eltype is known
+        D = CircularDeque(1:3)
+        @test_skip eltype(D) == Int
+        @test capacity(D) == 3
+        @test length(D) == 3
+
+        # when eltype is not known
+        D = CircularDeque(KnownLengthIter(10))
```
kmsquire: The comment above this line isn't quite accurate.  The eltype is `Any`, which is "known".

PR #517, opened 2019-07-20
comprehensive benchmark suite for AbstractDict

eulerkochy:
The objective of this PR is to write a strong and comprehensive (almost) benchmark suite for `AbstractDict`.
I'll add a commit for benchmarking `delete!` operation, having a fixed number of `evals` and estimating it's performance.

### DIFF
--- a/test/bench_robin_dict.jl
+++ b/test/bench_robin_dict.jl
@@ -8,6 +8,9 @@ function add_entries(h::AbstractDict, entries::Vector{Pair{K, V}}) where {K, V}
 	end
 end
 
+find_key_present(h::AbstractDict, entry) = h[entry]
+find_key_absent(h::AbstractDict, entry) = getkey(h, entry, -1)
+
 @printf(".\nSample #1 Key => Integer , Size => 10^6 entries\n.\n")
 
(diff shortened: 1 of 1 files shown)


oxinabox:
Would it be better to write this as a function per benchmark
taking the type/constructor as an input?

So we can write
```
benchmark_FeatureX(Dict)
benchmark_FeatureX(RobinDict)
benchmark_FeatureX(SortedDict)
benchmark_FeatureX(OrderedDict)
benchmark_FeatureX(LittleDict)

benchmark_FeatureY(Dict)
benchmark_FeatureY(...
```

Also,
I am not sure what needs to be done to make this work with PkgBenchmarks.jl
but that should be done.
https://juliaci.github.io/PkgBenchmark.jl/stable/ref.html#PkgBenchmark.benchmarkpkg

eulerkochy:
@oxinabox I am trying to compile information between the performance of `RobinDict` and `Dict`, in terms of time and memory. Will make the requested changes after I'm done with compilation. I prefer making those changes in a separate `bench_dict.jl`, since this is `bench_robin_dict.jl`, I am not in for making the changes here. Once `bench_dict.jl` is up, this file can be scrapped!

PR #523, opened 2019-08-12
MultiDict - store arrays as arrays-of-arrays

### LINKED ISSUE #522: MultiDict: insert!(a, k, ARRAY) appends array instead of adding entry
MultiDict has a very unexpected behavior. If one inserts a key=>val combination, the val is added to an array of previously inserted values of the corresponding key. 
BUT, if val<:AbstractArray, the elements in the val array are appended to the previous array.
This is completely broken behavior and not mentioned in documentation. 
Is it OK to make PR that fixes it or is that a well thought behavior - in that case I'll just make my own MultiDict?

phlavenk:
If merged closes #522.
Remove the automatic array appending functionality. Now arrays are treated as any other object.
Missing: 

- [x] implemented
- [ ] `enumerateall` functionality is broken now
- [ ] some documentation would be nice

### DIFF
--- a/src/multi_dict.jl
+++ b/src/multi_dict.jl
@@ -27,7 +27,7 @@ function multi_dict_with_eltype(kvs, ::Type{Tuple{K,V}}) where {K,V}
 end
 multi_dict_with_eltype(kvs, t) = MultiDict{Any,Any}(kvs)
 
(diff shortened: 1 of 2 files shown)


oxinabox:
@wildart  @malmaud  can one of you take a look at this?

wildart:
LGTM

oxinabox:
I'll trust @wildart and merge then

oxinabox:
Thanks!

Issue #496, opened 2019-04-10
Deprecate then Remove ClassifierCollections


oxinabox:
`ClassifierCollection`s is just a really weird data structure,
that I do not think is at all standard;
and I would be surprised if any julia code compatible with the current version of DataStructures was using it.

Please correct me if I am wrong.

It was added in
 https://github.com/JuliaCollections/DataStructures.jl/commit/064e5732df60d83256c1c852cf9f84dc13e5c39d
and basically had not changed since.
It has been in this package since before we had things like PRs being used by maintainers.
Wildwest days.

Somethings were added because they seemed useful for various projects and needed to go *Somewhere*.
(Similar for some of the weirder stuff in other old packages like StatsBase)
But now this is a package with over 750 dependents,
so we should try and keep to something vaguely standard.

oxinabox:
I do not see what purpose it servers that a `DefaultDict(Vector)` couldn't.

oxinabox:
@lindahua  thoughts?

kmsquire:
I think you can go ahead and remove it. Dahua hasn't been active here in a long time.

PR #524, opened 2019-08-14
deprecate ClassifierCollection

oxinabox:
First part of #496  

I don't know what to recommend using instead.
Probably a `DefaultDict(Collection)`?

### DIFF
--- a/src/classified_collections.jl
+++ b/src/classified_collections.jl
@@ -6,6 +6,11 @@
 
 mutable struct ClassifiedCollections{K, Collection}
     map::Dict{K, Collection}
+
+    function ClassifiedCollections{K, Collection}(map) where {K, Collection}
+        @warn "ClassifiedCollections will be removed from DataStructures.jl in a future release."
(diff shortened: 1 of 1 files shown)


oxinabox:
@kmsquire   can I request a review?
I would like to get this into 0.17.x so can remove in 0.18

src/classified_collections.jl
```diff
@@ -6,6 +6,11 @@
 
 mutable struct ClassifiedCollections{K, Collection}
     map::Dict{K, Collection}
+
+    function ClassifiedCollections{K, Collection}(map) where {K, Collection}
+        @warn "ClassifiedCollections will be removed from DataStructures.jl in a future release."
```
kmsquire: Maybe give the targeted release?

oxinabox:
```suggestion
        @warn "ClassifiedCollections will be removed from DataStructures.jl in v0.18."
```

Issue #532, opened 2019-09-26
Packed(Sparse?)IntSet Implementation


louisponet:
I've recently played around with implementing something like:
[Sparse Set (geeksforgeeks)](https://www.geeksforgeeks.org/sparse-set/),
but without the maximum value limitation and with a paging mechanism to make the reverse actually sparse.The idea is to have fast insertion, `in` operations but also very fast iteration over which indices are actually occupied.

I was wondering if this could find a place here, I can clean up the code and make decent tests for it. There are also the mathematical set functions that I didn't really need and thus didn't implement yet, but I wouldn't mind doing so if there's interest.

I'm using this to handle the indexing for an entity component system, and it seems to be pretty fast and reliable. There is also a fast zip-like iterator that returns the positions in the packed array that correspond to the same index, this can be then used with different component vectors for example. The current implementation is here:
[PackedIntSet](https://github.com/louisponet/PStdLib.jl/blob/master/src/datastructures/PackedIntSet.jl)

oxinabox:
Yes, PR would be welcomed

PR #533, opened 2019-09-28
implemented SparseIntSet

louisponet:
I've implemented the SparseIntSet datastructure as was talked about in #532 
I based the tests and functionality on that from `IntSet`, but I wasn't sure how to do `symdiff`, and also I'm not sure if it even makes sense in this case.

I also added some documentation.

Let me know if there are any things that need to be improved!

Cheers


src/sparse_int_set.jl
```diff
@@ -0,0 +1,258 @@
...
+		push!(set, i)
+	end
+	return set
+end
+
+SparseIntSet(i...) = SparseIntSet(i)
+
+eltype(::SparseIntSet) = Int
+eltype(::Type{SparseIntSet}) = Int
+empty(::SparseIntSet) = SparseIntSet()
```
oxinabox: Should have `empty!` defined here.

src/sparse_int_set.jl
```diff
@@ -0,0 +1,258 @@
+import Base: @propagate_inbounds, zip
+
+const INT_PER_PAGE = div(ccall(:jl_getpagesize, Clong, ()), sizeof(Int))
+
+#TODO: Batch creation and allocation
+mutable struct SparseIntSet
+	packed ::Vector{Int}
+	reverse::Vector{Vector{Int}}
+	function SparseIntSet()
```
oxinabox: Prefer outer constructors, so that it is always possible to construct directly

src/sparse_int_set.jl
```diff
@@ -0,0 +1,258 @@
...
+eltype(::SparseIntSet) = Int
+eltype(::Type{SparseIntSet}) = Int
+empty(::SparseIntSet) = SparseIntSet()
+isempty(s::SparseIntSet) = isempty(s.packed)
+
+copy(s::SparseIntSet) = copy!(SparseIntSet(), s)
+
+function copy!(to::SparseIntSet, from::SparseIntSet)
+	to.packed = copy(from.packed)
+	to.reverse = deepcopy(from.reverse)
```
oxinabox: This will need to change once we switch to outer constructors + immutable structs

src/sparse_int_set.jl
```diff
@@ -0,0 +1,258 @@
...
+function SparseIntSet(indices)
+	set = SparseIntSet()
+	for i in indices
+		i <= 0 && throw(ArgumentError("All Ints need to be positive."))
+		push!(set, i)
+	end
+	return set
+end
+
+SparseIntSet(i...) = SparseIntSet(i)
```
oxinabox: Do other sets have this overload?

src/sparse_int_set.jl
```diff
@@ -0,0 +1,258 @@
...
+	elseif !isassigned(s.reverse, pageid)
+		p = zeros(Int, INT_PER_PAGE)
+		@inbounds s.reverse[pageid] = p 
+		return p, true
+	end
+	return @inbounds s.reverse[pageid], false
+end
+
+function push!(s::SparseIntSet, i::Integer)
+	i <= 0 && throw(ArgumentError("All Ints need to be positive."))
```
oxinabox: A rare chance to use a DomainError, perhaps?

src/sparse_int_set.jl
```diff
@@ -0,0 +1,258 @@
...
+
+function cleanup!(s::SparseIntSet)
+	isused = x -> isassigned(s.reverse, x) && any(y -> y != 0, s.reverse[x])
+	indices = eachindex(s.reverse)
+	last_page_id = findlast(isused, indices)
+	if last_page_id === nothing
+		s.reverse = Vector{Int}[]
+		return s
+	else
+		new_pages    = Vector{Vector{Int}}(undef, last_page_id)
```
oxinabox: Spacing is extra weird here

src/sparse_int_set.jl
```diff
@@ -0,0 +1,258 @@
...
+function in_excluded(id, it)
+	for e in it.excluded_sets
+		if id in e
+			return true
+		end
+	end
+	return false
+end
+
+@inline id_tids(it, state) = (id = it.shortest_set.packed[state]; return id, map(x -> findfirst_packed_id(id, x), it.valid_sets))
```
oxinabox: Is this line <90 characters?
Seems really long
Might want to use a long form function

src/sparse_int_set.jl
louisponet: Hmm right, I think somehow kakoune is using tabs for indentation. This never seemed to have been an issue before but now somehow it does... I'll investigate and change.

src/sparse_int_set.jl
```diff
@@ -0,0 +1,258 @@
+import Base: @propagate_inbounds, zip
+
+const INT_PER_PAGE = div(ccall(:jl_getpagesize, Clong, ()), sizeof(Int))
+
+#TODO: Batch creation and allocation
+mutable struct SparseIntSet
```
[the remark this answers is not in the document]
louisponet: Yea, I didn't see the benefit of immutable if it has `Vector`s anyway, is there any? I don't mind making it immutable, just for my own understanding

src/sparse_int_set.jl
louisponet: Indeed not that I see, I changed it to mimic: [int_set.jl#L14](https://github.com/louisponet/DataStructures.jl/blob/c38959cbf3a771316b9d133be02530f47f8a942b/src/int_set.jl#L14), since all the functionality was in `union!` already anyway.

src/sparse_int_set.jl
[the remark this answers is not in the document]
oxinabox: Yeah, it is better, not entirely sure why
we may have the stack-reference-to-heap-allocated object thing working


```
julia> struct Foo
       x::Vector{Int}
       end

julia> mutable struct MFoo
       x::Vector{Int}
       end

julia> function testem(T)
       first([T([1,2,i]) for i in 1:10_000])
       end
testem (generic function with 1 method)
```

Timing on Julia 1.4-DEV
```
julia> @btime testem(Foo)
  1.507 ms (20005 allocations: 1.30 MiB)
Foo([1, 2, 1])

julia> @btime testem(MFoo)
  1.633 ms (20005 allocations: 1.30 MiB)
MFoo([1, 2, 1])
```

Timing on julia 1.0
```
julia> @btime testem(MFoo)
  1.112 ms (29495 allocations: 1.44 MiB)
MFoo([1, 2, 1])

julia> @btime testem(Foo)
  1.209 ms (29495 allocations: 1.44 MiB)
Foo([1, 2, 1])
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,263 @@
...
+# function SparseIntSet(indices)
+#     set = SparseIntSet()
+#     for i in indices
+#         i <= 0 && throw(DomainError("Only positive Ints allowed."))
+#         push!(set, i)
+#     end
+#     return set
+# end
+
+# SparseIntSet(i...) = SparseIntSet(i)
```
oxinabox: Don't forget to delete these commented out ones

src/sparse_int_set.jl
```diff
@@ -0,0 +1,263 @@
...
+end
+
+function in(i, s::SparseIntSet)
+    pageid, offset = pageid_offset(s, i)
+    isassigned(s.reverse, pageid) && @inbounds s.reverse[pageid][offset] != 0
+end
+
+length(s::SparseIntSet) = length(s.packed)
+
+function assure!(s::SparseIntSet, pageid)
```
oxinabox: can we have a comment about what this does?

src/DataStructures.jl
```diff
@@ -104,4 +104,6 @@ module DataStructures
     export PriorityQueue, peek
 
     include("priorityqueue.jl")
+    include("sparse_int_set.jl")
+    export SparseIntSet, cleanup!, current_id
```
oxinabox: Why are we exporting `current_id` ?

```suggestion
    export SparseIntSet, cleanup!
```

docs/src/sparse_int_set.md
```diff
@@ -0,0 +1,9 @@
...
+Only positive non-zero `Int`s are allowed inside the set. 
+The idea is to have one **packed** `Vector` storing all the `Int`s contained in the set as to allow for fast iteration, and a sparse, paged **reverse** `Vector` with the position of a particular `Int` inside the **packed** `Vector`. This allows for very fast iteration, insertion and deletion of indices.
+Most behavior is similar to a normal `IntSet`, however `collect`, `first` and `last` are with respected to the **packed** vector, in which the ordering is not guaranteed. 
+The **reverse** `Vector` is paged, meaning that it is a `Vector{Vector{Int}}` where each of the `Vector{Int}`s has the length of one memory page of `Int`s. Every time an index that was not yet in the range of the already present pages, a new one will be created and added to the **reverse**, allowing for dynamical growth. 
+If all the indices on a particular page are deleted from the set, it will not automatically get cleaned up for performance reasons. The `cleanup!` method is provided to facilitate the reclaiming of memory of a page when all it's corresponding indices have been deleted from the set.
```
oxinabox: Having to manually trigger cleanup bothers me.
Can we maybe trigger this to run after every N remove operations or something?
Or some how attach it to `sizehint!`
@ararslan  may have thoughts?

src/DataStructures.jl
louisponet: it's useful during iteration if you don't want the indices inside the packed array corresponding to the id but rather the id itself. I did it this way after experimenting with it and finding that making a tuple (id, packed_ids) during iteration impacts performance quite a lot, whereas storing it inside the iterator not at all. But yea, it's maybe not the best way?

docs/src/sparse_int_set.md
louisponet: yes that's certainly possible, I did it like this allowing for advanced users to manage the memory, without impacting the basic performance. Other ways of doing it are definitely worth considering

src/sparse_int_set.jl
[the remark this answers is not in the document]
louisponet: So you mean to keep it nutable for now since it's faster on current Julia, or make it immutable to future proof?

docs/src/sparse_int_set.md
louisponet: I don't think I like attaching it to `sizehint!` really, because I guess that implies that if I want to have index 1 million in there, that would kind of correspond to `sizehint!(1000_000)`, which in itself would mean that all pages for the first 1 million indices should be allocated wasting a lot of memory

src/DataStructures.jl
oxinabox: it has no docs, does it?

src/sparse_int_set.jl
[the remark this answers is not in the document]
oxinabox: oops I didn't splot that it flippped.

on current stable 1.2 it is also faster to be mutable
```
julia> @btime testem(MFoo)
  1.437 ms (20005 allocations: 1.30 MiB)
MFoo([1, 2, 1])

julia> @btime testem(Foo)
  1.571 ms (20005 allocations: 1.30 MiB)
Foo([1, 2, 1])
```

On 1.3-rc2
is where it becomes faster to be immutable
```
julia> @btime testem(Foo)
  1.535 ms (20005 allocations: 1.30 MiB)
Foo([1, 2, 1])

julia> @btime testem(MFoo)
  1.720 ms (20005 allocations: 1.30 MiB)
MFoo([1, 2, 1])
```

Lets go immutable, for the future proofing.
Since 1.3 will become the stable release soon

docs/src/sparse_int_set.md
oxinabox: I am not particularly keen on attaching it to `sizehint` either.

src/DataStructures.jl
louisponet: No indeed, would you feel like keeping it and include documentation, or just leaving it out? I don't know if it's a good api

src/sparse_int_set.jl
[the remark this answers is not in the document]
louisponet: Aside from the fact that mutable vs immutable speed switches, it's kind of sad/scary to see that with each next minor release this is getting slower...

src/DataStructures.jl
louisponet: One possibly convoluted way is to make a `IntPage` datastructure that holds the page and a counter, each time something gets pushed to the page the counter gets incremented, when popped, decremented. Then the set can check whether it's 0 to clean it up. But that will lower performance somewhat I guess since extra instructions per push and pop.

docs/src/sparse_int_set.md
louisponet: One possibly convoluted way is to make a IntPage datastructure that holds the page and a counter, each time something gets pushed to the page the counter gets incremented, when popped, decremented. Then the set can check whether it's 0 to clean it up. But that will lower performance somewhat I guess since extra instructions per push and pop.

ararslan: That kind of reference counting of sorts seems worth the overhead to me; having to manually manage memory is a bit unusual in Julia.

louisponet: One issue I have with something like this is that the worst-case scenario (pushing and popping the same single value a lot) is really really bad for performance. I would also really like for a way to just omit the automatic cleanup if you don't want it, maybe this can be done with two different `Page` types, and a parameter in the `SparseIntSet` which defaults to the automatically cleaned one. If this seems reasonable I think I'll have a go at that later

louisponet: On a semi side-note, is there a direct way of making an entry in a vector again `undefined` or  something like `unassign`? On what is this identification based under the hood?
That would make `cleanup!` probably a little bit better since I don't need to create a whole new page `Vector`

louisponet: I ended up sticking to a less over-engineered solution storing the counters in the set itself and adding a `dirty_pop!` that doesn't do any cleanup. There's a 30% overhead of a normal `pop!` that needs no cleaning vs a `dirty_pop!`, which is kind of a lot imo. If there are any suggestions on making this less severe that would be nice, but I don't think it can be reduced too much more.

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
...
+    return s
+end
+
+intersect!(s1::SparseIntSet, ss...) = intersect!(s1, intersect(ss...))
+
+#Is there a more performant way to do this?
+intersect!(s1::SparseIntSet, ns) = copy!(s1, intersect(s1, ns))
+
+setdiff(s::SparseIntSet, ns) = setdiff!(copy(s), ns)
+setdiff!(s::SparseIntSet, ns) = (for n in ns; pop!(s, n, nothing); end; s)
```
oxinabox: Multistatement single line functions are hard to read.
```suggestion
function setdiff!(s::SparseIntSet, ns)
    for n in ns
        pop!(s, n, nothing)
    end
    return s
end
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
...
+
+function ==(s1::SparseIntSet, s2::SparseIntSet)
+    length(s1) != length(s2) && return false
+    return all(x -> in(x, s1), s2)
```
oxinabox: Can use the single argument form of `in` for this:

```suggestion
    return all(in(s1), s2)
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
+import Base: @propagate_inbounds, zip
+
+const INT_PER_PAGE = div(ccall(:jl_getpagesize, Clong, ()), sizeof(Int))
+
+
+#TODO: Batch creation and allocation
+struct SparseIntSet
+    packed ::Vector{Int}
+    reverse::Vector{Vector{Int}}
+    counters::Vector{Int}
```
oxinabox:
```suggestion
    counters::Vector{Int}  # counts the number of real elements in each page of reverse. 
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
...
+                end
+            end
+        end
+    end
+    return b
+end
+#Can this be optimized?
+complement!(a::SparseIntSet) = copy!(a, complement(a))
+
+<(a::SparseIntSet, b::SparseIntSet) = (a<=b) && !isequal(a,b)
```
oxinabox:
```suggestion
<(a::SparseIntSet, b::SparseIntSet) = ( a<=b ) && !isequal(a, b)
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
...
+        else
+            new_pages    = Vector{Vector{Int}}(undef, last_page_id)
+            new_counters = zeros(Int, last_page_id)
+            for i in eachindex(s.reverse)
+                if isused(i)
+                    new_pages[i] = s.reverse[i]
+                    new_counters[i] = s.counters[i]
+                end
+            end
+            resize!(s.reverse, last_page_id)
```
oxinabox: if we do this `resize!` at the start then it is less work for the loop

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
...
+function findfirst_packed_id(i, s::SparseIntSet)
+    pageid, offset = pageid_offset(s, i)
+    if isassigned(s.reverse, pageid)
+        @inbounds id = s.reverse[pageid][offset]
+        return id
+    end
+    return 0
+end
+
+function cleanup!(s::SparseIntSet)
```
oxinabox: Here is the version of all my changes in the comments below.

is it correct and the same as current version?

```julia
function cleanup!(s::SparseIntSet)
    last_page_id = findlast(!iszero, s.counters)
    if last_page_id === nothing
        empty!(s.reverse)
        empty!(s.counters)
        return s
    else
        # shrink anything off the end
        resize!(s.counters, last_page_id)
        resize!(s.reverse, last_page_id)

        # Now we need to undef any page in `reverse` that is empty.
        # so we are allocating new pages, and then will copy it across.
        # since can't assign `undef` directly.
        new_pages = Vector{Vector{Int}}(undef, last_page_id)
        for i in eachindex(s.reverse)
            isused =  s.counters[i] !== 0 && isassigned(s.reverse, i)
            if isused
                new_pages[i] = s.reverse[i]
            end
        end
        s.reverse .= new_pages
        return s
    end
end
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
...
+issubset(a::SparseIntSet, b::SparseIntSet) = isequal(a, intersect(a, b))
+
+complement(a::SparseIntSet) = complement!(SparseIntSet(), a)
+function complement!(b::SparseIntSet, a::SparseIntSet)
+    empty!(b)
+    for i in eachindex(a.reverse)
+        if !isassigned(a.reverse, i)
+            resize!(b.reverse, i)
+            resize!(b.counters, i)
+            new_ids = (i-1)*INT_PER_PAGE+1:(i)*INT_PER_PAGE
```
oxinabox: I think clearer:
```suggestion
            new_ids = ((i - 1) * INT_PER_PAGE + 1) : (i * INT_PER_PAGE)
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
...
+        shortest = valid_sets[findmin(map(x->length(x), valid_sets))[2]]
+        new{typeof(valid_sets), typeof(exclude)}(zero(eltype(shortest)), valid_sets, shortest, exclude)
+    end
+end
+
+zip(s::SparseIntSet...;kwargs...) = ZippedSparseIntSetIterator(s...;kwargs...)
+
+@inline length(it::ZippedSparseIntSetIterator) = length(it.shortest_set)
+
+in_excluded(id, it::ZippedSparseIntSetIterator{VT,Tuple{}}) where {VT} = false
```
oxinabox:
```suggestion
# we know it is not in_excluded, as there are no excluded
in_excluded(id, it::ZippedSparseIntSetIterator{VT,Tuple{}}) where {VT} = false
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
...
+    id = it.shortest_set.packed[state]
+    return id, map(x -> findfirst_packed_id(id, x), it.valid_sets)
+end
+
+@propagate_inbounds function iterate(it::ZippedSparseIntSetIterator, state=1)
+    if state > length(it)
+        return nothing
+    end
+    id, tids = id_tids(it, state)
+    while !all(x -> x!=0, tids) || in_excluded(id, it)
```
oxinabox: Is this application of De Morgan's laws correct?
```suggestion
    while any(iszero, tids) || in_excluded(id, it)
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
...
+    in(id, s) ? (@inbounds pop!(s, id)) : default
+end
+popfirst!(s::SparseIntSet) = pop!(s, first(s))
+
+iterate(set::SparseIntSet, args...) = iterate(set.packed, args...) 
+
+last(s::SparseIntSet) = isempty(s) ? throw(ArgumentError("Empty set has no last element.")) : last(s.packed)
+
+union(s::SparseIntSet, ns) = union!(copy(s), ns)
+union!(s::SparseIntSet, ns) = (for n in ns; push!(s, n); end; s)
```
oxinabox: Multistatement single line functions are hard to read
```suggestion
function union!(s::SparseIntSet, ns)
    for n in ns
        push!(s, n)
    end
    return s
end
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,292 @@
...
+
+@propagate_inbounds function pop!(s::SparseIntSet, id::Integer)
+    id = dirty_pop!(s, id)
+    cleanup!(s)
+    return id
+end
+
+function pop!(s::SparseIntSet, id::Integer, default)
+    id < 0 && throw(ArgumentError("Int to pop needs to be positive."))
+    in(id, s) ? (@inbounds pop!(s, id)) : default
```
oxinabox: For code clarity it is good to always state the return.
```suggestion
    return in(id, s) ? (@inbounds pop!(s, id)) : default
```

louisponet:
Okay I think I implemented all your suggestions and made it mutable again. I also made cleaned up `cleanup!` adding an `elseif any(iszero, s.counters)` just so the cleanup only actually happens when needed.

src/sparse_int_set.jl
```diff
@@ -0,0 +1,291 @@
...
+# This makes sure that when adding (pushing) an Int,
+# it's respective page is allocated and put at the index of the reverse such that
+# pageid_offset works as intended.
+# Pages will be allocated only once, when pushing an Int that belongs to them.
+# Other not used pages (created during resize!) will be undefs until one Int belonging to them gets added. 
+function assure!(s::SparseIntSet, pageid)
+    if pageid > length(s.reverse)
+        diff = pageid - length(s.reverse)
+
+        resize!(s.reverse, pageid - 1)
```
oxinabox: Probably better not to do that?  `push!` will resize for you.
And it has heuristics to better handle the resizing

src/sparse_int_set.jl
```diff
@@ -0,0 +1,291 @@
...
+    end
+    @inbounds begin
+        packed_endid           = s.packed[end] 
+        from_page, from_offset = pageid_offset(s, id)
+        to_page, to_offset     = pageid_offset(s, packed_endid)
+
+        packed_id                         = s.reverse[from_page][from_offset]
+        s.packed[packed_id]               = packed_endid
+        s.reverse[to_page][to_offset]     = s.reverse[from_page][from_offset]
+        s.reverse[from_page][from_offset] = 0
```
oxinabox: only 1 space per on each size of the `=` is the convention followed in this package

oxinabox:
I've been thinking about this a bit.
I think cleanup should replace the vectors it wants to remove with `Vector{Int}()` a zero length vector. We can even define just one of those. Call it `const NULL_PAGE=Vector{Int}()` 
Then we have `assure!` check `===NULL_PAGE` instead of checking `isassigned`.

If we make a function `_pop_and_pageid` that does a `pop` and also returns both the `pop`ed value,
and the `pageid` it came from,
then `cleanup!` would not have to search the whole array to decide what to clean,
it would just check if that counter is zero, 
and if so would replace that page with the `NULL_PAGE`
and if we have made it that cheap,
then maybe we don't need to have `pop!` and `dirty_pop!` and `cleanup!`
since we can just do that check of if the counter is equal to `zero`
at every removal,
and then occationally it will be, but that operation will be cheap as well.

We would never get to resize `reverse` down, but I think that is ok, generally we can't do that very often anyway, since it would need everything removed to come from the far end.

What do you think?

src/sparse_int_set.jl
louisponet: Not sure I understand what you mean, do you mean to move `assure!` into `push!` I guess that's ok since I don't think it's used anywhere else

louisponet:
That sounds reasonable to me, I find the whole `undef` thing a little fuzzy to begin with. I guess the difference in memory usage for `undef` vs a  `Vector{Int}()` is not huge anyway. I'll have a go at that

src/sparse_int_set.jl
oxinabox: no, I mean why are we doing `resize!` then `push!` ?

oxinabox:
>  I guess the difference in memory usage for undef vs a Vector{Int}() is not huge anyway

It should be nothing at all. (baring the single allocation for the global constant `NULL_PAGE`)
Because at the lowest level undef should be a pointer to 0x0000_000...,
and pointer to a constant will be some other pointer of same size

louisponet:
Okay cool I got it working, it's definitely a lot cleaner, I like it! Some quick little benchmarks show it to be between 5-7% slower compared to pop without doing anything else, so that's good. I just realized that I forgot to change the docs

oxinabox:
> One other thing I think is a tad sketchy is the complement stuff, I mean it works, but since there is no fixed definition of the size of the universe of Ints, it's defined in terms of the pages that have ever been in use. It's a little strange to me

Feel free to drop it, if you don't like it.
We have `setdiff` which to me is much more useful.


> Some quick little benchmarks show it to be between 5-7% slower compared to pop without doing anything else, so that's good.

That is good. I think it is acceptable to keep things simple.
Might be able to cut it down more with some `@inbounds` ?
Since `cleanup!` isn't exported anymore right, so could just do that there.
(else would need to `@propagate_inbounds` and do it callsite right?)

docs/src/sparse_int_set.md
```diff
@@ -0,0 +1,9 @@
+# DataStructures.SparseIntSet
+
+Implementation of a __Sparse Integer Set__, for background see [Sparse Sets](https://www.computist.xyz/2018/06/sparse-sets.html).
+Only positive non-zero `Int`s are allowed inside the set. 
+The idea is to have one **packed** `Vector` storing all the `Int`s contained in the set as to allow for fast iteration, and a sparse, paged **reverse** `Vector` with the position of a particular `Int` inside the **packed** `Vector`. This allows for very fast iteration, insertion and deletion of indices.
+Most behavior is similar to a normal `IntSet`, however `collect`, `first` and `last` are with respected to the **packed** vector, in which the ordering is not guaranteed. 
+The **reverse** `Vector` is paged, meaning that it is a `Vector{Vector{Int}}` where each of the `Vector{Int}`s has the length of one memory page of `Int`s. Every time an index that was not yet in the range of the already present pages, a new one will be created and added to the **reverse**, allowing for dynamical growth. 
+Popping the last `Int` of a particular page will automatically trigger `cleanup!` which cleans up the memory of any pages that aren't in use.
```
oxinabox:
```suggestion
Popping the last `Int` of a particular page will automatically cleans up the memory of that page.
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,265 @@
+import Base: @propagate_inbounds, zip
+
+const INT_PER_PAGE = div(ccall(:jl_getpagesize, Clong, ()), sizeof(Int))
+const NULL_INT_PAGE = Vector{Int}()
```
oxinabox:
```suggestion
# we use this to mark pages not in use, it must never be written to.
const NULL_INT_PAGE = Vector{Int}()
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,265 @@
...
+
+#TODO: Batch creation and allocation
```
oxinabox: delete comment and open issue

src/sparse_int_set.jl
```diff
@@ -0,0 +1,265 @@
...
+end
+
+function pageid_offset(s::SparseIntSet, i)
+    pageid = div(i - 1, INT_PER_PAGE) + 1
+    return pageid, (i - 1) & (INT_PER_PAGE - 1) + 1
+end
+
+function in(i, s::SparseIntSet)
+    pageid, offset = pageid_offset(s, i)
+    isassigned(s.reverse, pageid) && @inbounds s.reverse[pageid][offset] != 0
```
oxinabox:
```suggestion
    page = @inbounds s.reverse[pageid]
    return page !=== NULL_INT_PAGE && @inbounds page[offset] != 0
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,265 @@
...
+
+function push!(s::SparseIntSet, i::Integer)
+    i <= 0 && throw(DomainError("Only positive Ints allowed."))
+
+    pageid, offset = pageid_offset(s, i)
+    pages = s.reverse
+    plen = length(pages)
+
+    if pageid > plen
+        #create new null pages up to pageid and a virgin one at pageid
```
oxinabox: clearer i think:
```suggestion
        # Create new null pages up to pageid and fresh (zero-filled) one at pageid
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,265 @@
...
+        sizehint!(pages, pageid)
+        sizehint!(s.counters, pageid)
+        for i = 1:pageid - plen - 1
```
oxinabox:
```suggestion
        for i in 1:pageid - plen - 1
```

src/sparse_int_set.jl
oxinabox: Need a test that catches this case

src/sparse_int_set.jl
```diff
@@ -0,0 +1,265 @@
...
+function findfirst_packed_id(i, s::SparseIntSet)
+    pageid, offset = pageid_offset(s, i)
+    if pageid > length(s.counters) || s.counters[pageid] == 0
+        return 0
+    end
+    @inbounds id = s.reverse[pageid][offset]
+    return id
+end
+
+collect(s::SparseIntSet) = s.packed
```
oxinabox: Don't want to be returning something that could mutate the underlying data.

```suggestion
collect(s::SparseIntSet) = copy(s.packed)
```

test/test_sparse_int_set.jl
```diff
@@ -0,0 +1,244 @@
...
+        @test_throws DomainError SparseIntSet(-1)
+        @test !(-1 in SparseIntSet(1:10))
+    end
+    @testset "Copy, copy!, empty" begin
+        s1 = SparseIntSet([1,2,3])
+        s2 = empty(s1)
+        copy!(s2, s1)
+        s3 = copy(s2)
+        @test s3 == s2 == s1
+        @test collect(s3) == collect(s2) == [1,2,3]
```
oxinabox: Add test here showing that mutating various copies do not mutate original?

test/test_sparse_int_set.jl
```diff
@@ -0,0 +1,244 @@
...
+        @test SparseIntSet(2:2:10) < SparseIntSet(1:10)
+        @test !(SparseIntSet(2:2:10) < SparseIntSet(2:2:10))
+        @test SparseIntSet(2:2:10) <= SparseIntSet(2:10)
+        @test SparseIntSet(2:2:10) <= SparseIntSet(2:2:10)
+        @test SparseIntSet(1) < complement!(SparseIntSet([5]))
+        @test SparseIntSet(1) <= complement!(SparseIntSet([20]))
+        @test !(SparseIntSet(1) < complement!(SparseIntSet(1)))
+    end
+
+    @testset "other 1" begin
```
oxinabox: What is this?

test/test_sparse_int_set.jl
```diff
@@ -0,0 +1,244 @@
...
+        @test pop!(s) == 1000
+        @test s == SparseIntSet([1, 2, 100])
+        push!(s, 5000)
+        push!(s, 2000)
+        pop!(s, 5000)
+        @test s.reverse[end] === DataStructures.NULL_INT_PAGE
+
+    end
+
+    @testset "other 2" begin
```
oxinabox: what is this?

src/sparse_int_set.jl
louisponet: Before this was to generate the `undef`s but now indeed i'm just using `push!` after `sizehint!`, I don't know if that's optimal

test/test_sparse_int_set.jl
louisponet: yea I literally straight copied it from the `test_int_set.jl` so not sure where that naming came from.

src/sparse_int_set.jl
louisponet: Wait a second, won't this potentially segfault if the pageid is larger than the amount of pages?

louisponet: I added a length check, seems to be fine now.

src/sparse_int_set.jl
oxinabox: it is

louisponet: did a quick benchmark with doing `resize!` followed by assignment to index, difference is basically swallowed by noise on the benchmark it seems.

louisponet:
About the iteration through `zip`s (my main usecase, like in the benchmark), is there any way to speed this up? I guess not really because it's mainly the random access pattern taking up most time rather than the logic itself?

I also ran into a strange inference issue on `id and tids` when I didn't have the `id_tids` function, and just did it inside the `iterate` function. I don't get how inference could fail since literally everything is `Int`s.

oxinabox:
I will look into it in 20 hours time

src/sparse_int_set.jl
```diff
@@ -0,0 +1,252 @@
+import Base: @propagate_inbounds, zip
```
oxinabox: imports should be at to of the main DataStructures.jl file.
and ratehr than `import` use `using`, and then to overload `zip` define the function as `Base.zip(`
```
import Base: @propagate_inbounds, zip
```

louisponet:
Okay, then for sure int_set.jl should be cleaned up too, since it imports
everything there.

src/sparse_int_set.jl
louisponet: robin_dict.jl and int_set.jl also import a lot of stuff there, they will also have to be cleaned up.

src/sparse_int_set.jl
```diff
@@ -0,0 +1,250 @@
...
+    if pageid > length(s.counters) || s.counters[pageid] == 0
+        return 0
+    end
+    @inbounds id = s.reverse[pageid][offset]
+    return id
+end
+
+collect(s::SparseIntSet) = copy(s.packed)
+
+mutable struct ZippedSparseIntSetIterator{VT,IT}
```
oxinabox: this can be immutable.

src/sparse_int_set.jl
```diff
@@ -0,0 +1,250 @@
...
+    current_id::Int
```
oxinabox: This is never used, and should be never used as the state should (where possible) be the only thing that uis changed during iteration.
```suggestion
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,250 @@
...
+    id, tids = id_tids(it, state)
+    while any(iszero, tids) || in_excluded(id, it)
+        state += 1
+        if state > length(it)
+            return nothing
+        end
+
+        id, tids = id_tids(it, state)
+    end
+    it.current_id = id
```
oxinabox: This is never used
```suggestion
```

src/sparse_int_set.jl
```diff
@@ -0,0 +1,250 @@
...
+end
+
+Base.@propagate_inbounds function iterate(it::ZippedSparseIntSetIterator, state=1)
+    if state > length(it)
+        return nothing
+    end
+    id, tids = id_tids(it, state)
+    while any(iszero, tids) || in_excluded(id, it)
+        state += 1
+        if state > length(it)
```
oxinabox: Assigning `length(it)` to a constant outside the loop, seems to give a very marginal improvement in performance.

Tht is all I have managed to find so far

src/sparse_int_set.jl
oxinabox: yeah, we do still need to do more housekeeping on the repo.
But lets not make it worse

oxinabox:
I think with the last few comments resolve we will be able to merge this

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.0"
+version = "0.18.0"
```
oxinabox:
```suggestion
version = "0.17.1"
```

This is a nonbreaking change, which means that pre-1.0.0, it bumps the final digit

src/sparse_int_set.jl
```diff
@@ -0,0 +1,249 @@
...
+    id = it.shortest_set.packed[state]
+    return id, map(x -> findfirst_packed_id(id, x), it.valid_sets)
+end
+
+Base.@propagate_inbounds function iterate(it::ZippedSparseIntSetIterator, state=1)
+    if state > length(it)
+        return nothing
+    end
+    id, tids = id_tids(it, state)
+    il = length(it)
```
oxinabox: If this was done at the start of the function it could also be used in the first `if`

Also why `il` and not something more descriptive, like `it_len` or even `iterator_length`

louisponet: yes, sorry I was too fast in doing this my bad

louisponet:
Great! thanks for all your help with this!

oxinabox:
no problem, thanks for your contribution😀

PR #536, opened 2019-10-01
Adding == and empty! for queue

mcognetta:
These were missing from `Queue` but present in `Stack` and `Deque`.


oxinabox:
Looks good to me, thanks!
I've just pushed a commit to breakup your new tests into their own testsets.
(This package doesn't use as many small testsets as it should for historical reasons: it is older than testsets. So I get why it seems like adding more tests to the end of one of the large testsets seemed reasonable.)

Assuming I didn't break the build, will merge soon.

PR #540, opened 2019-10-05
SparseIntSet speedups

louisponet:
So far,  the inlining made the pop!/push! benchmark around 15% faster, the new iteration style improved the iteration speed with 25-30%. 
So far I don't see any other directly obvious speedups.

### DIFF
--- a/src/sparse_int_set.jl
+++ b/src/sparse_int_set.jl
@@ -241,10 +241,10 @@ end
         return nothing
     end
     for i in state:iterator_length
-	    @inbounds id = it.shortest_set.packed[i]
-		if in_valid(id, it) && !in_excluded(id, it)
-			return map(x->findfirst_packed_id(id, x), it.valid_sets), i + 1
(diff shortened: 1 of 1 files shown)


src/sparse_int_set.jl
```diff
@@ -226,24 +226,25 @@ function in_excluded(id, it)
...
-        id, tids = id_tids(it, state)
-    end
-    return tids, state + 1
+    for i in state:iterator_length
+	    @inbounds id = it.shortest_set.packed[i]
+		if in_valid(id, it) && !in_excluded(id, it)
+			return map(x->findfirst_packed_id(id, x), it.valid_sets), i + 1
+		end
+	end
+	return nothing
```
oxinabox: indenting seems to have flipped over to tabs again 😂

louisponet: oh yes, dammit! I need to figure out why that happens 😂

PR #543, opened 2019-10-20
CircularBuffer: Added first/last methods that slightly improve performance

sairus7:
With less call indirection and calculations.

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -193,3 +193,26 @@ capacity(cb::CircularBuffer) = cb.capacity
 Test whether the buffer is full.
 """
 isfull(cb::CircularBuffer) = length(cb) == cb.capacity
+
+"""
+    first(cb)
+
+Get the first element of CircularBuffer.
+"""
+Base.@propagate_inbounds function Base.first(cb::CircularBuffer)
+    @boundscheck if cb.length == 0
+        throw(BoundsError(cb, 1))
+    end
+    cb.buffer[cb.first]
+end
+"""
+    last(cb)
+
+Get the last element of CircularBuffer.
+"""
+Base.@propagate_inbounds function Base.last(cb::CircularBuffer)
+    @boundscheck if cb.length == 0
+        throw(BoundsError(cb, 1))
+    end
+    cb.buffer[_buffer_index(cb, cb.length)]
+end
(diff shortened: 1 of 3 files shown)


sairus7:
How can I do that? I've added release tag v0.17.3 to my master branch, but nothing happens here.

sairus7:
I've added also a `resize!` method for circular buffer.

oxinabox:
You need to edit the version number in the Project.toml

src/circular_buffer.jl
```diff
@@ -193,3 +193,46 @@ capacity(cb::CircularBuffer) = cb.capacity
...
+
+Resize CircularBuffer to the maximum capacity of n elements. If n is smaller than the current buffer length, the first n elements will be retained.
+"""
+function Base.resize!(cb::CircularBuffer{T}, n::Int) where {T}
+    if n != capacity(cb)
+        buf_new = Vector{T}(undef, n)
+        len_new = min(length(cb), n)
+        for i = 1 : len_new
+            @inbounds buf_new[i] = cb[i]
+        end
```
oxinabox: Why not just call `resize!(cb.buffer)`

sairus7: Because `cb.first` can point to any index within old size.

sairus7: Actually, it can be `circshift`'ed: [first:length, 1:first]

sairus7:
Changed version in Project.toml from v0.17.2 to v0.17.3

src/circular_buffer.jl
oxinabox: You can use `splice!` to delete inplace,
It should even avoid copies when only deleting from one end or the other

sairus7: In the most complex case, where buffer is not full, 
e.g. `cb.first` = 5, `cb.length` = 6, `cb.capacity` = 8: [5 6 x x 1 2 3 4], and:
a) resized to size-1 => [5 6 x 1 2 3 4] (deleted from middle)
b) resized to size-5 => [1 2 3] (deleted from both sides)
c) resized to size+1 => [1 2 3 4 5 6 x x x] (circshifted and added to the end)
So, should we really detect all possible cases, or end-up just with copying to another array?

oxinabox: More broadly it is unclear,
should `resize!` upwards only increase the capacity
and not the length  (docstring currently says that, but code increases capacity and length).

It is not clear what this is for to me, and if this is the best  way to do this,
and wether it belongs here, or in the `CircularDeque`.

Might be best to let this PR just do `first` and `last`
and make a follow up to add this.

sairus7: Ok, I've reverted that change, will be another PR.

PR #545, opened 2019-10-25
Use separate files for package benchmarks

milesfrain:
An additional enhancement (for a followup PR) is to rely on a consistent naming convention and either:
* Automatically find all valid `bench_` files. See [Transducers.jl](https://github.com/tkf/Transducers.jl/blob/master/benchmark/benchmarks.jl)
* List all benchmarks to include. See [DataStructures.jl tests](https://github.com/JuliaCollections/DataStructures.jl/blob/master/test/runtests.jl)


oxinabox:
thanks! I have been wanting this

PR #546, opened 2019-10-25
Expand benchmarking in anticipation of heap ordering change

milesfrain:
This PR includes #545.
It should be rebased after the above is merged in order for CI to track benchmarking differences.

PR #547, opened 2019-10-25
Use Base.Ordering for heap, and other performance improvements

milesfrain:
This merge sequence should be followed for generating useful benchmark comparisons:
1. #545 splits existing benchmarks into separate files
2. #546 adds more heap benchmarks
3. #547 makes significant heap changes

@oxinabox This change introduces API changes, and deprecation warnings should be added. Could we replace the previous deprecation warnings with a version bump?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 environment:
   matrix:
-  - julia_version: 0.7
+  - julia_version: 1.0
   - julia_version: latest
 
 platform:
--- a/benchmark/bench_heap.jl
+++ b/benchmark/bench_heap.jl
@@ -26,13 +26,12 @@ heaptypes = [BinaryHeap, MutableBinaryHeap]
 aexps = [1,3]
 datatypes = [Int, Float64]
 baseorderings = Dict(
-    "Min" => DataStructures.LessThan,
-    #"Max" => DataStructures.GreaterThan,
+    "Min" => Base.ForwardOrdering,
+    #"Max" => Base.ReverseOrdering,
     )
 fastfloatorderings = Dict(
-    # These will be enabled upon reordering change
-    #"FastMin" => DataStructures.FasterForward(),
-    #"FastMax" => DataStructures.FasterReverse(),
+    "Min" => DataStructures.FasterForward,
+    "Max" => DataStructures.FasterReverse,
     )
 
 for heap in heaptypes
@@ -41,7 +40,8 @@ for heap in heaptypes
             Random.seed!(0)
             a = rand(dt, 10^aexp)
 
-            orderings = baseorderings
+            # Dict types to force use of abstract type if containing single value
+            orderings = Dict{String, DataType}(baseorderings)
             if dt == Float64
                 # swap to faster ordering operation
                 for (k,v) in orderings
@@ -66,38 +66,22 @@ for heap in heaptypes
     end
 end
 
-# Quick check to ensure no Float regressions with Min/Max convenience functions
-# These don't fit in well with the above loop, since ordering is hardcoded.
-heapalias = Dict(
-    "BinaryMinHeap" => BinaryMinHeap,
-    "BinaryMaxHeap" => BinaryMaxHeap,
-    "BinaryMinMaxHeap" => BinaryMinMaxHeap, # <- no alias issue
-)
-for (heapname, heap) in heapalias
-    for aexp in aexps
-        for dt in [Float64]
-            Random.seed!(0)
-            a = rand(dt, 10^aexp)
-            prepath = [heapname]
-            postpath = [string(dt), "10^"*string(aexp)]
-            suite[vcat(prepath, ["make"], postpath)] =
-                @benchmarkable $(heap)($a)
-            suite[vcat(prepath, ["push"], postpath)] =
-                @benchmarkable push_heap(h, $a) setup=(h=$(heap){$dt}())
-            suite[vcat(prepath, ["pop"], postpath)] =
-                @benchmarkable pop_heap(h) setup=(h=$(heap)($a))
-        end
-    end
-end
+fast_extreme_orderings = Dict(
+    nsmallest => DataStructures.FasterForward(),
+    nlargest => DataStructures.FasterReverse(),
+    )
 
 for func in [nlargest, nsmallest]
+    fastord = fast_extreme_orderings[func]
     for aexp in [4]
         Random.seed!(0);
         a = rand(10^aexp);
         for nexp in [2]
             n = 10^nexp
-            suite[[string(func), "a=rand(10^"*string(aexp)*")", "n=10^"*string(nexp)]] =
+            suite[["Slow " * string(func), "a=rand(10^"*string(aexp)*")", "n=10^"*string(nexp)]] =
                 @benchmarkable $(func)($n, $a)
+            suite[[string(func), "a=rand(10^"*string(aexp)*")", "n=10^"*string(nexp)]] =
+                @benchmarkable DataStructures.nextreme($fastord, $n, $a)
         end
     end
 end
--- a/src/heaps.jl
+++ b/src/heaps.jl
@@ -24,7 +24,7 @@
 #
 #   - sizehint!(h, s)     set size hint to a heap
 #
-#   - first(h)              return the first (top) value of a heap
+#   - first(h)            return the first (top) value of a heap
 #
 #   - pop!(h)             removes the first (top) value, and
 #                         returns it
@@ -37,7 +37,7 @@
 #   - update!(h, i, v)    updates the value of an element
 #                         (referred to by the handle i)
 #
-#   - delete!(h, i)      deletes the node with
+#   - delete!(h, i)       deletes the node with
 #                         handle i from the heap
 #
 #   - top_with_handle(h)  return the top value of a heap
@@ -55,28 +55,24 @@ abstract type AbstractMutableHeap{VT,HT} <: AbstractHeap{VT} end
 
 abstract type AbstractMinMaxHeap{VT} <: AbstractHeap{VT} end
 
-# comparer
-
-struct LessThan
-end
-
-struct GreaterThan
-end
-
-compare(c::LessThan, x, y) = x < y
-compare(c::GreaterThan, x, y) = x > y
-
 # heap implementations
 
 include("heaps/binary_heap.jl")
 include("heaps/mutable_binary_heap.jl")
-include("heaps/arrays_as_heaps.jl")
 include("heaps/minmax_heap.jl")
 
 # generic functions
 
 Base.eltype(::Type{<:AbstractHeap{T}}) where T = T
 
+"""
+    extract_all!(h)
+
+Return an array of heap elements in sorted order (heap head at first index).
+
+Note that for simple heaps (not mutable or minmax)
+sorting the internal array of elements in-place is faster.
+"""
 function extract_all!(h::AbstractHeap{VT}) where VT
     n = length(h)
     r = Vector{VT}(undef, n)
@@ -86,6 +82,14 @@ function extract_all!(h::AbstractHeap{VT}) where VT
     return r
 end
 
+"""
+    extract_all_rev!(h)
+
+Return an array of heap elements in reverse sorted order (heap head at last index).
+
+Note that for simple heaps (not mutable or minmax)
+sorting the internal array of elements in-place is faster.
+"""
 function extract_all_rev!(h::AbstractHeap{VT}) where VT
     n = length(h)
     r = Vector{VT}(undef, n)
@@ -97,30 +101,31 @@ end
 
 # Array functions using heaps
 
-function nextreme(comp::Comp, n::Int, arr::AbstractVector{T}) where {T, Comp}
+"""
+    nextreme(ord, n, arr)
+
+return an array of the first `n` values of `arr` sorted by `ord`.
+"""
+function nextreme(ord::Base.Ordering, n::Int, arr::AbstractVector{T}) where T
     if n <= 0
         return T[] # sort(arr)[1:n] returns [] for n <= 0
     elseif n >= length(arr)
-        return sort(arr, lt = (x, y) -> compare(comp, y, x))
+        return sort(arr, order = ord)
     end
 
-    buffer = BinaryHeap{T,Comp}()
+    rev = Base.ReverseOrdering(ord)
 
-    for i = 1 : n
-        @inbounds xi = arr[i]
-        push!(buffer, xi)
-    end
+    buffer = heapify(arr[1:n], rev)
 
     for i = n + 1 : length(arr)
         @inbounds xi = arr[i]
-        if compare(comp, first(buffer), xi)
-            # This could use a pushpop method
-            pop!(buffer)
-            push!(buffer, xi)
+        if Base.lt(rev, buffer[1], xi)
+            buffer[1] = xi
+            percolate_down!(buffer, 1, rev)
         end
     end
 
-    return extract_all_rev!(buffer)
+    return sort!(buffer, order = ord)
 end
 
 """
@@ -128,10 +133,17 @@ end
 
 Return the `n` largest elements of the array `arr`.
 
-Equivalent to `sort(arr, lt = >)[1:min(n, end)]`
+Equivalent to:
+    sort(arr, order = Base.Reverse)[1:min(n, end)]
+
+Note that if `arr` contains floats and is free of NaN values,
+then the following alternative may be used to achieve 2x performance.
+    DataStructures.nextreme(DataStructures.FasterReverse(), n, arr)
+This faster version is equivalent to:
+    sort(arr, lt = >)[1:min(n, end)]
 """
-function nlargest(n::Int, arr::AbstractVector{T}) where T
-    return nextreme(LessThan(), n, arr)
+function nlargest(n::Int, arr::AbstractVector)
+    return nextreme(Base.Reverse, n, arr)
 end
 
 """
@@ -139,8 +151,15 @@ end
 
 Return the `n` smallest elements of the array `arr`.
 
-Equivalent to `sort(arr, lt = <)[1:min(n, end)]`
+Equivalent to:
+    sort(arr, order = Base.Forward)[1:min(n, end)]
+
+Note that if `arr` contains floats and is free of NaN values,
+then the following alternative may be used to achieve 2x performance.
+    DataStructures.nextreme(DataStructures.FasterForward(), n, arr)
+This faster version is equivalent to:
+    sort(arr, lt = <)[1:min(n, end)]
 """
-function nsmallest(n::Int, arr::AbstractVector{T}) where T
-    return nextreme(GreaterThan(), n, arr)
+function nsmallest(n::Int, arr::AbstractVector)
+    return nextreme(Base.Forward, n, arr)
 end
--- a/src/heaps/binary_heap.jl
+++ b/src/heaps/binary_heap.jl
@@ -1,100 +1,6 @@
 # Binary heap (non-mutable)
 
-#################################################
-#
-#   core implementation
-#
-#################################################
-
-function _heap_bubble_up!(comp::Comp, valtree::Array{T}, i::Int) where {Comp,T}
-    i0::Int = i
-    @inbounds v = valtree[i]
-
-    while i > 1  # nd is not root
-        p = i >> 1
-        @inbounds vp = valtree[p]
-
-        if compare(comp, v, vp)
-            # move parent downward
-            @inbounds valtree[i] = vp
-            i = p
-        else
-            break
-        end
-    end
-
-    if i != i0
-        @inbounds valtree[i] = v
-    end
-end
-
-function _heap_bubble_down!(comp::Comp, valtree::Array{T}, i::Int) where {Comp,T}
-    @inbounds v::T = valtree[i]
-    swapped = true
-    n = length(valtree)
-    last_parent = n >> 1
-
-    while swapped && i <= last_parent
-        lc = i << 1
-        if lc < n   # contains both left and right children
-            rc = lc + 1
-            @inbounds lv = valtree[lc]
-            @inbounds rv = valtree[rc]
-            if compare(comp, rv, lv)
-                if compare(comp, rv, v)
-                    @inbounds valtree[i] = rv
-                    i = rc
-                else
-                    swapped = false
-                end
-            else
-                if compare(comp, lv, v)
-                    @inbounds valtree[i] = lv
-                    i = lc
-                else
-                    swapped = false
-                end
-            end
-        else        # contains only left child
-            @inbounds lv = valtree[lc]
-            if compare(comp, lv, v)
-                @inbounds valtree[i] = lv
-                i = lc
-            else
-                swapped = false
-            end
-        end
-    end
-
-    valtree[i] = v
-end
-
-
-function _binary_heap_pop!(comp::Comp, valtree::Array{T}) where {Comp,T}
-    # extract root
-    v = valtree[1]
-
-    if length(valtree) == 1
-        empty!(valtree)
-    else
-        valtree[1] = pop!(valtree)
-        if length(valtree) > 1
-            _heap_bubble_down!(comp, valtree, 1)
-        end
-    end
-    return v
-end
-
-
-function _make_binary_heap(comp::Comp, ty::Type{T}, xs) where {Comp,T}
-    n = length(xs)
-    valtree = copy(xs)
-    for i = 2 : n
-        _heap_bubble_up!(comp, valtree, i)
-    end
-    return valtree
-end
-
+include("arrays_as_heaps.jl")
 
 #################################################
 #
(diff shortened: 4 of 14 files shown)


milesfrain:
Here are the [expected benchmark results](https://travis-ci.org/milesfrain/DataStructures.jl/jobs/602606833#L2031) once the first 2 PRs are merged. There are occasional regressions for very small heaps (n = 10), but I suspect this is mostly due to Travis noise.

Here are the local results:
```
                                                             ID  time ratio memory ratio
––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––– ––––––––––– ––––––––––––
                    ["SparseIntSet", "iterate two exclude one"] 0.88 (5%) ✅    1.00 (1%)
       ["heap", "BinaryHeap", "make", "Float64", "10^1", "Min"] 0.39 (5%) ✅    1.00 (1%)
       ["heap", "BinaryHeap", "make", "Float64", "10^3", "Min"] 0.69 (5%) ✅    1.00 (1%)
         ["heap", "BinaryHeap", "make", "Int64", "10^1", "Min"] 0.26 (5%) ✅    1.00 (1%)
         ["heap", "BinaryHeap", "make", "Int64", "10^3", "Min"] 0.59 (5%) ✅    1.00 (1%)
        ["heap", "BinaryHeap", "pop", "Float64", "10^1", "Min"] 0.94 (5%) ✅    1.00 (1%)
          ["heap", "BinaryHeap", "pop", "Int64", "10^1", "Min"] 0.89 (5%) ✅    1.00 (1%)
       ["heap", "BinaryHeap", "push", "Float64", "10^1", "Min"] 0.93 (5%) ✅    1.00 (1%)
       ["heap", "BinaryHeap", "push", "Float64", "10^3", "Min"] 0.89 (5%) ✅    1.00 (1%)
         ["heap", "BinaryHeap", "push", "Int64", "10^1", "Min"] 0.94 (5%) ✅    1.00 (1%)
         ["heap", "BinaryHeap", "push", "Int64", "10^3", "Min"] 0.86 (5%) ✅    1.00 (1%)
           ["heap", "BinaryMaxHeap", "make", "Float64", "10^1"] 1.14 (5%) ❌    1.00 (1%)
           ["heap", "BinaryMaxHeap", "make", "Float64", "10^3"] 0.73 (5%) ✅    1.00 (1%)
            ["heap", "BinaryMaxHeap", "pop", "Float64", "10^1"] 0.92 (5%) ✅    1.00 (1%)
            ["heap", "BinaryMaxHeap", "pop", "Float64", "10^3"] 0.94 (5%) ✅    1.00 (1%)
           ["heap", "BinaryMaxHeap", "push", "Float64", "10^1"] 0.90 (5%) ✅    1.00 (1%)
           ["heap", "BinaryMaxHeap", "push", "Float64", "10^3"] 0.90 (5%) ✅    1.00 (1%)
           ["heap", "BinaryMinHeap", "make", "Float64", "10^1"] 1.09 (5%) ❌    1.00 (1%)
           ["heap", "BinaryMinHeap", "make", "Float64", "10^3"] 0.72 (5%) ✅    1.00 (1%)
            ["heap", "BinaryMinHeap", "pop", "Float64", "10^1"] 0.87 (5%) ✅    1.00 (1%)
            ["heap", "BinaryMinHeap", "pop", "Float64", "10^3"] 0.95 (5%) ✅    1.00 (1%)
           ["heap", "BinaryMinHeap", "push", "Float64", "10^3"] 0.91 (5%) ✅    1.00 (1%)
["heap", "MutableBinaryHeap", "make", "Float64", "10^1", "Min"] 0.44 (5%) ✅    1.00 (1%)
  ["heap", "MutableBinaryHeap", "make", "Int64", "10^1", "Min"] 0.32 (5%) ✅    1.00 (1%)
                 ["heap", "nlargest", "a=rand(10^4)", "n=10^2"] 0.49 (5%) ✅  0.57 (1%) ✅
                ["heap", "nsmallest", "a=rand(10^4)", "n=10^2"] 0.59 (5%) ✅  0.57 (1%) ✅
```

oxinabox:
> @oxinabox This change introduces API changes, and deprecation warnings should be added. Could we replace the previous deprecation warnings with a version bump?


Here is the rule for deprecations:
1. Deprecations should be added in nonbreaking releases
2. Deprecations should be removed in breaking releases.


So any time we bump the leading nonzero we can (and really should) remove all deprecations.

milesfrain:
@oxinabox I'm having trouble setting-up deprecation warnings, and could use some guidance.

Here's an example of the API change:
``` julia
BinaryMinHeap{Int}() # old
BinaryMinHeap(Int)   # new
```

I was hoping something like this would work:
``` julia
@deprecate BinaryMinHeap{T}() where {T} BinaryMinHeap(T)
```

src/heaps/binary_heap.jl
```diff
@@ -1,124 +1,41 @@
...
-function _make_binary_heap(comp::Comp, ty::Type{T}, xs) where {Comp,T}
-    n = length(xs)
-    valtree = copy(xs)
-    for i = 2 : n
-        _heap_bubble_up!(comp, valtree, i)
-    end
-    valtree
-end
-
+include("arrays_as_heaps.jl")
```
oxinabox: Why did this move?

src/heaps/binary_heap.jl
```diff
@@ -1,124 +1,41 @@
...
 #
 #   heap type and constructors
 #
 #################################################
 
-mutable struct BinaryHeap{T,Comp} <: AbstractHeap{T}
-    comparer::Comp
+#=
+These structs may be substituted by Base.Forward and Base.Reverse,
+but float comparison will be 2x slower to preserve ordering with NAN values.
```
oxinabox:
```suggestion
but float comparison will be 2x slower to preserve ordering with NaN values.
```

src/heaps/binary_heap.jl
```diff
@@ -1,124 +1,41 @@
...
+Base.lt(o::FasterForward, a, b) = a < b
+Base.lt(o::FasterReverse, a, b) = a > b
+
+mutable struct BinaryHeap{T, O <: Base.Ordering} <: AbstractHeap{T}
     valtree::Vector{T}
+    ordering::O
 
-    BinaryHeap{T,Comp}() where {T,Comp} = new{T,Comp}(Comp(), Vector{T}())
+    # min heap by default
+    function BinaryHeap(::Type{T}, ordering::O = FasterForward()) where {T,O}
```
oxinabox: Are you sure this needs to be an inner constructor?
Looks like it can be achieved by calling the existing inner constructor

src/heaps/binary_heap.jl
```diff
@@ -1,124 +1,41 @@
...
 end
 
-const BinaryMinHeap{T} = BinaryHeap{T, LessThan}
-const BinaryMaxHeap{T} = BinaryHeap{T, GreaterThan}
-
-BinaryMinHeap(xs::AbstractVector{T}) where T = BinaryMinHeap{T}(xs)
-BinaryMaxHeap(xs::AbstractVector{T}) where T = BinaryMaxHeap{T}(xs)
+BinaryMinHeap(xs::AbstractVector) = BinaryHeap(xs, FasterForward())
+BinaryMaxHeap(xs::AbstractVector) = BinaryHeap(xs, FasterReverse())
+BinaryMinHeap(::Type{T}) where T = BinaryHeap(T, FasterForward())
```
oxinabox: Do we need to take the type as an argument?
I believe we can take it as a type-parameter as in:
```
julia> struct Foo{T}
       x::Vector{T}
       end

julia> Foo{T}(n::Int) where T = Foo{T}(ones(n))

julia> Foo{UInt8}(10)
Foo{UInt8}(UInt8[0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01])
```

so we can just do:
```suggestion
BinaryMinHeap{T}() where T = BinaryHeap(T, FasterForward())
```

Except we should probably replace that one too so it would be:
```suggestion
BinaryMinHeap{T}() where T = BinaryHeap{T}(FasterForward())
```

src/heaps/minmax_heap.jl
```diff
@@ -7,9 +7,9 @@
 mutable struct BinaryMinMaxHeap{T} <: AbstractMinMaxHeap{T}
     valtree::Vector{T}
 
-    BinaryMinMaxHeap{T}() where {T} = new{T}(Vector{T}())
+    BinaryMinMaxHeap(::Type{T}) where T = new{T}(Vector{T}())
```
oxinabox: above comments also apply here

src/heaps/mutable_binary_heap.jl
```diff
@@ -154,34 +154,33 @@ end
...
-MutableBinaryMaxHeap(xs::AbstractVector{T}) where T = MutableBinaryMaxHeap{T}(xs)
+MutableBinaryMinHeap(xs::AbstractVector) = MutableBinaryHeap(xs, FasterForward())
+MutableBinaryMaxHeap(xs::AbstractVector) = MutableBinaryHeap(xs, FasterReverse())
+MutableBinaryMinHeap(::Type{VT}) where VT = MutableBinaryHeap(VT, FasterForward())
+MutableBinaryMaxHeap(::Type{VT}) where VT = MutableBinaryHeap(VT, FasterReverse())
 
 # deprecated constructors
 
-@deprecate mutable_binary_minheap(::Type{T}) where {T} MutableBinaryMinHeap{T}()
+@deprecate mutable_binary_minheap(::Type{T}) where {T} MutableBinaryMinHeap(T)
```
oxinabox: all these deprecations should be removed in `v"0.18.0"` before/during this PR.

src/heaps/binary_heap.jl
dillondaudert: Not sure I see the reasoning behind changing the inner constructor this way, the current way of instantiating empty heaps seems more idiomatic.

oxinabox:
> @oxinabox I'm having trouble setting-up deprecation warnings, and could use some guidance.

I think you need `@deprecate_binding` maybe.
But lets hold off on deciding if we want to do that in the first place.

src/heaps.jl
```diff
@@ -55,28 +55,22 @@ abstract type AbstractMutableHeap{VT,HT} <: AbstractHeap{VT} end
...
-end
-
-compare(c::LessThan, x, y) = x < y
-compare(c::GreaterThan, x, y) = x > y
-
 # heap implementations
 
 include("heaps/binary_heap.jl")
 include("heaps/mutable_binary_heap.jl")
-include("heaps/arrays_as_heaps.jl")
```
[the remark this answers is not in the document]
milesfrain: Moved to `heaps/binary_heap.jl`, since it is only used there.
Are these functions part of the official API? If so, I will undo moving this and add documentation.

src/heaps.jl
```diff
@@ -55,28 +55,22 @@ abstract type AbstractMutableHeap{VT,HT} <: AbstractHeap{VT} end
...
 include("heaps/minmax_heap.jl")
 
 # generic functions
 
 Base.eltype(::Type{<:AbstractHeap{T}}) where T = T
 
+#=
+Note that extract_all and extract_all_rev are slower than
+sorting the array of values in-place.
+Leaving these function here for use in testing.
```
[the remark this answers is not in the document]
milesfrain: Yes, I'll move these to tests.

src/heaps/binary_heap.jl
milesfrain: `arrays_as_heaps.jl` is only used by `binary_heap.jl`.
If these functions are part of the official API, I will undo moving this and add documentation.

src/heaps/binary_heap.jl
milesfrain: Will include this fix with next revision.

src/heaps/binary_heap.jl
milesfrain: My concern with the existing inner constructors are that they allow specifying redundant (and potentially mismatching) element types. This is described more in the next comment.

src/heaps/binary_heap.jl
milesfrain: If we allow a type parameter and input vector, then mismatched types will trigger an implicit conversion and performance penalty. I think it's better to force explicit conversion by only allowing either type or vector. I also think it's nicer for each of these pieces of information (type or vector) to occupy the same argument position. Open to other API suggestions though.
```
# Baseline performance with matching types

julia> @btime Foo{Float64}(ones(3)) # ones defaults to Float64
  29.514 ns (2 allocations: 128 bytes)
Foo{Float64}([1.0, 1.0, 1.0])

# Performance penalty to convert mismatched types

julia> @btime Foo{Int}(ones(3))
  59.776 ns (3 allocations: 240 bytes)
Foo{Int64}([1, 1, 1])

# Explicit conversion performance is no worse than implicit conversion, but intent is move obvious.

julia> @btime Foo{Int}(convert(Vector{Int}, ones(3)))
  59.305 ns (3 allocations: 240 bytes)
Foo{Int64}([1, 1, 1])
```

milesfrain:
@oxinabox @dillondaudert 

I was unaware of #472, and don't want to undo previous changes, but types as arguments seems like a simpler way to construct heaps when passing either the `element type` or `vector of elements` and an optional `ordering`. Maybe ordering should be mandatory for BinaryHeap.

Here's an example of the API that I'm proposing:
```
BinaryMinHeap(Int)
BinaryMinHeap(rand(3))

BinaryHeap(rand(3))
BinaryHeap(Int)
BinaryHeap(rand(3), Base.Forward)
BinaryHeap(Int, Base.Forward)
```

Let me know what constructor API you'd like to see instead.

src/heaps/binary_heap.jl
oxinabox: we should be consistent with the API of `Base`.
Base allows this, we should also.
Base made the call that allowing this implict conversion was acceptable.
We should to.


```
julia> Vector{Float64}([1,0])
2-element Array{Float64,1}:
 1.0
 0.0
```


Its <2x overhead and if you are 
constructing new heaps in your critical loop then that is its own problem.

Further we also (again inline with  Base), we can push! the wrong typpe and get conversion. Which I feel is a useful feature.

oxinabox:
> 

Let me know what constructor API you'd like to see instead.

I would like to see the current API,
but using `Base.Forward` etc rather than `DataStructures.GreateThan` etc

dillondaudert:
I agree that the API should match `Base`. 

Also changing floating point numbers to a partial order could result in surprising behavior with `NaN`s, I think that should be considered when weighing the performance trade-off of `isless`.

milesfrain:
@dillondaudert 
That's a good point about unexpected behavior with `NaN`s. I'll change this to use the slower and safer comparison by default, and document how users can substitute the riskier and faster comparison if they want better performance.

@oxinabox 
Here's a [branch](https://github.com/milesfrain/DataStructures.jl/commit/cf5e738b1859d75b4aacc22a7d7595c138b3eae8) which demonstrates a dilemma with replacing `DataStructures.<GreaterThan/LessThan>` with `Base.<Forward/Reverse>`.

A `ForwardOrdering` instance can easily be created from its type, but it's not straightforward to do this for `ReverseOrdering`. It seems that the instances `Base.Forward` and `Base.Reverse` should be passed into the constructors directly, but I don't think it is possible to tack on an extra ordering function parameter to `BinaryHeap` using the existing `const Binary<Min/Max>Heap{T}` type alias strategy, which leads me to believe that handling `Base.Ordering` requires undoing some of #472 to make `Binary<Min/Max>Heap` behave as a pseudoconstructor(?) and not as a parametric composite type constructor alias. Is it problematic for heap constructors to behave more like the array-constructing `zeros(Int,3)` and `rand(Int,3)`, rather than `Vector{Int}(undef,3)`?

Here's a log showing the difficulties of constructing a `ReverseOrdering` instance from its type.
```
julia> Base.Forward
Base.Order.ForwardOrdering()

julia> Base.ForwardOrdering()
Base.Order.ForwardOrdering()


julia> Base.Reverse
Base.Order.ReverseOrdering{Base.Order.ForwardOrdering}(Base.Order.ForwardOrdering())

julia> Base.ReverseOrdering()
ERROR: MethodError: no method matching Base.Order.ReverseOrdering()

julia> Base.Order.ReverseOrdering{Base.Order.ForwardOrdering}()
ERROR: MethodError: no method matching Base.Order.ReverseOrdering{Base.Order.ForwardOrdering}()
```

oxinabox:
Yeah, fair call.
To be clear, I think it is fine for `BinaryHeap` to take an ordering object instance, (it already does),
since then can have `By` orderings etc.
It isn't fine for it to take a `Type{Int}` etc argument.
As that puts us out of aligment for conventions for data structures.

We just need to add:
`Base.Order.ReverseOrdering() = Base.Order.Reverse`
like you say to fix this.
(I tested this on a local copy of your ord4 branch)
I will open a PR to Julia itself now to add that missing definition, and see how that goes.
Then if it is accepted we can add a compatability definition for it.

oxinabox:
@milesfrain  sorry for the delay, was waiting to see if anyone upstream in JuliaLang objected to the addition   of `Reverse()`

They did not.
So we can now proceed as if it were a thing.

I will have it added to Compat.jl
and then we can just load Compat.jl and it will be a thing in all julia versions

milesfrain:
@oxinabox No worries. I was also waiting to see what happens with the `Reverse()` PR. I'll continue working through the requested changes.

oxinabox:
Ok, @milesfrain  you can now add a Compat v3 dependency, and start using the `ReverseOrdering()` constructor

src/heaps/binary_heap.jl
milesfrain: Are there any concerns with moving this `include`? I figure it's best to move it to the smallest scope where it is used.

milesfrain:
@oxinabox

I believe this latest changeset address all of your latest review feedback, but let me know if I missed anything, or if you have additional suggestions.

I need some help troubleshooting why Travis is failing with Compat V3 and Julia 0.7.

Local benchmark results:
```
                                                             ID  time ratio memory ratio
––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––– ––––––––––– ––––––––––––
                        ["SparseIntSet", "pop push worst case"] 0.93 (5%) ✅    1.00 (1%)
       ["heap", "BinaryHeap", "make", "Float64", "10^1", "Min"] 1.14 (5%) ❌    1.00 (1%)
       ["heap", "BinaryHeap", "make", "Float64", "10^3", "Min"] 0.68 (5%) ✅    1.00 (1%)
         ["heap", "BinaryHeap", "make", "Int64", "10^3", "Min"] 0.62 (5%) ✅    1.00 (1%)
        ["heap", "BinaryHeap", "pop", "Float64", "10^1", "Min"] 0.91 (5%) ✅    1.00 (1%)
        ["heap", "BinaryHeap", "pop", "Float64", "10^3", "Min"] 0.94 (5%) ✅    1.00 (1%)
          ["heap", "BinaryHeap", "pop", "Int64", "10^1", "Min"] 0.84 (5%) ✅    1.00 (1%)
          ["heap", "BinaryHeap", "pop", "Int64", "10^3", "Min"] 0.92 (5%) ✅    1.00 (1%)
       ["heap", "BinaryHeap", "push", "Float64", "10^1", "Min"] 0.94 (5%) ✅    1.00 (1%)
       ["heap", "BinaryHeap", "push", "Float64", "10^3", "Min"] 0.93 (5%) ✅    1.00 (1%)
         ["heap", "BinaryHeap", "push", "Int64", "10^1", "Min"] 0.95 (5%) ✅    1.00 (1%)
         ["heap", "BinaryHeap", "push", "Int64", "10^3", "Min"] 0.91 (5%) ✅    1.00 (1%)
["heap", "MutableBinaryHeap", "make", "Float64", "10^1", "Min"] 1.13 (5%) ❌    1.00 (1%)
   ["heap", "MutableBinaryHeap", "pop", "Int64", "10^3", "Min"] 1.09 (5%) ❌    1.00 (1%)
                 ["heap", "nlargest", "a=rand(10^4)", "n=10^2"] 0.44 (5%) ✅  0.57 (1%) ✅
                ["heap", "nsmallest", "a=rand(10^4)", "n=10^2"] 0.44 (5%) ✅  0.57 (1%) ✅
```

oxinabox:
We should just drop support for Julia 0.7.
For now can just delete it from Travis

milesfrain:
@oxinabox 
I'm ready for a re-review.

Regarding the CI errors:
* There's an incompatibility with nightly Julia for Travis and AppVeyor. It's passing on stable Julia. 
* Coveralls reports a lack of coverage for `FasterForward`, even though it's [exercised in tests](https://github.com/JuliaCollections/DataStructures.jl/pull/547/commits/17c1578cd0d3e7e98c1929620bc557d1225dbcb6#diff-754314a336d6f20bcef49fa9ff0643eaR160).

See the "construct heap" tests for examples of all the available heap constructors. Both immutable and mutable heaps match the requested API discussed earlier, but MinMax heap is missing a way to specify an element type along with an array of elements (e.g. `BinaryMinMaxHeap{Int}(vs)`). Should MinMax heap be changed as part of this PR? It is untouched otherwise.

dillondaudert:
I don't see any issue with adding `BinaryMinMaxHeap{T}(xs::AbstractVector{T}) where T` as a constructor.

oxinabox:
Nightly is failing. because  Julia nightly is I think a bit borked.
 we should allow failures on nightly.

appveyor.yml
```diff
@@ -1,6 +1,6 @@
 environment:
   matrix:
-  - julia_version: 0.7
+  - julia_version: 1.3
```
oxinabox:
```suggestion
  - julia_version: 1.0
  - julia_version: 1.3
```

need to test the LTS

src/heaps.jl
```diff
@@ -55,28 +55,24 @@ abstract type AbstractMutableHeap{VT,HT} <: AbstractHeap{VT} end
...
 include("heaps/minmax_heap.jl")
 
 # generic functions
 
 Base.eltype(::Type{<:AbstractHeap{T}}) where T = T
 
+"""
+    extract_all!(h)
+
+returns an array of heap elements in sorted order (heap head at first index).
```
oxinabox:
```suggestion
Return an array of heap elements in sorted order (heap head at first index).
```

and below

src/heaps/binary_heap.jl
```diff
@@ -1,153 +1,101 @@
...
 #################################################
 #
 #   heap type and constructors
 #
 #################################################
 
-mutable struct BinaryHeap{T,Comp} <: AbstractHeap{T}
-    comparer::Comp
+"""
+FasterForward enables 2x faster float comparison versus Base.ForwardOrdering,
```
oxinabox:
```suggestion
    FasterForward()

FasterForward enables 2x faster float comparison versus Base.ForwardOrdering,
```

src/heaps/binary_heap.jl
```diff
@@ -1,153 +1,101 @@
...
+but ordering is undefined if the data contains NaN values.
+Enable this higher-performance option by calling the BinaryHeap
+constructor instead of the BinaryMinHeap helper constructor.
+"""
+struct FasterForward <: Base.Ordering end
+Base.lt(o::FasterForward, a, b) = a < b
+
+"""
+FasterReverse enables 2x faster float comparison versus Base.ReverseOrdering,
```
oxinabox:
```suggestion
    FasterReverse()

FasterReverse enables 2x faster float comparison versus Base.ReverseOrdering,
```

src/heaps/binary_heap.jl
oxinabox: Are we now inline with `Base` ?

benchmark/bench_heap.jl
```diff
@@ -66,38 +66,22 @@ for heap in heaptypes
...
-                @benchmarkable push_heap(h, $a) setup=(h=$(heap){$dt}())
-            suite[vcat(prepath, ["pop"], postpath)] =
-                @benchmarkable pop_heap(h) setup=(h=$(heap)($a))
-        end
-    end
-end
+fast_extreme_orderings = Dict(
+    nsmallest => DataStructures.FasterForward(),
+    nlargest => DataStructures.FasterReverse(),
+    )
```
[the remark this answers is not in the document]
milesfrain: Is this a suggestion to change indentation from 4 spaces to 3 spaces? 4 spaces is used everywhere else in the file.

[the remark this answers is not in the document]
oxinabox: I don't know what this comment is about

jzazo:
What's missing for this to be tagged? 
Should I stick with [this](https://github.com/JuliaCollections/DataStructures.jl/pull/529) for the time being? Thanks!

milesfrain:
> What's missing for this to be tagged?

I was just waiting on additional feedback:

> @ararslan @kmsquire @scls19fr
> could one of you give this a final look and bring it home?

But perhaps I was being a bit too patient.

> Last thing is deprecations.
> SemVer does not require us to do deprecations, if we are tagging this as a breaking change.
> Which I think we should because I don't think we can easily depprecate it fully.
> Maybe we can (if so that would be great)

I was hoping we could just make this a breaking change, rather than add partial deprecations. I don't think everything in this new version will translate fully via deprecations.

oxinabox:
Ok, I have taken another look.
this still looks good to me.
Rebase it, set the version to `0.18.0-DEV`
and I will look to making a breaking release this weekend with this change.

src/heaps/binary_heap.jl
milesfrain: > Are we now inline with `Base` ?

Implicit conversion is enabled in latest pushed commit: https://github.com/JuliaCollections/DataStructures.jl/pull/547/commits/1eeafc81efea37453b7e068f033b276e8ba66ae9

milesfrain:
This is ready for review.

Latest benchmark results show some minor regressions for the small (10^1 element) heaps. Might just be noise, since some differences are also reported for the untouched `SparseIntSet`.
```
                                                             ID  time ratio memory ratio
––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––– ––––––––––– ––––––––––––
                                ["SparseIntSet", "create_fill"] 0.90 (5%) ✅    1.00 (1%)
                        ["SparseIntSet", "pop push worst case"] 1.11 (5%) ❌    1.00 (1%)
       ["heap", "BinaryHeap", "make", "Float64", "10^1", "Min"] 1.24 (5%) ❌    1.00 (1%)
       ["heap", "BinaryHeap", "make", "Float64", "10^3", "Min"] 0.79 (5%) ✅    1.00 (1%)
         ["heap", "BinaryHeap", "make", "Int64", "10^1", "Min"] 1.16 (5%) ❌    1.00 (1%)
         ["heap", "BinaryHeap", "make", "Int64", "10^3", "Min"] 0.81 (5%) ✅    1.00 (1%)
        ["heap", "BinaryHeap", "pop", "Float64", "10^1", "Min"] 1.42 (5%) ❌    1.00 (1%)
          ["heap", "BinaryHeap", "pop", "Int64", "10^1", "Min"] 1.47 (5%) ❌    1.00 (1%)
          ["heap", "BinaryHeap", "pop", "Int64", "10^3", "Min"] 0.90 (5%) ✅    1.00 (1%)
       ["heap", "BinaryHeap", "push", "Float64", "10^3", "Min"] 0.94 (5%) ✅    1.00 (1%)
         ["heap", "BinaryHeap", "push", "Int64", "10^3", "Min"] 0.95 (5%) ✅    1.00 (1%)
["heap", "MutableBinaryHeap", "make", "Float64", "10^1", "Min"] 1.19 (5%) ❌    1.00 (1%)
["heap", "MutableBinaryHeap", "make", "Float64", "10^3", "Min"] 0.93 (5%) ✅    1.00 (1%)
["heap", "MutableBinaryHeap", "push", "Float64", "10^1", "Min"] 0.94 (5%) ✅    1.00 (1%)
                 ["heap", "nlargest", "a=rand(10^4)", "n=10^2"] 0.71 (5%) ✅  0.57 (1%) ✅
                ["heap", "nsmallest", "a=rand(10^4)", "n=10^2"] 0.51 (5%) ✅  0.57 (1%) ✅
```

milesfrain:
I'm also curious what the best local testing workflow is.
`test/README` suggests running `julia runtests.jl binheap`, but that command seems to be missing something. Best I could come up with is:
```
julia -e 'using Pkg; Pkg.activate("."); include("test/runtests.jl")'
```
With a temporary edit of the `tests` array in `runtests.jl` to just select `binheap`.

oxinabox:
`julia --project="."` then `] test` in the REPL is typical.
`julia --project="." -e "using Pkg; Pkg.test()"`  should work if you don't want to enter the REPL

I think `test/README` is just wrong and very outdated

oxinabox:
Sorry for they delay in this, just hoping to get a few more deprecations in before making the breaking change

jlumpe:
I didn't see this PR and essentially wrote a [duplicate](#646) of it. Linking it here in case you would like to compare the two.

milesfrain:
Bummer about your duplicated effort @jlumpe . I've made that same mistake myself a few times.

oxinabox:
Ok finally we are in a position in master that we can merge this.
Master is now 0.18-DEV.
so if you can rebase this against master and then we will merge.

oxinabox:
oh, i didn't see this had been rebased.
Yay, thank you

PR #548, opened 2019-10-26
CircularBuffer: add resize! method

sairus7:
It is adopted from `Base.resize!`:

    resize!(a::Vector, n::Integer) -> Vector

  Resize `a` to contain `n` elements. If `n` is smaller than the current collection length, the first `n` elements will be retained. If `n` is larger, the new elements are not guaranteed to be initialized.

---
With the following meaning:

    resize!(cb, n)

Resize `CircularBuffer` to the maximum capacity of `n` elements. If `n` is smaller than the current buffer length, the first n elements will be retained.

---
Note that it's not changing the current length of the buffer, but its maximum capacity (length of the underlying Array).


sairus7:
But maybe we should retain last elements, dropping first ones in a similar manner, as `push!` to a full buffer overrides first elements?

oxinabox:
Can you rebase this off current master?

src/circular_buffer.jl
```diff
@@ -216,3 +216,23 @@ Base.@propagate_inbounds function Base.last(cb::CircularBuffer)
     end
     cb.buffer[_buffer_index(cb, cb.length)]
 end
+"""
```
oxinabox:
```suggestion

"""
```

Line breaking

src/circular_buffer.jl
```diff
@@ -216,3 +216,23 @@ Base.@propagate_inbounds function Base.last(cb::CircularBuffer)
...
+    resize!(cb, n)
+
+Resize CircularBuffer to the maximum capacity of n elements. If n is smaller than the current buffer length, the first n elements will be retained.
+"""
+function Base.resize!(cb::CircularBuffer{T}, n::Int) where {T}
+    if n != capacity(cb)
+        buf_new = Vector{T}(undef, n)
+        len_new = min(length(cb), n)
+        for i = 1 : len_new
```
oxinabox:
```suggestion
        for i in 1 : len_new
```

src/circular_buffer.jl
```diff
@@ -216,3 +216,23 @@ Base.@propagate_inbounds function Base.last(cb::CircularBuffer)
...
+            @inbounds buf_new[i] = cb[i]
+        end
+
+        cb.buffer = buf_new
+        cb.first = 1
+        cb.capacity = n
+        cb.length = len_new
+    end
+    cb
```
oxinabox:
```suggestion
    return cb
```

oxinabox:
Some minor stylistic things.

I think for now this is fine.
If it becomes a performance bottleneck for anyone we can look into improving it then

oxinabox:
Can you also bumpp the patch version number so I can tag a release?

sairus7:
> Can you rebase this off current master?

To what branch should I rebase this commit? Or should I change my fork branch? 
Looks like I can change only the target branch:
image

oxinabox:
Seems like no conflict , so no need to rebase.
(You should add a remote for JuliaCollections, and then rebase off that remote's master)

sairus7:
> Now you have done a merge which makes history ugly, but its not the end of the world,
> so you don't need to rebase now.

Will it make ugly history in the `JuliaCollections:master`?
Sorry, I don't quite understand yet the correct workflow. You can point me to, if there is one.

oxinabox:
> Will it make ugly history in the JuliaCollections:master?

Yeah, but it isn't that big a deal.

> Sorry, I don't quite understand yet the correct workflow. You can point me to, if there is one.

Try:
 - https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github
 - https://www.digitalocean.com/community/tutorials/how-to-rebase-and-update-a-pull-request

eulerkochy:
@sairus7 still up on this PR?

rfourquet:
Calling this functionality `resize!` seems misleading to me, as the lenght is left unchanged (so the behavior diverges from `resize!` on a vector). What about making this a method of `sizehint!` ?

sairus7:
Sorry, forgot about it, closing since it's merged.

PR #549, opened 2019-11-09
improved docstring in circular_buffer.jl

### LINKED ISSUE #510: Docstrings should annotate types in the docstrings
Consider the docstrings here https://github.com/JuliaCollections/DataStructures.jl/blob/35295770d216d111036c85fcb7ffe6b619564d81/src/circular_buffer.jl#L158-L193.

These have no type annotation for the input arguments. If one look at the docs for e.g. `size` in the REPL one gets:

```jl
help?> size
search: size sizeof sizehint! Csize_t resize! filesize Cssize_t displaysize @specialize @nospecialize

  size(A::AbstractArray, [dim])

...

  ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  size(cb)

  Return a tuple with the size of the buffer.
```

The lack of any type on `cb` makes this quite confusing because it looks like this method applies to `cb::Any`.

AtsushiSakai:
I improved docstring in circular_buffer.jl to fix #510

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -19,7 +19,7 @@ end
 CircularBuffer(capacity) = CircularBuffer{Any}(capacity)
 
 """
-    empty!(cb)
+    empty!(cb::CircularBuffer)
 
 Reset the buffer.
 """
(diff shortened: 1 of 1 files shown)


oxinabox:
Thanks!

PR #550, opened 2019-11-10
improved docstrings in deque.jl

AtsushiSakai:
I realized the deque.jl is missing some docstrings like #549. So I improved it.


oxinabox:
excellent, thanks

PR #551, opened 2019-11-11
improved docstrings in circ_deque.jl

AtsushiSakai:
I realized the circ_deque.jl is missing some docstrings like #549. So I improved it.

### DIFF
--- a/src/circ_deque.jl
+++ b/src/circ_deque.jl
@@ -13,18 +13,7 @@ end
 
 CircularDeque{T}(n::Int) where {T} = CircularDeque(Vector{T}(undef, n), n, 0, 1, n)
 
-"""
-    length(D::CircularDeque)
-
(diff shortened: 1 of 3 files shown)


oxinabox:
A bunch of these don't need docstrings at all as the docstring in `Base` is fine for them.

```
help?> length
search: length

  length(collection) -> Integer

  Return the number of elements in the collection.
```

AtsushiSakai:
@oxinabox Thank you for your reviewing. Then I removed unnecessary docstrings.

oxinabox:
LGTM, thanks

AtsushiSakai:
@oxinabox By the way. I found other many files are missing docstrings. Would you like me to keep fixing these files? If not, I will stop to make a PR.

oxinabox:
Yeah sure, keep going its great.
I really appreciate it.

Issue #539, opened 2019-10-04
Adopt a style guide


oxinabox:
I am happy with [BlueStyle](https://github.com/invenia/BlueStyle),
or with [YSAGuide](https://github.com/jrevels/YASGuide)

I am willing to make the PR to clean most things up inline with BlueStyle.
But if someone else is willing to make the PR to switch things over to be mostly in line with YASGuide then that is fine also (@ararslan).

ararslan:
I haven't touched this code in a long time so I don't really have an opinion here.

PR #553, opened 2019-11-21
Allow accumulators to take an AbstractDict

nickrobinson251:


src/accumulator.jl
```diff
@@ -7,9 +7,10 @@ end
 ## constructors
 
 Accumulator{T, V}() where {T,V<:Number} = Accumulator{T,V}(Dict{T,V}())
+Accumulator(map::AbstractDict) = Accumulator(Dict(map))
 
 counter(T::Type) = Accumulator{T,Int}()
-counter(dct::Dict{T,V}) where {T,V<:Integer} = Accumulator{T,V}(copy(dct))
+counter(dct::AbstractDict{T,V}) where {T,V<:Integer} = Accumulator{T,V}(Dict(dct))
```
nicoleepp:
```suggestion
counter(dct::AbstractDict{T, V}) where {T, V<:Integer} = Accumulator{T, V}(Dict(dct))
```

test/test_accumulator.jl
```diff
@@ -90,6 +90,16 @@
         @test ctm["b"] == 0
     end
 
+    @testset "From AbstractDict" begin
+        kv = (:a => 1, :b => 2)
+        for f in (Accumulator, counter)
+            ct_odict = f(OrderedDict(kv))
+            ct_dict = f(Dict(kv))
+            @test ct_odict isa Accumulator{Symbol,Int}
```
nicoleepp:
```suggestion
            @test ct_odict isa Accumulator{Symbol, Int}
```

oxinabox:
LGTM.
please bump versiom number so i can tag release

src/accumulator.jl
oxinabox: See also #539

nickrobinson251: i think no space is the dominant style in the file here (except that one case above on L9).

oxinabox: Lets make the change it gets close to the goal of #539

nickrobinson251: let's leave style changes to a different PR (happy to see an issue for it!)

PR #556, opened 2019-11-25
Add missing contents about fenwick tree for the documentation top page.

AtsushiSakai:
The documentation top page is missing contents about fenwick tree.
So I added it.


oxinabox:
thanks!

PR #557, opened 2019-11-26
update docstrings of Accumulator and clean up codes based on its docstrings.

AtsushiSakai:
Update docstrings of Accumulator and clean up codes based on its docstrings.

### DIFF
--- a/src/accumulator.jl
+++ b/src/accumulator.jl
@@ -1,5 +1,10 @@
 #A counter type
 
+"""
+    Accumulator{T, V<:Number}
+
+A accumulator is a data structure that maintains an accumulated number for each key.
+"""
 struct Accumulator{T, V<:Number} <: AbstractDict{T,V}
     map::Dict{T,V}
 end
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
I can change the key "x" to "k".
Do other contributors have any other ideas?

oxinabox:
> not sure about opinion of others contributors... but I personally don't like the use of x for key... maybe this PR could be used to name it k.

I am fine with `x`
accumulators are not exactly dictionaries (I mean technically they are), 
 thinking of it is a key  is a bit funny.

src/accumulator.jl
```diff
@@ -1,5 +1,10 @@
 #A counter type
 
+"""
+    Accumulator{T, V<:Number}
+
+A accumulator is a data structure that maintains an accumulated number for each key.
```
oxinabox:
```suggestion
A accumulator is a data structure that maintains an accumulated total for each key.
The particular case where those totals are integers is a counter.
```

AtsushiSakai:
@oxinabox I fixed it based on your suggestion.

oxinabox:
thanks again for all this.

PR #564, opened 2019-12-24
Binary heap constructor

hiemstar:
Modified internal constructor of BinaryHeap:

    function BinaryHeap{T,Comp}(::Comp, xs::AbstractVector{T}) where {T,Comp}
        valtree = _make_binary_heap(Comp(), T, xs)
        new{T,Comp}(Comp(), valtree)
    end

to

    function BinaryHeap{T,Comp}(comp::Comp, xs::AbstractVector{T}) where {T,Comp}
        valtree = _make_binary_heap(comp, T, xs)
        new{T,Comp}(comp, valtree)
    end

This makes it possible to heapify an array based on data not stored in the array itself.

### DIFF
--- a/src/heaps/binary_heap.jl
+++ b/src/heaps/binary_heap.jl
@@ -101,18 +101,18 @@ end
 #   heap type and constructors
 #
 #################################################
-
 mutable struct BinaryHeap{T,Comp} <: AbstractHeap{T}
     comparer::Comp
     valtree::Vector{T}
 
     BinaryHeap{T,Comp}() where {T,Comp} = new{T,Comp}(Comp(), Vector{T}())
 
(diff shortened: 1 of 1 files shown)


oxinabox:
@dillondaudert can you review?

dillondaudert:
Hi, thanks for the PR. Could you clarify with examples what exactly this change allows?

You'll also want to make sure that all the heaps (including the MinMax heap) follow the same API and have unit tests to cover any changed functionality.

jakobnissen:
@hiemstar It appears to me that what this PR does is allows you to instantiate a `BinaryMinHeap` that has the max heap property and vice versa. To me, this would break the expected behaviour of the heap. Can you clarify what you intend this PR to do?

Issue #562, opened 2019-12-22
Missing constructor for DisjointSets


dpsanders:
```jl
julia> DisjointSets([1, 2, 3])
ERROR: MethodError: no method matching DisjointSets(::Array{Int64,1})
Stacktrace:
 [1] top-level scope at REPL[2]:1

julia> DisjointSets{Int}([1, 2, 3])
DisjointSets{Int64}(Dict(2 => 2,3 => 3,1 => 1), [1, 2, 3], IntDisjointSets([1, 2, 3], [0, 0, 0], 3))
```
There doesn't seem to be any reason not to have a constructor 
```jl
DataStructures(v::Vector{T}) where {T} = DisjointSets{T}(v)
```

oxinabox:
Yes, we should have that.

Miore generally one for iterators.

dpsanders:
I think it would also be useful to have a function that extracts the grouped sets at the end (similar to connected components of a graph). I wrote an implementation for my use case but it feels like there should be an efficient algorithm for that.

oxinabox:
Probably worth making a seperate issue

dpsanders:
Good idea: #563

PR #567, opened 2020-01-11
Add additional constructors for DisjointSets

c-p-murphy:
Add additional `DisjointSets` constructors to address: #562.

### DIFF
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -128,17 +128,21 @@ mutable struct DisjointSets{T}
 end
 
 DisjointSets() = DisjointSets{Any}()
-DisjointSets(T::Type) = DisjointSets{T}()
-DisjointSets(xs::T) where T = DisjointSets{eltype(T)}(xs)
+DisjointSets(::Type{T}) where T = DisjointSets{T}()
 DisjointSets(xs::T...) where T = DisjointSets{T}(xs)
 DisjointSets{T}(xs::T...) where T = DisjointSets{T}(xs)
-function DisjointSets(xs::T) where {T<:Base.Generator}
-    return DisjointSets{Base.@default_eltype(xs)}(xs)
+DisjointSets(xs) = _DisjointSets(xs, Base.IteratorEltype(xs))
+_DisjointSets(xs, ::Base.HasEltype) = DisjointSets{eltype(xs)}(xs)
+function _DisjointSets(xs, ::Base.EltypeUnknown)
+    T = Base.@default_eltype(xs)
+    (isconcretetype(T) || T === Union{}) || return grow_to!(DisjointSets{T}(), xs)
+    return DisjointSets{T}(xs)
 end
 
 length(s::DisjointSets) = length(s.internal)
 num_groups(s::DisjointSets) = num_groups(s.internal)
 Base.eltype(::Type{DisjointSets{T}}) where T = T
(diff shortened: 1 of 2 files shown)


src/disjoint_set.jl
```diff
@@ -126,6 +127,15 @@ mutable struct DisjointSets{T}
     end
 end
 
+DisjointSets() = DisjointSets{Any}()
+DisjointSets(T::Type) = DisjointSets{T}()
```
oxinabox: Don't need this one.
These should all have been removed. this was a julia 0.3ish thing.
Before type-parameters were allowed in general circumstancess.

```suggestion
```

src/disjoint_set.jl
```diff
@@ -126,6 +127,15 @@ mutable struct DisjointSets{T}
...
+DisjointSets(xs::T) where T = DisjointSets{eltype(T)}(xs)
+DisjointSets(xs::T...) where T = DisjointSets{T}(xs)
+DisjointSets{T}(xs::T...) where T = DisjointSets{T}(xs)
+function DisjointSets(xs::T) where {T<:Base.Generator}
+    return DisjointSets{Base.@default_eltype(xs)}(xs)
+end
```
oxinabox: We should do this the same way Base does.

```
Set(itr) = _Set(itr, IteratorEltype(itr))

_Set(itr, ::HasEltype) = Set{eltype(itr)}(itr)

function _Set(itr, ::EltypeUnknown)
    T = @default_eltype(itr)
    (isconcretetype(T) || T === Union{}) || return grow_to!(Set{T}(), itr)
    return Set{T}(itr)
end
```

c-p-murphy:
Thanks for taking the time to review this—I think it should be all set now, but feel free to let me know if I misunderstood any of your suggestions or if you notice anything else that should be modified.

src/disjoint_set.jl
```diff
@@ -126,9 +127,22 @@ mutable struct DisjointSets{T}
     end
 end
 
+DisjointSets() = DisjointSets{Any}()
+DisjointSets(::Type{T}) where T = DisjointSets{T}()
```
oxinabox: This still needs to go.
Passing type in as an arg, rather than a type-param is a thing we used to have to do in Julia 0.3,
but not anymore.

```
julia> Set(Int)
ERROR: MethodError: no method matching length(::Type{Int64})
```

src/disjoint_set.jl
```diff
@@ -126,9 +127,22 @@ mutable struct DisjointSets{T}
...
+function _DisjointSets(xs, ::Base.EltypeUnknown)
+    T = Base.@default_eltype(xs)
+    (isconcretetype(T) || T === Union{}) || return grow_to!(DisjointSets{T}(), xs)
+    return DisjointSets{T}(xs)
+end
+
 length(s::DisjointSets) = length(s.internal)
 num_groups(s::DisjointSets) = num_groups(s.internal)
 Base.eltype(::Type{DisjointSets{T}}) where T = T
+empty(s::DisjointSets{T}, ::Type{U}=T) where {T,U} = DisjointSets{U}()
```
oxinabox: this needs a test

src/disjoint_set.jl
c-p-murphy: Ok, thanks for clarifying.

oxinabox:
Thanks.
Releasing now

PR #568, opened 2020-01-19
Fix DisjointSets constructor error

c-p-murphy:
Address error introduced by #567 in situations like the following:

```julia
DisjointSets(x % 2 == 0 ? x+1//x : x*im for x = 1:10)
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.9"
+version = "0.17.10"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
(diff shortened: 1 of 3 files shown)


c-p-murphy:
Just realized this actually still needs work, so don't worry about reviewing or merging it yet.

oxinabox:
Definately add that test case . I should have checked coverage before merging

c-p-murphy:
Everything seems to be working as expected now, but I'm not entirely confident that this is the best approach to solving the problem. Both this and the Set constructor rely on [these lines](https://github.com/JuliaLang/julia/blob/2d5741174ce3e6a394010d2e470e4269ca54607f/base/array.jl#L681-L716) in arrays.jl, which seems a bit counterintuitive to me, but maybe I'm missing something. If it doesn't seem like a problem to you though, I'm happy to have this merged and then see about opening a new PR down the road if a more elegant solution becomes apparent. 

Either way, thanks for taking the time to review this.

Issue #291, opened 2017-06-02
SortedSet should be a subtype of Base.AbstractSet


StephenVavasis:
I think SortedSet should be a subtype of Base.AbstractSet.  I'm not sure if there is an 'official' description of API for AbstractSet, but SortedSet is implements most of the important functions that Set implements.

NHDaly:
+1; this surprised me today.

PR #571, opened 2020-02-04
Make `SortedSet{K}` be a subtype of `AbstractSet{K}`.

NHDaly:
Have `SortedSet{K}` implement <: `AbstractSet{K}`. Fixes https://github.com/JuliaCollections/DataStructures.jl/issues/291.

Are there other things one needs to do to complete this? 
https://docs.julialang.org/en/v1/base/collections/

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.10"
+version = "0.18.0"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.18.0"
+version = "0.17.13"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
--- a/src/sorted_set.jl
+++ b/src/sorted_set.jl
@@ -15,7 +15,7 @@ Construct a SortedSet using keys given by iterable `iter` (e.g., an
 array) and ordering object `o`. The ordering object defaults to
 `Forward` if not specified.
 """
-mutable struct SortedSet{K, Ord <: Ordering}
+mutable struct SortedSet{K, Ord <: Ordering} <: AbstractSet{K}
     bt::BalancedTree23{K,Nothing,Ord}
 
     function SortedSet{K,Ord}(o::Ord=Forward, iter=[]) where {K,Ord<:Ordering}


oxinabox:
Can you check this against:
https://github.com/JuliaLang/julia/issues/34677

oxinabox:
@NHDaly  bump

NHDaly:
@oxinabox Ooops, i hadn't seen your response! Thanks for the bump.

Here's my notes (though note that i've never actually used this package before, i came upon this tangentially in unrelated code):

------------------------------------------------------------------------
#### ALL
- [x] iterator interface.
- [x] `in`  (may not actually be required by any methods)
```julia
julia> s = SortedSet{Int}([2,3,4])
SortedSet([2, 3, 4],
Base.Order.ForwardOrdering())

julia> 2 in s
true

julia> collect(s)
3-element Array{Int64,1}:
 2
 3
 4
```

#### Optional
- [ ] `Base.empty(myset::MySet, ::Type{U}=eltype(myset) where U`: returns a empty set able to hold elements of type `U`. (defaults to `Set{U}`). Should over load to make a empty copy of ones own set. (cf #34676)
- [x] `Base.hasfastin(::Type{<:MySet})=false` if not checkign `in` is not logarithmic or better time

The empty + type overload is missing:
```julia
julia> empty(s)
SortedSet(Int64[],
Base.Order.ForwardOrdering())

julia> empty(s, Float32)
Set(Float32[])
```
```julia
julia> Base.hasfastin(typeof(s))
true
```

### If immutable:
#### Required:
- [ ] `copy(::MySet)`

No copy (but not immutable):
```julia
julia> copy(s)
ERROR: MethodError: no method matching copy(::SortedSet{Int64,Base.Order.ForwardOrdering})

julia> isimmutable(typeof(s))
false
```

#### Strongly Recommended.
Else will fall back to returning `Set` / result of `emptymutable`/`copymutable`.

- [ ] `setdiff(::Myset, itr)`:
- [ ] `intersect(::Myset, itr)`
- [x] `union(::MySet), itr)`
- [x] `empty`

The first two are all available but only if the RHS is also a `SortedSet`. This should probably also have the generic overload:
```julia
julia> setdiff(s, [2,3])
Set([4])

julia> setdiff(s, SortedSet([2,3]))
SortedSet([4],
Base.Order.ForwardOrdering())

julia> intersect(s, [1,2])
Set([2])

julia> intersect(s, SortedSet([1,2]))
SortedSet([2],
Base.Order.ForwardOrdering())
```

Interestingly, `union` seems okay:
```julia
julia> union(s, [1,2])
SortedSet([1, 2, 3, 4],
Base.Order.ForwardOrdering())

julia> union(s, SortedSet([1,2]))
SortedSet([1, 2, 3, 4],
Base.Order.ForwardOrdering())

julia> union(s, Set([1,2]))
SortedSet([1, 2, 3, 4],
Base.Order.ForwardOrdering())

julia> union(Set([1,2]), s)
Set([4, 2, 3, 1])
```

### If mutable:
Optionally anything required or recommended for immutable can be implemented for mutable.

#### Required
- [x] `empty!(::MySet)`,
- [x] `delete!(::MySet, x)`
- [x] `push!(::MySet, x)`

`empty!` is almost working, but it's returning `nothing` instead of the now-empty set.
```julia
julia> empty!(Set([2,3]))
Set(Int64[])

julia> empty!(s)

julia> s
SortedSet(Int64[],
Base.Order.ForwardOrdering())
```


#### Strongly recommended:
- [ ] `Base.emptymutable(myset::MySet, ::Type{U}=eltype(myset) where U`: returns a empty mutable set able to hold elements of type `U`. (defaults to `Set{U}`)
- [ ] `Base.copymutable(::MySet)` produces a mutable copy of the set (defaults to `Set`)
- [ ] `copy` as alias for `copymutable`

None of these are implemented. It defaults to `Set`.

#### Optional
- [ ] `sizehint!(::Myset, n)`

I don't see think this is implemented:
```julia
julia> @which sizehint!(s, 100)
sizehint!(s::AbstractSet, n) in Base at abstractset.jl:4
```

----------------------------

So i think it hits all the required interfaces, and is missing some of the Strongly Recommendeds.

NHDaly:
Update: i've just opened https://github.com/JuliaCollections/DataStructures.jl/pull/584 which adds a few more of the missing functions. (Some easy ones 😅.) I'm not going to add any more myself, so if anyone else wants to pick this up, it's there! :)

Thanks for the nice list to follow, @oxinabox! :)

StephenVavasis:
Notice that there is an inconsistency with `delete!` that I didn't fix because I was afraid it would be a breaking change: https://github.com/JuliaCollections/DataStructures.jl/issues/387

```
julia> w = SortedSet([4,5,9])
SortedSet([4, 5, 9],
Base.Order.ForwardOrdering())

julia> delete!(w, 12)
ERROR: KeyError: key 12 not found
Stacktrace:
 [1] delete!(::SortedSet{Int64,Base.Order.ForwardOrdering}, ::Int64) at C:\Users\vavasis\.julia\packages\DataStructures\bepIY\src\sorted_set.jl:232
 [2] top-level scope at REPL[9]:1

julia> w = Set([4,5,9])
Set([4, 9, 5])

julia> delete!(w,12)
Set([4, 9, 5])
```

oxinabox:
> Notice that there is an inconsistency with delete! that I didn't fix because I was afraid it would be a breaking change: #387

It is the general concensus that
turning errors into non-errors is not a breaking  change.

NHDaly:
Does #387 need to be addressed, then, before this can be made to be an `<:AbstractSet`?

NHDaly:
Does #387 need to be addressed, then, before this can be made to be an `<:AbstractSet`?

Also, I just updated this branch to also bump the minor version number, since changing supertype is an API change. Feel free to adjust that if you're not happy with that! :)

oxinabox:
> Also, I just updated this branch to also bump the minor version number, since changing supertype is an API change. Feel free to adjust that if you're not happy with that! :)

Pretty sure this is not breaking.
Because it's old super type is the super type of its new super type.
So it still subtypes the old thing.
(`Any`)

oxinabox:
> Does #387 need to be addressed, then, before this can be made to be an <:AbstractSet?

Seems like it should yes.

NHDaly:
@oxinabox fair point. BUT it _could be_ breaking in that it will start dispatching into methods where it wasn't dispatching previously.

But i agree probably not a likely breaking scenario.

NHDaly:
> > Does #387 need to be addressed, then, before this can be made to be an <:AbstractSet?
> 
> Seems like it should yes.

👍 SGTM. I've opened https://github.com/JuliaCollections/DataStructures.jl/pull/607 to do that.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.12"
+version = "0.18.0"
```
```suggestion
version = "0.17.13"
```

oxinabox:
> @oxinabox fair point. BUT it could be breaking in that it will start dispatching into methods where it wasn't dispatching previously.

True.
but we do have to be a bit pragmatic.
I think I will add this to my list of "Things that are not considered breaking"
(along with converting errors to nonerrors etc)
and include in it a disclaimer that if a change breaks someones code and they were not being too evil then it was breaking afterall

NHDaly:
Awesome. Thanks for the merges! :)

PR #576, opened 2020-02-09
Switch DefaultDict to use `get`

### LINKED ISSUE #575: Handle alternative types as default in DefaultDict
This is different than #530, which is asking about handling `missing` as a key. This issue relates to enabling `missing` as the default value.

This works and returns `42.0`: 
```julia
dd_real = DefaultDict(42.0,Dict(i => i for i in 1:3))
dd_real[0]
```

But this errors:
```julia
dd_missing = DefaultDict(missing,Dict(i => i for i in 1:3))
dd_missing[0]
```
```
MethodError: Cannot `convert` an object of type Missing to an object of type Int64
```

oxinabox:
Closes #575 

using `get!` means that the default must be the same type as the normal value.
Or we must during construction change the type of the value in the backing dict to be `Union{V, typeof(default)}`

using `get` does not have this requirement of matching types.

Note: this PR does change as to if multiple calls asking about the same key each call the `default` function (for the function case).
I think I prefer that it repeatedly call the default function over storing the default.
I think in most cases the default function is something cheap to compute, like a constructor.
where as hanging on to the defaults for something that is very rare could get very expensive.

E.g. if one has a set of pretrained word embeddings and one is looking them up to find the embeddings for each token in a text, then often one will encounter out of vocabulary words,
and storing all those keys and default values could start to use a lot of memory.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.10"
+version = "0.17.11"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
--- a/src/default_dict.jl
+++ b/src/default_dict.jl
@@ -66,15 +66,15 @@ function iterate(v::Base.ValueIterator{T}, i::Int) where {T <: DefaultDictBase}
     return (v.dict.d.vals[i], Base.skip_deleted(v.dict.d, i+1))
 end
 
-getindex(d::DefaultDictBase, key) = get!(d.d, key, d.default)
+getindex(d::DefaultDictBase, key) = get(d.d, key, d.default)
 
 function getindex(d::DefaultDictBase{K,V,F}, key) where {K,V,F<:Base.Callable}
     if d.passkey
-        return get!(d.d, key) do
+        return get(d.d, key) do
             d.default(key)
         end
     else
-        return get!(d.d, key) do
+        return get(d.d, key) do
             d.default()
         end
     end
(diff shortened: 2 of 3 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.10"
+version = "0.17.11"
```
oxinabox: I am unclear as to if this is a breaking change.
I don't *think* we expost this as part of our public API, but I am not sure.
Reviewer should take note and consider carefully.

alecloudenback:
Thanks! I just tested this and it works for the issue I described in #575. However, I think I lack the expertise to opine on the implementation concerns in your OP or about whether this would be breaking.

kmsquire:
Given that this breaks different workflows (well, my workflow), how about another keyword argument, `store_default`, which controls whether or not the default value is stored in the dictionary?

oxinabox:
We could instead go for the other solution.
Which would be to make sure the default value can fit in the backing store.
Via unioning it in the constructor.

kmsquire:
Not at a computer, but does unioning two types alway produce a boxed value?  What if they’re the same type?

A value type of `Union{Int32,Int64}` doesn’t seem very nice if someone specifies a value type of `Int32`, but a default of `0` (on a 64-bit system).

oxinabox:
If same type it doesn't.
And if it's less than 5 types, then the small union optimisation kicks in.
Performance should not be too bad

StephenVavasis:
I would like to vote against this change.  The OP wanted the following to work:
```
h = DefaultDict(()->90, 1=>3)
h["abc"]
   90
```
However, I don't agree that this is a good idea.  In particular, a benefit of requiring strong typing on a Dict is to catch programmer errors.  Accepting keys of any type undermines this benefit.  Furthermore, the behavior requested by the OP is already accessible by explicitly declaring the types:
```
h = DefaultDict{Any,Int}(() -> 90)
h[1] = 3
h["abc"]
    90
```

oxinabox:
Values, not keys.

Issue #570, opened 2020-01-30
DisjointSets hangs if input is a collection of Any


scheinerman:
If the input to `DisjointSets` is a collection with elements of type `Any` the constructor hangs. 
This works:
```julia
julia> A = [1,2,3];

julia> DisjointSets(A)
DisjointSets{Int64}(Dict(2 => 2,3 => 3,1 => 1), [1, 2, 3], IntDisjointSets([1, 2, 3], [0, 0, 0], 3))
```
but this just hangs:
```julia
julia> B = Any[1,2,3];

julia> DisjointSets(B)
```

oxinabox:
I will investigate and fix

oxinabox:
I think this is now fixed

jwbaugh:
Still hangs:

```
julia> versioninfo()
Julia Version 1.10.0
Commit 3120989f39b (2023-12-25 18:01 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
  Threads: 5 on 4 virtual cores

julia> DisjointSets([1, 2, 3])
DisjointSets{Int64} with 3 elements:
  1
  2
  3

julia> DisjointSets(Any[1, 2, 3])
```

PR #578, opened 2020-02-10
Remove DisjointSet constructor taking splatted elements

oxinabox:
This closes https://github.com/JuliaCollections/DataStructures.jl/issues/570
I can't work out how to do this without a breaking change.
I also can't workout how to deprecate the old behavior.

Normal julia collections do not allow construction with passing the iterator elements in.
That is why we have `Tuple((1,2,3))` and `tuple(1,2,3)`.

```julia
julia> Set(1, 2, 3)
ERROR: MethodError: no method matching Set(::Int64, ::Int64, ::Int64)
Closest candidates are:
  Set(::Any) at set.jl:21
Stacktrace:
 [1] top-level scope at REPL[1]:1

julia> Set((1, 2, 3))
Set([2, 3, 1])
```


oxinabox:
There is one way I can think of deprecating it.
We deprecate the whole `DisjointSets` type  in favour of a new `DisjointSet` type.
https://github.com/JuliaCollections/DataStructures.jl/issues/579

PR #582, opened 2020-02-21
Update deque.md

matthieugomez:
Not sure what `unshift!` means but it sounds like it was what `pushfirst!` is now


oxinabox:
correct, thanks

PR #588, opened 2020-03-07
Allow @inbounds for deque, queue and stack

simonschoelly:
This PR allows disabling boundchecks for some methods in deque, queue and stack.

I should maybe mention, that the method `first` from the Julia base, does not allow one to disable boundchecks for Vectors, although this package already allows one to disable boundchecks for `first` on `CircularDeque`. 

I will now make a basic benchmark and post the results here.

### DIFF
--- a/src/deque.jl
+++ b/src/deque.jl
@@ -97,6 +97,9 @@ Base.eltype(::Type{Deque{T}}) where T = T
     front(q::Deque)
 
 Returns the first element of the deque `q`.
+
+Throws an `ArgumentError` if the deque is empty. This check
+can be disabled with `@inbounds`.
 """
 @inline Base.@propagate_inbounds function first(q::Deque)
     @boundscheck isempty(q) && throw(ArgumentError("Deque must be non-empty"))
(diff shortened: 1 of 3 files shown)


src/deque.jl
```diff
@@ -94,23 +94,29 @@ num_blocks(q::Deque) = q.nblocks
...
+    first(q::Deque)
 
 Returns the first element of the deque `q`.
+
+Throws an `ArgumentError` if the deque is empty. This check
+can be disabled with `@inbounds`.
 """
-function first(q::Deque)
-    isempty(q) && throw(ArgumentError("Deque must be non-empty"))
+@inline Base.@propagate_inbounds function first(q::Deque)
```
oxinabox: is the `@inline` needed?

simonschoelly:
Let's wait  a bit before merging, I originally wrote some dfs and bfs connected components algorithms for benchmarking, but did not get better results (maybe even slightly worse). I will have to check, if it was a problem with the benchmarks, or if it is something else (maybe the inlining might cause some issues).

I don't really have time during the week, but I can work on that on the weekend.

oxinabox:
bump

oxinabox:
bump

eulerkochy:
bump

PR #592, opened 2020-03-08
Feat(Fibonacci_heap) : first simple implementation of fibonacci Heap

karan0299:
A Fibonacci heap is a data structure for priority queue operations, consisting of a collection of heap-ordered trees. It has a better amortized running time than many other priority queue data structures including the binary heap and binomial heap.
For the Fibonacci heap, the find-minimum operation takes constant (O(1)) amortized time. The insert and decrease key operations also work in constant amortized time.Deleting an element (most often used in the special case of deleting the minimum element) works in O(log n) amortized time, where n is the size of the heap.
Interfaces are:
1. Push 
2. Extractmin
3. Top  
4. Delete
5. Decreasekey
6. show

Here I have provided a simple implementation of the Fibonacci heap.
**Tests and Documentation regarding the same will in follow up commits**


oxinabox:
(will wait for tests and docs before reviewing)

src/heaps/fibonacci_heap.jl
```diff
@@ -0,0 +1,367 @@
...
+    else
+        node.right = parent.child.right
+        node.left = parent.child
+        parent.child.right.left = node
+        parent.child.right = node
+    end
+end
+
+# extract (delete) the min node from the heap in O(log n) time
+function extractMin(root::MutableFibonacciHeap)
```
jakobnissen: This PR has a mix of camelCase and snake_case. Julia's style guide (and the rest of DataStructures, as far as I know) uses snake_case, so that is preferable.

src/heaps/fibonacci_heap.jl
```diff
@@ -0,0 +1,367 @@
...
+        else
+            cut(root, y, pNode)
+            cascadingcut(root, pNode)
+        end
+    end
+end
+
+# Get list of nodes attached to head (of a circular doubly linked list)
+function aslist(head::MutableFibonacciHeapNode)
+    nodelist, node, stop = MutableFibonacciHeapNode[], head, head
```
jakobnissen: The type of `MutableFibonacciHeapNode[]` is not fully specified, since the type parameter is unspecified. You can do `typeof(head)[]` instead. This impacts performance.
You also have un-specified types in lines 224 and 282.

src/heaps/fibonacci_heap.jl
```diff
@@ -0,0 +1,367 @@
...
+            y = A[d]
+            if x.value > y.value
+                x, y = y, x
+            end
+            heaplink(root, y, x)
+            A[d] = nothing
+            d += 1
+        end
+        A[d] = x
+    end
```
jakobnissen: It might be worth adding `@inbounds` to this loops if it's a hot spot. Not sure though!

src/heaps/fibonacci_heap.jl
```diff
@@ -0,0 +1,367 @@
...
+function push!(root::MutableFibonacciHeap,data::T) where T
+    node = MutableFibonacciHeapNode(data)
+    node.left = node.right = node
+    _merge_with_root_list(root, node)
+    if root.minNode === nothing || node.value < root.minNode.value
+        root.minNode = node
+    end
+    root.nodeCount += 1
+    return node
+end
```
jakobnissen: The signature of this method both in the code and the documentation is a bit  off:
* First, the docs write `pop!{T}`, even though functions are not parameterized.
* Second, `MutableFibonacciHeap` in the type signature should probably be `MutableFibonacciHeap{T}`? I think it's preferrable instead to specify the type of `data` and instead have `dataT = convert(T, data)` in the first line. This will allow e.g. pushing a `UInt` to a `MutableFibonacciHeap{Int}`.

src/heaps/fibonacci_heap.jl
```diff
@@ -0,0 +1,367 @@
...
+pop!(root::MutableFibonacciHeap) = extractMin(root)
+
+popmin!(root::MutableFibonacciHeap) = extractMin(root)
+"""
+    merge!{T}(h1::MutableFibonacciHeap{T}, h2::MutableFibonacciHeap{T})
+
+merge two fibonacci heapsby concatenating the root lists of heap `h1` and `h2`
+returns handler to merged list
+"""
+function merge!(h1::MutableFibonacciHeap{T}, h2::MutableFibonacciHeap{T}) where T
```
jakobnissen: Same as above.

src/heaps/fibonacci_heap.jl
```diff
@@ -0,0 +1,367 @@
...
+    newh.nodeCount = h1.nodeCount + h2.nodeCount
+    return newh
+end
+
+"""
+    update!{T}(h::MutableFibonacciHeap{T}, x::Any, k::Any}
+
+modify the data of node `x` in the heap `h` to value `k`
+"""
+function update!(root::MutableFibonacciHeap{T}, oldkey::T, newkey::T) where T
```
jakobnissen: Same as above, and in both this function and `delete!`, the signature of the documentation do not match the signature of the function.

PR #595, opened 2020-03-13
Display the type of list

### LINKED ISSUE #591: Distinction between the type of `list`
```
julia> cons(1, cons(2, nil()))
list(1, 2)

julia> cons(1, cons(2, nil(Int32)))
list(1, 2)

julia> typeof(cons(1, cons(2, nil())))
Cons{Any}

julia> typeof(cons(1, cons(2, nil(Int32))))
Cons{Int32}
```

what I'm calling out for is that `Base.show` for list should display the type of list, just to make a distinction in the above mentioned case.

eulerkochy:
Fixes #591 .
Current behaviour:
```
julia> cons(1, cons(2, nil()))
list(1, 2)

julia> cons(1, cons(2, nil(Int32)))
list(1, 2)
```

On this branch:
```
julia> cons(1, cons(2, nil()))
list(1, 2)

julia> cons(1, cons(2, nil(Int32)))
list{Int32}(1, 2)
```

### DIFF
--- a/src/list.jl
+++ b/src/list.jl
@@ -29,7 +29,11 @@ function show(io::IO, l::LinkedList{T}) where T
             print(io, "nil(", T, ")")
         end
     else
-        print(io, "list(")
+        if T === Any
+            print(io, "list(")
+        else
+            print(io, "list{$T}(")
+        end
         show(io, head(l))
         for t in tail(l)
             print(io, ", ")


oxinabox:
@nickrobinson251 we need to write a blog post about overloading show.

---

I think this looks right, does the output satisfy `eval(Mera.parse(repr(x))) == x`?

eulerkochy:
> I think this looks right, does the output satisfy `eval(Mera.parse(repr(x))) == x`?

Throws a `TypeError` :cry:  ..
image

PR #597, opened 2020-03-16
[DO NOT MERGE]Feat(Datastructures.jl): binomial heap implementation

karan0299:
Binomial Heap is an extension of Binary Heap that provides faster union or merge operation together with other operations provided by Binary Heap.
Operations of Binomial Heap:
The main operation in Binomial Heap is union(), all other operations mainly use this operation. The union() operation is to combine two Binomial Heaps into one.
1)union(H1,H2) : Merges heap H1 and H2 into single heap H1. Complexity of the operation is O(Logn).
2) insert(H, k): Inserts a key ‘k’ to Binomial Heap ‘H’. This operation first creates a Binomial Heap with single key ‘k’, then calls union on H and the new Binomial heap.
3) getMin(H): A simple way to getMin() is to traverse the list of root of Binomial Trees and return the minimum key. This implementation requires O(Logn) time. It can be optimized to O(1) by maintaining a pointer to minimum key root.
4) extractMin(H): This operation also uses union(). We first call getMin() to find the minimum key Binomial Tree, then we remove the node and create a new Binomial Heap by connecting all subtrees of the removed minimum node. Finally, we call union() on H and the newly created Binomial Heap. This operation requires O(Logn) time.
5) delete(H): Like Binary Heap, delete operation first reduces the key to minus infinite, then calls extractMin().
6) decreaseKey(H): decreaseKey() is also similar to Binary Heap. We compare the decreases key with it parent and if parent’s key is more, we swap keys and recur for the parent. We stop when we either reach a node whose parent has a smaller key or we hit the root node. Time complexity of decreaseKey() is O(Logn).
**Docs and Tests for Binomial Heap will be in followup commits**


karan0299:
@oxinabox whenever you find time , this PR is ready for review . Thanks

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,368 @@
...
+end
+
+function printTree(io::IO, h::Union{MutableBinomialHeapNode{T},Nothing}) where T
+    while h !== nothing 
+        print(io, h.data)
+        print(io, ",")
+        printTree(io, h.child) 
+        h = h.sibling; 
+    end
+end
```
jakobnissen: While this is only a matter of style, the Julia style guide recommends that functions are all lowercase with underscores between them (`print_tree`). This is also the style followed in the rest of DataStructures. That would be nice to change.
Furthermore, many of your internal functions that are mutating do not end with `!`, even though this is the convention.

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,368 @@
...
+    if n == 0
+        return nothing
+    end
+    decreaseKey!(heap, i, typemin(T)) 
+    extractMin!(heap)
+    return heap
+end
+
+setindex!(h::MutableBinomialHeap, v, i::Int) = update!(h, i, v)
+getindex(h::MutableBinomialHeap, i::Int) =  findNode(h, i).data
```
jakobnissen: These two functions are not covered in the tests, but they should be.

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,368 @@
...
+#################################################
+# make Binomial_heap from Vector
+function _make_binomial_heap(xs::AbstractVector{T}, Typ::Type{T}) where T
+    len = length(xs)
+    BinomialHeap = MutableBinomialHeap{Typ}()
+    for i = 1:len
+        push!(BinomialHeap, xs[i])
+    end
+    BinomialHeap
+end
```
jakobnissen: Just a tip regarding the iteration and the type specification.
```
function _make_binomial_heap(xs::AbstractVector{T}, ::Type{T}) where T
    BinomialHeap = MutableBinomialHeap{T}()
    for i in xs
        push!(BinomialHeap, i)
    end
    return BinomialHeap
end
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,368 @@
...
+
+function insertATreeInHeap(_heap::MutableBinomialHeap{T}, tree::MutableBinomialHeapNode{T}) where T 
+    temp = MutableBinomialHeap{T}()
+    push!(temp.rootList, tree)
+    union!(_heap, temp, true)
+    adjust(_heap)
+    return
+end   
+function removeMinFromTreeReturnBHeap(tree::MutableBinomialHeapNode{T}) where T 
+    heap::Vector{Union{Nothing,MutableBinomialHeapNode{T}}} = []
```
jakobnissen: Instead of instantiating a `Vector{Any}` and converting it, thereby allocating two arrays, you can simply do `heap = Union{Nothing,MutableBinomialHeapNode{T}}[]`

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,368 @@
...
+        try
+            node = findNode(h, i)
+            node.handle = node.handle + c
+        catch _
+            continue
+        end
+    end
+end
+
+function union!(h1::MutableBinomialHeap{T}, h2::MutableBinomialHeap{T}, flag::Bool) where T
```
jakobnissen: I think it would be nice to write what `flag` does, either in the name (i.e. call it `update_handle` or something), or in a comment above the function. Perhaps more importantly, consider renaming this function. `union!` is a `Base` function, but this signature does not match the signature expected of `Base.union!`. Perhaps just call it `_union!`.

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,368 @@
...
+minimum(heap::MutableBinomialHeap) =  top_with_handle(heap)[1]
+"""
+top_with_handle(h::MutableBinomialHeap)
+
+Returns the minimum element of the heap `h` and its handle.
+"""
+function top_with_handle(h::MutableBinomialHeap) 
+    (val, node) = top_elem_with_handle(h)
+    return (val, node.handle)
+end
```
jakobnissen: This interface here could use a rewrite, I think. If the heap is empty, `top_elem_with_handle(h)` will return `nothing`, which causes this to fail with an obscure error. Also, since the compiler can't guarantee that `top_elem_with_handle` will return a 2-tuple, it needs all sorts of boilerplate code in this function which prevents all these 3 functions from inlining, despite the `@inline` notation.
Perhaps rewrite it such that `top_elem_with_handle` either raises an error, or returns a 2-tuple, and then make these three functions `@inline` and `@inbounds` for efficiency?

src/heaps/binomial_heap.jl
karan0299: Although I tried to follow style guidelines still, I missed a few places, thanks, doing it right now

src/heaps/binomial_heap.jl
karan0299: Yeah i add them

src/heaps/binomial_heap.jl
karan0299: yeah, I am on this

src/heaps/binomial_heap.jl
karan0299: thanks, this is surely better.

src/heaps/binomial_heap.jl
jakobnissen: By the way, `Base.generator` uses the following trick: Say `foo` is a function that can return either a 2-tuple or `nothing`. Then you can do:
```
val = foo()
if val === nothing
    # do stuff with nothing
else
    val = val::Tuple{Any,Any} # force compiler to know val is a 2-tuple
    return val[1] # compiler will now remove the boundscheck
end
```
This is just as efficient as rewriting `top_elem_with_handle` to never return `nothing`, but may be easier to write.

docs/src/heaps.md
```diff
@@ -87,6 +87,32 @@ h = BinaryMinMaxHeap{Int}()          # create an empty min-max heap with integer
 h = BinaryMinMaxHeap([1, 2, 3, 4]) # create a min-max heap from a vector
 ```
 
+## Binomial heap
+Binomial heaps allow union of two heaps in O(logn) time.
+The Binomial heap in this package are subtypes of `AbstractHeap`
+and have the same interface as other heaps with the following additions:
```
oxinabox, line 93:
```suggestion
and has the same interface as other heaps with the following additions:
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    cumm_nodecount::Int
+    MutableBinomialHeap{T}() where T = new{T}([], 0, 0)
+
+    function MutableBinomialHeap{T}(xs::AbstractVector{T}) where T
+        BinomialHeap = _make_binomial_heap(xs, T)
+        new{T}(BinomialHeap.rootList, BinomialHeap.nodecount, BinomialHeap.cumm_nodecount)
+    end
+    
+end
+const BinomialHeap{T} = MutableBinomialHeap{T}
```
oxinabox, line 28: Why do we have this line?
Why have we not just called this type `BinomialHeap`?
Or `MutableBinomialHeap`?
What is the advantage of this alias?

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    sibling::Union{MutableBinomialHeapNode,Nothing}
+end
+MutableBinomialHeapNode(data::T, handle::Int) where T = MutableBinomialHeapNode{T}(data, 0, handle, nothing, nothing, nothing)
+#################################################
+#
+#   heap type and constructors
+#
+#################################################
+mutable struct MutableBinomialHeap{T} <: AbstractHeap{T}
+    rootList::Vector{Union{Nothing,MutableBinomialHeapNode{T}}}
```
oxinabox, line 17:
```suggestion
    rootList::Vector{Union{Nothing, MutableBinomialHeapNode{T}}}
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
+# Binomial Heap
+mutable struct MutableBinomialHeapNode{T} 
+    data::T
+    degree::Int
+    handle::Int
+    parent::Union{MutableBinomialHeapNode,Nothing}
+    child::Union{MutableBinomialHeapNode,Nothing}
+    sibling::Union{MutableBinomialHeapNode,Nothing}
```
oxinabox, line 8:
```suggestion
    parent::Union{MutableBinomialHeapNode, Nothing}
    child::Union{MutableBinomialHeapNode, Nothing}
    sibling::Union{MutableBinomialHeapNode, Nothing}
```

docs/src/heaps.md
```diff
@@ -87,6 +87,32 @@ h = BinaryMinMaxHeap{Int}()          # create an empty min-max heap with integer
...
+
+This package includes an implementation of a binomial heap (`BinomialHeap`).
+> Cormen, Thomas H.; Leiserson, Charles E.; Rivest, Ronald L.; Stein, Clifford (2001) [1990]. "Chapter 19: Binomial Heaps". Introduction to Algorithms (2nd ed.). MIT Press and McGraw-Hill. pp. 455–475. 
+>ISBN 0-262-03293-7.
+
+Examples:
+```julia
+h = MutableBinomialHeap{Int}()          # create an empty binomial heap with integer values
+
+h = MutableBinomialHeap([1, 2, 3, 4]) # create a binomial heap from a vector
```
oxinabox, line 113:
```suggestion
h = MutableBinomialHeap([1, 2, 3, 4])  # create a binomial heap from a vector
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    rootList::Vector{Union{Nothing,MutableBinomialHeapNode{T}}}
+    nodecount::Int
+    cumm_nodecount::Int
+    MutableBinomialHeap{T}() where T = new{T}([], 0, 0)
+
+    function MutableBinomialHeap{T}(xs::AbstractVector{T}) where T
+        BinomialHeap = _make_binomial_heap(xs, T)
+        new{T}(BinomialHeap.rootList, BinomialHeap.nodecount, BinomialHeap.cumm_nodecount)
+    end
+
```
oxinabox, line 26:
```suggestion
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+function show(io::IO, _heap::MutableBinomialHeap{T}) where T
+    print(io, "MutableBinomialHeap")
+    list = _heap.rootList
+    n = length(list)
+    if n == 0
+        print(io, "{$(T)}")
+    end
+    print(io, "(")
+    for i = 1:n
+        print_tree(io, list[i])
```
oxinabox, line 44:
```suggestion
    for node in list
        print_tree(io, node)
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    end
+    print(io, ")")
+end
+
+function print_tree(io::IO, h::Union{MutableBinomialHeapNode{T},Nothing}) where T
+    while h !== nothing 
+        print(io, h.data)
+        print(io, ",")
```
oxinabox, line 52:
```suggestion
        print(io, ", ")
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+        new{T}(BinomialHeap.rootList, BinomialHeap.nodecount, BinomialHeap.cumm_nodecount)
+    end
+    
+end
+const BinomialHeap{T} = MutableBinomialHeap{T}
+BinomialHeap(xs::AbstractVector{T}) where T = MutableBinomialHeap{T}(xs)
+
+struct HeapBoundsError <: Exception
+    msg::AbstractString
+end
```
oxinabox, line 33: Why now use the existing `BoundsError` ?

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    print(io, "(")
+    for i = 1:n
+        print_tree(io, list[i])
+    end
+    print(io, ")")
+end
+
+function print_tree(io::IO, h::Union{MutableBinomialHeapNode{T},Nothing}) where T
+    while h !== nothing 
+        print(io, h.data)
```
oxinabox, line 51: Since overloading `show` need to call `show` on elements.
Recall that `show` promises the output will be parsable (its the io-stream version of `repr`)
```suggestion
function show_tree(io::IO, h::Union{MutableBinomialHeapNode{T},Nothing}) where T
    while h !== nothing 
        show(io, h.data)
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    list = _heap.rootList
+    n = length(list)
+    if n == 0
+        print(io, "{$(T)}")
+    end
+    print(io, "(")
+    for i = 1:n
+        print_tree(io, list[i])
+    end
+    print(io, ")")
```
oxinabox, line 46:
```suggestion
    print(io, "])")
```

Needed to make it a valid constructor

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+end
+
+function show(io::IO, _heap::MutableBinomialHeap{T}) where T
+    print(io, "MutableBinomialHeap")
+    list = _heap.rootList
+    n = length(list)
+    if n == 0
+        print(io, "{$(T)}")
+    end
+    print(io, "(")
```
oxinabox, line 42: `show` needs to return a valid constructor
```suggestion
    print(io, "([")
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+#
+#################################################
+# make Binomial_heap from Vector
+function _make_binomial_heap(xs::AbstractVector{T}, Typ::Type{T}) where T
+    BinomialHeap = MutableBinomialHeap{Typ}()
+    for x in xs
+        push!(BinomialHeap, x)
+    end
+    BinomialHeap
+end
```
oxinabox, line 69:
```suggestion
function _make_binomial_heap(xs)
    T = eltype(xs)
    BinomialHeap = MutableBinomialHeap{T}()
    for x in xs
        push!(binomial_heap, x)
    end
    return binomial_heap
end
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+#   heap type and constructors
+#
+#################################################
+mutable struct MutableBinomialHeap{T} <: AbstractHeap{T}
+    rootList::Vector{Union{Nothing,MutableBinomialHeapNode{T}}}
+    nodecount::Int
+    cumm_nodecount::Int
+    MutableBinomialHeap{T}() where T = new{T}([], 0, 0)
+
+    function MutableBinomialHeap{T}(xs::AbstractVector{T}) where T
```
oxinabox, line 22: Where-ever possible, a collection should have a constructor that accepts an interator,

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    if b1.data > b2.data
+        b1, b2 = b2, b1
+    end
+    b2.parent = b1 
+    b2.sibling = b1.child 
+    b1.child = b2 
+    b1.degree += 1 
+    return b1
+end
+# adjust! function rearranges the heap so that
```
oxinabox, line 82:
```suggestion

# adjust! function rearranges the heap so that 
```

It this this the same as what we call `heapify` elsewhere?

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    b2.parent = b1 
+    b2.sibling = b1.child 
+    b1.child = b2 
+    b1.degree += 1 
+    return b1
+end
+# adjust! function rearranges the heap so that 
+# heap is in increasing order of degree and 
+# no two binomial trees have same degree in this heap 
+function adjust!(heap::MutableBinomialHeap{T}) where T
```
oxinabox, line 85:
```suggestion
function adjust!(heap::MutableBinomialHeap)
```

`T` is never used so it does not need ot be in the type signature

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+            end
+        elseif (k < current_size)  && (_heap[i].degree == _heap[j].degree) && (_heap[i].degree == _heap[k].degree)
+            i += 1
+            j += 1
+            k += 1  
+        elseif _heap[i].degree == _heap[j].degree
+            temp = merge_binomial_trees!(_heap[i], _heap[j])
+            _heap[i] = temp
+            @inbounds _heap[j] = nothing
+            filter!(e->e ≠ nothing, _heap)
```
oxinabox, line 117: Is this the same as:
```suggestion
            deleteat!(_heap, j)
```

Or are there other elements we are filtering out at the same time?

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+            j += 1
+            if k !== current_size
+                k += 1
+            end
+        elseif (k < current_size)  && (_heap[i].degree == _heap[j].degree) && (_heap[i].degree == _heap[k].degree)
+            i += 1
+            j += 1
+            k += 1  
+        elseif _heap[i].degree == _heap[j].degree
+            temp = merge_binomial_trees!(_heap[i], _heap[j])
```
oxinabox, line 114: is there are clearer name we can give this?
Maybe `new_subheap` ? or something like that?

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+            filter!(e->e ≠ nothing, _heap)
+            if k < (length(_heap) + 1)
+                k = j + 1 
+            end
+        end
+    end
+end
+
+function insert_tree_in_heap!(_heap::MutableBinomialHeap{T}, tree::MutableBinomialHeapNode{T}) where T 
+    temp = MutableBinomialHeap{T}()
```
oxinabox, line 126: is there a clere name we can give this?

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    push!(temp.rootList, tree)
+    _union!(_heap, temp, true)
+    adjust!(_heap)
+    return
+end   
+function remove_min__from_tree!(tree::MutableBinomialHeapNode{T}) where T
```
oxinabox, line 132:
```suggestion

function remove_min_from_tree!(tree::MutableBinomialHeapNode{T}) where T 
```

Seemed to have a double `_` and also mised the line break between functions

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    return reverse!(heap) 
+end
+
+function extract_min!(heap::MutableBinomialHeap{T}) where T
+    _heap = heap.rootList
+    new_heap = MutableBinomialHeap{T}()
+    lo = MutableBinomialHeap{T}()
+    temp = top_element_node(heap)
+    it = iterate(_heap)
+    while it !== nothing
```
oxinabox, line 150: is there a reason this is not written as 
`for item in _heap` ?

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+        return nothing
+    end
+    if node.handle == i 
+        return node
+    end
+    res = find_helper(node.child, i)
+    if res !== nothing 
+        return res
+    end
+    return find_helper(node.sibling, i)
```
oxinabox, line 176: Shorter for this kind of thing to use short-cut evaluation
```suggestion
    node === nothing && return nothing
    node.handle == i && return node
    res = find_helper(node.child, i)
    res  !== nothing && return res
    return find_helper(node.sibling, i)
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+        return nothing
+    end
+    for j = 1:n
+        node = find_helper(_heap[j], i)
+        if node !== nothing
+            break
+        end
+    end
+    if node === nothing
+        throw(HeapBoundsError("attempt to access $(heap.nodecount) MutableBinomialHeap{$(T)} at index [$(heap.nodecount + 1)]"))
```
oxinabox, line 193: If we are going to use a custoim exception type for this, we are better to make these interpolated things fields of the exceptiom,
and then overload `Base.showerror`

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    n = length(_heap)
+    if n == 0
+        return nothing
+    end
+    for j = 1:n
+        node = find_helper(_heap[j], i)
+        if node !== nothing
+            break
+        end
+    end
```
oxinabox, line 191:
```suggestion
    length(_heap) == 0 && return nothing
    node = nothing
    for sub in _heap
        node = find_helper(sub, i)
        node !== nothing && break
    end
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    if node === nothing
+        throw(HeapBoundsError("attempt to access $(heap.nodecount) MutableBinomialHeap{$(T)} at index [$(heap.nodecount + 1)]"))
+    end
+    return node
+end
+function increase_key!(node::MutableBinomialHeapNode{T}, new_val::T) where T
```
oxinabox, line 197:
```suggestion

function increase_key!(node::MutableBinomialHeapNode{T}, new_val::T) where T
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+    node.data = new_val
+    child = node.child
+    while child !== nothing && node.data > child.data
+        node.data, child.data = child.data, node.data
```
oxinabox, line 201:
```suggestion
        # swap node and child
        node.data, child.data = child.data, node.data
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,371 @@
...
+        node.handle, child.handle = child.handle, node.handle
+        node = child
+        child = child.child
+    end
+end
+function decrease_key!(heap::MutableBinomialHeap{T}, i::Int, new_val) where T
```
oxinabox, line 207:
```suggestion

function decrease_key!(heap::MutableBinomialHeap{T}, i::Int, new_val) where T
```

src/DataStructures.jl
```diff
@@ -38,6 +38,7 @@ module DataStructures
     export MutableBinaryHeap, MutableBinaryMinHeap, MutableBinaryMaxHeap
     export heapify!, heapify, heappop!, heappush!, isheap
     export BinaryMinMaxHeap, popmin!, popmax!, popall!
+    export MutableBinomialHeap, MutableBinomialHeapNode, push!, show, top, top_with_handle, minimum, _union!, update!, delete!, HeapBoundsError, find_node
```
oxinabox, line 41: We do not need to reexport things we overload from `Base`, 

Also what is up with experingign `_union!` ?

src/heaps/binomial_heap.jl
karan0299: Ok I will shift to Mutable Binomial Heap

src/heaps/binomial_heap.jl
karan0299: Ok , I got it. Thanks

karan0299:
Thanks, @oxinabox  for taking time out of busy schedule, I will surely make all the required changes in quick time.

src/heaps/binomial_heap.jl
karan0299: Since the above implementation uses vector only for root nodes, so for the nodes other than root nodes, I had to resort to this method to show out of Bounds access.
Furthermore, I also wish to wrap the bounds error for all types of heaps, as I have opened the issue #598.

src/heaps/binomial_heap.jl
karan0299: Yes, semantically it is similar.

src/DataStructures.jl
karan0299: _union! is used to merge two binomial heaps

src/heaps/binomial_heap.jl
oxinabox: We should then call it `heapify` and let dispatch take care of it.

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,392 @@
...
+    while (iter1 !== nothing)
+        (elem1, state1) = iter1 
+        push!(_new.rootList, elem1) 
+        iter1 = iterate(l1, state1) 
+    end
+    while (iter2 !== nothing)
+        (elem2, state2) = iter2
+        push!(_new.rootList, elem2)
+        iter2 = iterate(l2, state2)
+    end
```
oxinabox, line 306:
```suggestion
        append!(_new.rootList, Iterators.rest(l1, state1))
        append!(_new.rootList, Iterators.rest(l2, state2)) 
    end
```

src/heaps/binomial_heap.jl
```diff
@@ -0,0 +1,392 @@
...
+length(h::MutableBinomialHeap) = h.nodecount
+
+isempty(h::MutableBinomialHeap) = isempty(h.rootList)
+
+"""
+_union!(h1::MutableBinomialHeap, h2::MutableBinomialHeap)
+
+merges heap `h2` into heap `h1`
+"""
+_union!(h1::MutableBinomialHeap, h2::MutableBinomialHeap) = _union!(h1, h2, true)
```
oxinabox, line 334: we can't have part of the public API prefixed with `_`.
That is a common convention for "Not part of public API`

We should just call this `Base.union!`

test/test_binomial_heap.jl
```diff
@@ -0,0 +1,215 @@
...
+    vs = Vector{VT}(undef, n)
+    k = 1
+    val::Union{Nothing,MutableBinomialHeapNode{VT}} = nothing
+    for i = 1:h.cumm_nodecount
+        val = nothing
+        try
+            val = find_node(h, i)
+            vs[k] = val.data
+            k += 1
+        catch _
```
oxinabox, line 17: why are we doing this?
Probably we want to be catching an error and then checking if it is the right type
and if it is `continue` and if not `rethrow()`

Catching all exceptions makes debugging harder

PR #599, opened 2020-03-30
type-parameterized IntDisjointSets

sbromberger:
This allows the creation of type-parameterized `IntDisjointSets`, and keeps the default as `Int` for compatibility. (It also explicitly creates an `IntDisjointSets{Int}` for `DisjointSets`, which does not further type-specialize.)

### DIFF
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -69,7 +69,7 @@ end
 Find the root element of the subset that contains an member x.
 Path compression is implemented here.
 """
-find_root(s::IntDisjointSets{T}, x::T) where {T<:Integer} = find_root_impl!(s.parents, x)
+find_root!(s::IntDisjointSets{T}, x::T) where {T<:Integer} = find_root_impl!(s.parents, x)
 
 """
     in_same_set(s::IntDisjointSets{T}, x::T, y::T)
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -107,10 +107,10 @@ function root_union!(s::IntDisjointSets{T}, x::T, y::T) where {T<:Integer}
     if xrank < yrank
         x, y = y, x
     elseif xrank == yrank
-        rks[x] += 1
+        rks[x] += one(T)
     end
     @inbounds parents[y] = x
-    s.ngroups -= 1
+    s.ngroups -= one(T)
     return x
 end
 
@@ -124,10 +124,10 @@ capacity of the set would be exceeded.
 function push!(s::IntDisjointSets{T}) where {T<:Integer}
     l = length(s)
     l < typemax(T) || throw(ArgumentError(_intdisjointsets_bounds_err_msg(T))) 
-    x = l + 1
+    x = l + one(T)
     push!(s.parents, x)
-    push!(s.ranks, 0)
-    s.ngroups += 1
+    push!(s.ranks, zero(T))
+    s.ngroups += one(T)
     return x
 end
 
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -76,7 +76,7 @@ find_root!(s::IntDisjointSets{T}, x::T) where {T<:Integer} = find_root_impl!(s.p
 
 Returns `true` if `x` and `y` belong to the same subset in `s` and `false` otherwise.
 """
-in_same_set(s::IntDisjointSets{T}, x::T, y::T) where {T<:Integer} = find_root(s, x) == find_root(s, y)
+in_same_set(s::IntDisjointSets{T}, x::T, y::T) where {T<:Integer} = find_root!(s, x) == find_root!(s, y)
 
 """
     union!(s::IntDisjointSets{T}, x::T, y::T)
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -13,23 +13,27 @@
 #
 ############################################################
 
+_intdisjointsets_bounds_err_msg(T) = "the maximum number of elements in IntDisjointSets{$T} is $(typemax(T))"
+
 """
-    IntDisjointSets(n::Integer)
+    IntDisjointSets{T<:Integer}(n::Integer)
 
 A forest of disjoint sets of integers, which is a data structure
 (also called a union–find data structure or merge–find set)
 that tracks a set of elements partitioned
 into a number of disjoint (non-overlapping) subsets.
 """
-mutable struct IntDisjointSets
-    parents::Vector{Int}
-    ranks::Vector{Int}
-    ngroups::Int
+mutable struct IntDisjointSets{T<:Integer}
+    parents::Vector{T}
+    ranks::Vector{T}
+    ngroups::T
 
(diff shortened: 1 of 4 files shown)


src/disjoint_set.jl
```diff
@@ -111,13 +115,16 @@ function root_union!(s::IntDisjointSets, x::Integer, y::Integer)
...
+capacity of the set would be exceeded.
 """
-function push!(s::IntDisjointSets)
-    x = length(s) + 1
+function push!(s::IntDisjointSets{T}) where {T<:Integer}
+    l = length(s)
+    l < typemax(T) || throw(ArgumentError(_intdisjointsets_bounds_err_msg(T))) 
+    x = l + 1
     push!(s.parents, x)
     push!(s.ranks, 0)
```
eulerkochy:
```suggestion
    push!(s.ranks, zero(T))
```

eulerkochy: similar changes for 1, required in the rest of the code. eg in L110, 113

sbromberger: absolutely.

test/test_disjoint_set.jl
```diff
@@ -1,52 +1,70 @@
...
-            @test num_groups(s) == 10
-
-            for i = 1:10
-                @test find_root(s, i) == i
+        for T in [Int, UInt8, Int8, UInt16, Int16, UInt32, Int32, UInt64]
+            @testset "eltype = $(T)" begin
+                s = IntDisjointSets(T(10))
+                s2 = IntDisjointSets{T}(10)
+
+
```
eulerkochy:
```suggestion
```

test/test_disjoint_set.jl
```diff
@@ -1,52 +1,70 @@
...
+                    @test length(s) == 10
+                    @test length(s2) == 10
+                    @test eltype(s) == T
+                    @test eltype(s2) == T
+                    @test eltype(typeof(s)) == T
+                    @test eltype(typeof(s2)) == T
+                    @test num_groups(s) == T(10)
+                    @test num_groups(s2) == T(10)
+
+                    for i = 1:10
```
eulerkochy:
```suggestion
                    for i in 1:10
```

test/test_disjoint_set.jl
```diff
@@ -1,52 +1,70 @@
...
+                        @test find_root(s, T(i)) == T(i)
```
eulerkochy:
```suggestion
                        @test find_root!(s, T(i)) == T(i)
```

test/test_disjoint_set.jl
eulerkochy: in this codebase, the standard way is to use `in` instead of `=` for looping over a range. Please correct this too.

sbromberger: OK - but I didn't change this from the original:

https://github.com/JuliaCollections/DataStructures.jl/blob/a4246e7584f3f3d3e94058e686fdf748e5151479/test/test_disjoint_set.jl#L12

eulerkochy: cool .. leave that .. i can take of it in a separate PR

sbromberger: Also, since we're changing the signature of `find_root`, perhaps we should

```
@deprecate find_root find_root!
```
?

eulerkochy: Sure! I see you have just committed that change.

sbromberger: yup. I think everything's ready to go.

eulerkochy: cool. let me take a quick glance.

sbromberger:
Would you please tag a new version after (if) this gets merged? Thank you. ❤️

eulerkochy:
@sbromberger, I have released `DataStructures v0.17.11`. Hope you can now continue your PR without any further hiccups. :heart:

sbromberger:
Thanks for all your help, @eulerkochy !!

mlubin:
I have a minor complaint. Deprecating `find_root` in a patch release means, in this case, that JuMP users are now getting deprecation warnings. This is a poor user experience, and bad for the whole ecosystem because if it happens repeatedly we'll have to fix our `DataStructures` dependency to an exact patch release.

eulerkochy:
@mlubin sorry for that. I should list out the reason for updating `find_root` to `find_root!`. The function modifies internally `parents` due to the use of path compression. So, I thought it would be better to maintain the convention of `!`. use the patch release `v0.17.10`, there you wont be facing this issue. Sorry for the bad experience tho'.

sbromberger:
@mlubin we're struggling with this in LightGraphs as well. This is arguably a bugfix - a mutating function did not follow convention - but without the deprecation the bugfix would be a breaking change. With a passthrough nobody would fix it and it would require another release to introduce a deprecation.

Furthermore, I justified it by realizing that below v1.0.0 SemVer rules don't apply. :)

mlubin:
I understand the reasons for the change, but I think it should have a triggered a 0.18 release.

> use the patch release v0.17.10

We can update the registry to set JuMP to depend on 0.17.10, but that will inevitably lead to users having version conflicts with other packages that might need a later release (i.e., any packages that have been updated to use `find_root!`).

> I justified it by realizing that below v1.0.0 SemVer rules don't apply. :)

That's true, but the Julia package manager acts as if the SemVer rules for patch releases do apply when it resolves compatible versions. It would be a pretty bad user experience in terms of version conflicts if we had to fix JuMP's dependencies to exact patch releases out of fear that new patch releases might have API changes. A deprecation is an API change.

oxinabox:
> I have a minor complaint. Deprecating find_root in a patch release means, in this case, that JuMP users are now getting deprecation warnings. This is a poor user experience, and bad for the whole ecosystem because if it happens repeatedly we'll have to fix our DataStructures dependency to an exact patch release.

This true.

We should think about this longer term and make a policy around this.
It could be that we should something like only add deprecations shortly before a breaking release that removes it?

Hopefully soon we will be up to 1.0 and this can mostly stop.

@mlubin  would you be able to open an issue to discuss this rather than continuing via comment on this PR?

sbromberger:
Should we revert the deprecation and make it a pass-through (i.e., `const find_root = find_root!`)?

mlubin:
@oxinabox Do you mean an issue on DataStructures.jl or a broader discussion somewhere?

ccoffrin:
FWIW, I have observed the following impact on some of my work with some JuMP models.

DataStructures v0.17.10 ⚲
model build time: `1.349736 seconds (3.55 M allocations: 415.578 MiB, 6.39% gc time)`

DataStructures v0.17.11 ⚲
model build time: `16.691958 seconds (10.46 M allocations: 1.302 GiB, 4.30% gc time)`

With v0.17.11 I am seeing the deprecation warning as Miles mentioned.  I observed that running julia with `--depwarn=no` resolves the slow down issue.

DataStructures v0.17.11 ⚲ and `julia --depwarn=no`
`1.380376 seconds (3.55 M allocations: 415.578 MiB, 5.96% gc time)`

PR #600, opened 2020-03-30
Update disjoint_set.jl

sbromberger:
Corrected docstring for `find_root` (non-mutating).


eulerkochy:
I don't think, it's correct. It should stay the way it is. `find_root!` implements path-compression, so it changes the value of `parents`, and therefore the `!` should stay.

PR #602, opened 2020-04-02
prefixsum for non-positive index in FenwickTree

awestover:
I think that `prefixsum(ft, x)` should give `0` when `x<1` instead of an ArgumentError, just like how `sum([12,3,4,5,6][1:0]) = 0`.
In particular I was editing some code, and I assumed that `sum(my_array[1:x]) == prefixsum(FenwickTree(my_array), x)` would always be `true`. I was surprised that it wasn't `true` for all `x`, and think that this should be changed.

### DIFF
--- a/src/fenwick.jl
+++ b/src/fenwick.jl
@@ -84,8 +84,6 @@ julia> prefixsum(f, 3)
 function prefixsum(ft::FenwickTree{T}, ind::Integer) where T
     sum = zero(T)
     ind < 1 && return sum
-        return 0
-    end
     i = ind
     n = ft.n
     @boundscheck 1 <= i <= n || throw(ArgumentError("$i should be in between 1 and $n"))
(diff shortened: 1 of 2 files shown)


eulerkochy:
cool! I would certainly say that since `sum(my_array[1:x])` allows `x` to be `0`, I'm definitely in for this change. Nice catch btw!

src/fenwick.jl
```diff
@@ -83,6 +83,9 @@ julia> prefixsum(f, 3)
 """
 function prefixsum(ft::FenwickTree{T}, ind::Integer) where T
     sum = zero(T)
+    if ind < 1
```
eulerkochy, line 86:
```suggestion
    ind < 1 && return 0
```

eulerkochy, line 86:
```suggestion
    ind < 1 && return sum
```

eulerkochy: just a small change, to ensure the return types match. And concise syntax!
```julia
julia> sum(rand(5)[1:0])
0.0

julia> sum(rand(Int, 5)[1:0])
0
```

PR #603, opened 2020-04-03
reverse deprecation of find_root, make const

sbromberger:


mlubin:
LGTM, thanks for the fix!

sbromberger:
There were several ways of doing this, but I chose one that should be the least intrusive and the easiest to properly deprecate in the future.

sbromberger:
ping @oxinabox @eulerkochy - this is intended to ameliorate the issues with JuMP.

mlubin:
Is there a plan to merge and tag this or should we figure out our own workaround in JuMP? https://github.com/JuliaLang/julia/pull/35362 is a good long-term solution but doesn't fix the immediate issue.

src/DataStructures.jl
```diff
@@ -69,6 +69,7 @@ module DataStructures
     include("accumulator.jl")
     include("classified_collections.jl")
     include("disjoint_set.jl")
+    const find_root = find_root!  # remove when deprecating find_root. See deprecations.jl
```
eulerkochy, line 72: just put this in `disjoint_sets.jl`

oxinabox:
I think this is the only way forward.
I intensely do not love this fact.

What is our exit plan? 
- Option 1)  not  deprecate this until v0.18, and not remove it to v0.19?
- Option 2) remove without deprecation in 0.18
- Option 3)  hold-off deprecating it til wherever in JuMP land updates to use new version, then we deprecate it properly, in a 0.17.x release.
- Option 4) deprecate in last release before we make release of 0.18



I am going to keep out of this  statements about how SemVer is *supposed* to work from this discussion, to find the practical solution.
I would request others do the same, and refrain from statements e.g. about if this was a breaking change or not.

mlubin:
You could consider a slight variant of option 2: remove it in 0.18 with a helpful error message instead of a deprecation warning.

src/DataStructures.jl
oxinabox: I think i would put this in `deprecations.jl`
Because that is the file I check everytime I am about to make a minor release.

src/deprecations.jl
```diff
@@ -1,4 +1,4 @@
 @deprecate front(x) first(x)
 @deprecate back(x) last(x)
 @deprecate top(x) first(x)
-@deprecate find_root find_root!   # 2020-03-31
+# @deprecate find_root find_root!   # 2020-03-31 - reimplement with new minor version
```
oxinabox, line 4:
```suggestion
# @deprecate find_root find_root!   # 2020-03-31 - deprecate in v0.18, or when Julia 1.5 is released.
```

oxinabox:
Ok, my conclusion is to either: deprecate it in 0.18, or once julia 1.5 is released, which ever comes out first.

so I think I am am happy with this. 
a few suggestions then we can do it.

oxinabox:
closed in #604

PR #604, opened 2020-04-12
Restore find_root

ccoffrin:
@oxinabox @eulerkochy - I believe this addresses the change requests from #603.

If we could merge and tag this ASAP that would be much appreciated.  Some of my production code has been 10x slower since this change took effect.

CC @mlubin, @sbromberger

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -59,7 +59,6 @@ module DataStructures
 
     export findkey
 
-    include("deprecations.jl")
     include("delegate.jl")
 
     include("deque.jl")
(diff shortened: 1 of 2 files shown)


src/DataStructures.jl
```diff
@@ -29,7 +29,7 @@ module DataStructures
     export ClassifiedCollections
     export classified_lists, classified_sets, classified_counters
 
-    export IntDisjointSets, DisjointSets, num_groups, find_root!, in_same_set, root_union!
+    export IntDisjointSets, DisjointSets, num_groups, find_root, find_root!, in_same_set, root_union!
```
oxinabox, line 32:
```suggestion
    export IntDisjointSets, DisjointSets, num_groups, find_root!, in_same_set, root_union!
```

ccoffrin:
Thanks!

PR #606, opened 2020-04-17
stack and queue helpers

harryscholes:
We have `deque`. This implements `stack` and `queue`.

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -21,7 +21,8 @@ module DataStructures
     export complement, complement!
 
     export Deque, Stack, Queue, CircularDeque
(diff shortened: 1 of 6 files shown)


oxinabox:
Thanks for your interest, however this the opposite of what we should do.

`deque(::Type)` is a left-over from Julia 0.3, where conventions were different, because of restrictions on the use of things like `Foo{T}`  when `T` was not used in the argument.

Tracking down and deprecating all of those is the main thing blocking us from a 1.0 release.

harryscholes:
Ah I see. Makes sense. Should I make a PR deprecating `deque(::Type)`?

oxinabox:
> Ah I see. Makes sense. Should I make a PR deprecating deque(::Type) instead?

That would be very much appreciated

PR #565, opened 2020-01-02
Use first and last, rather than front, back, top etc

krish8484:
Changes made in docs,src and test folders

PR #608, opened 2020-04-20
Fix `first` and `last` function names in `Deque` docstrings

harryscholes:
Function names were changed in #565, but the docstrings for methods on `Deque` weren't updated


oxinabox:
thanks

PR #609, opened 2020-04-20
Deprecate `deque` constructor

harryscholes:
From #606

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.13"
+version = "0.17.14"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -5,4 +5,4 @@
 export find_root
 const find_root = find_root!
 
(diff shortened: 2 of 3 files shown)


harryscholes:
Will do. There are a few other deprecations scattered around the codebase. Do you also want these moved into `deprecations.jl`?

oxinabox:
> Will do. There are a few other deprecations scattered around the codebase. Do you also want these moved into deprecations.jl?

In general yes, not too rushed about fixing old failures, just about avoiding creating new ones.

oxinabox:
I think this is not quite right since it causes an error. is there a missing `where` ?

harryscholes:
Oops! Fixed now :)

oxinabox:
thanks!

giordano:
In v0.17.14 I'm seeing this warning
```
WARNING: Method definition deque(Type{T}) where {T} in module DataStructures at /home/runner/.julia/packages/DataStructures/w35Mo/src/deque.jl:89 overwritten at deprecated.jl:65.
  ** incremental compilation may be fatally broken for this module **
```

harryscholes:
Hmm why is this happening?

harryscholes:
I think this is happening because `@deprecate` exports the deprecated function, but `deque` is still exported in `DataStructures.jl`.

harryscholes:
Actually, no it's not because of that. Discussion continued in #615

PR #612, opened 2020-04-27
Fix out of bounds access in _binary_heap_pop!

### LINKED ISSUE #611: BoundsError in MutableBinaryHeap delete!
If run with ```--check-bounds=yes```, the following throws ```BoundsError```:
```Julia
x = MutableBinaryMinHeap{Int}()
h1 = push!(x, 1)
h2 = push!(x, 2)
delete!(x, h2)
```
[Offending line](https://github.com/JuliaCollections/DataStructures.jl/blob/v0.17.13/src/heaps/mutable_binary_heap.jl#L116)
```Julia
        @inbounds nodes[nd_id] = new_rt = pop!(nodes)
```

Looks to happen when deleting from the end of the heap. The code is safe under the assumption that ```pop!(::Vector)``` doesn't resize underlying storage, but not ideal to have this throw when bounds checks are forced.

ancapdev:
Fixes #611

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.13"
+version = "0.17.14"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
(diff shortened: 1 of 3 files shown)


oxinabox:
cool, I assume you checked this case fails without your fix?
If so, I am happy with it. Thanks!

Could you also bump the rightmost version number (nonbreaking) in the Project.toml so I can tag a release?

ancapdev:
> cool, I assume you checked this case fails without your fix?
> If so, I am happy with it. Thanks!
> 
> Could you also bump the rightmost version number (nonbreaking) in the Project.toml so I can tag a release?

I did, yes. Version bumped for release. Cheers!

oxinabox:
thanks

PR #613, opened 2020-04-27
RobinDict empty! not zeroing hashes bugfix

tozanski:
I believe without this change garbage keys from before `empty!` or garbage keys can be found in the dictionary which should be the case

### DIFF
--- a/src/robin_dict.jl
+++ b/src/robin_dict.jl
@@ -285,6 +285,7 @@ function empty!(h::RobinDict{K,V}) where {K, V}
     resize!(h.keys, sz)
     resize!(h.vals, sz)
     resize!(h.hashes, sz)
+    fill!(h.hashes, 0)
     h.count = 0
     h.idxfloor = 0
     return h
--- a/test/test_robin_dict.jl
+++ b/test/test_robin_dict.jl
@@ -330,6 +330,9 @@ end
     @test h.count == 0
     @test h.idxfloor == 0
     @test length(h.hashes) == length(h.keys) == length(h.vals) == length0
+    for i=-1000:1000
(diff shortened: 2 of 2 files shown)


eulerkochy:
Hey! Can you explain your thought process a bit more elaborately ? 'Cuz I am unable to fathom the purpose of `fill`ing `hashes` with `0` or infact anything for that sake. Only after an successful `rh_insert`, the `hashes` table fills up with meaningful data. Letting it stay garbage was a choice I purposefully did while coding this up.

tozanski:
The following code
```using DataStructures
dict = RobinDict{Int, Int}()

push!(dict, 2=>3)
haskey(dict, 1) |> display
haskey(dict, 2) |> display
empty!(dict)
haskey(dict, 1) |> display
haskey(dict, 2) |> display```
Produces unexpected output
```false
true
false
true```

**Comment by eulerkochy** [comment]
Cool ! Yes, it shouldn't be the case that `haskey` return `true` for any key!
Nice catch, btw ! Sharp eyes 🦅 !
Thanks for your contribution 😸

File: src/robin_dict.jl [hunk 1]
```diff
@@ -285,6 +285,7 @@ function empty!(h::RobinDict{K,V}) where {K, V}
     resize!(h.keys, sz)
     resize!(h.vals, sz)
     resize!(h.hashes, sz)
+    fill!(h.hashes, 0)
```
eulerkochy, line 288: Add one/two test for this change.

eulerkochy:
This was a very critical bug ! Thanks a ton @tozanski for correcting it !

tozanski:
Thanks, I'm glad I could help

PR #615, opened 2020-04-29
Remove deque export

### LINKED ISSUE #614: deque method definition overwritten
In v0.17.14 I'm seeing this warning
```
WARNING: Method definition deque(Type{T}) where {T} in module DataStructures at /home/runner/.julia/packages/DataStructures/w35Mo/src/deque.jl:89 overwritten at deprecated.jl:65.
  ** incremental compilation may be fatally broken for this module **
```

_Originally posted by @giordano in https://github.com/JuliaCollections/DataStructures.jl/pull/609#issuecomment-621216809_

I think this is happening because @deprecate exports the deprecated function, but deque is still exported in DataStructures.jl.

harryscholes:
Fixes #614 

Ref #609

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.14"
+version = "0.17.15"
 
 [deps]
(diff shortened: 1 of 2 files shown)


harryscholes:
This was a hasty fix on my part that seemed to work on my local machine, but it didn't actually work. Please ignore

PR #617, opened 2020-05-03
corrected description for CircularDeque push! function

mantasnaris:
noticed that the description for push! and pushfirst! were duplicated


eulerkochy:
Nice catch ! Thanks for your contribution :smiley_cat:

PR #618, opened 2020-05-11
dibit_vector

sbromberger:
After discussions with @eulerkochy I would like to merge this into DataStructures instead of keeping it as a separate package.

`DiBitVector` is a memory- and cache-friendly representation of values that are 2 bits wide. This has practical benefits to bioinformatics and graph analysis. For large vectors, use of a `DiBitVector` results in improved performance (due to cache locality) and 75% memory savings over a `Vector{UInt8}` (which is what we're currently using).

The selfish reason I have for asking for this to be incorporated into DataStructures.jl is because we already have DataStructures as a dependency for LightGraphs, and I'd like to use DiBitVector in our upcoming 2.0 version – but I don't want to add another dependency to LG, and I don't want to implement this directly in LG.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.15"
+version = "0.17.16"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.16"
+version = "0.17.17"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
--- a/docs/src/dibit_vector.md
+++ b/docs/src/dibit_vector.md
@@ -1,0 +1,47 @@
+```@meta
+DocTestSetup = :(using DataStructures)
+```
+
+# DiBitVector
+
+`DiBitVector` provides a memory-efficient vector of elements that represent four different values from `0` to `3`. This structure is comparable to a `SparseArrays.BitVector` in its performance and memory characteristics.
+
+Examples:
+
+```jldoctest
+julia> v = DiBitVector(4, 0)
+4-element DiBitVector:
+ 0x00
+ 0x00
+ 0x00
+ 0x00
+
+julia> w = DiBitVector(4, 2)
+4-element DiBitVector:
+ 0x02
+ 0x02
+ 0x02
+ 0x02
+
+julia> v[1] = 2
+2
+
+julia> v[2:4] .= 2
+3-element view(::DiBitVector, 2:4) with eltype UInt8:
+ 0x02
+ 0x02
+ 0x02
+
+julia> v == w
+true
+
+julia> pop!(v)
+0x02
+
+julia> length(v)
+3
+```
+
+```@meta
+DocTestSetup = nothing
+```
--- a/src/dibit_vector.jl
+++ b/src/dibit_vector.jl
@@ -1,0 +1,69 @@
+"""
+    DiBitVector <: AbstractVector{UInt8}
+
+A bitvector whose elements are two bits wide, allowing
+storage of integer values between 0 and 3. Optimized
+for performance and memory savings.
+"""
+mutable struct DiBitVector <: AbstractVector{UInt8}
+    data::Vector{UInt64}
+    len::UInt
+
+    function DiBitVector(n::Integer, v::Integer)
+        if !(Int(v) in 0:3)
+            throw(ArgumentError("v must be in 0:3"))
+        end
+        fv = (0x0000000000000000, 0x5555555555555555,
+        0xaaaaaaaaaaaaaaaa, 0xffffffffffffffff)[v + 1]
+        vec = Vector{UInt64}(undef, cld(n, 32))
+        fill!(vec, fv)
+        return new(vec, n % UInt)
+    end
+end
+
+@inline checkbounds(D::DiBitVector, n::Integer) =  0 < n ≤ length(D.data) << 5 || throw(BoundsError(D, n))
+
+DiBitVector(n::Integer) = DiBitVector(n, 0)
+DiBitVector() = DiBitVector(0, 0)
+
+@inline Base.length(x::DiBitVector) = x.len % Int
+@inline Base.size(x::DiBitVector) = (length(x),)
+
+@inline index(n::Integer) = ((n-1) >>> 5) + 1
+@inline offset(n::Integer) = ((n-1) << 1) & 63
+
+@inline function Base.getindex(x::DiBitVector, i::Int)
+    @boundscheck checkbounds(x, i)
+    return UInt8((@inbounds x.data[index(i)] >>> offset(i)) & 3)
+end
+
+@inline function unsafe_setindex!(x::DiBitVector, v::UInt64, i::Int)
+    bits = @inbounds x.data[index(i)]
+    bits &= ~(UInt(3) << offset(i))
+    bits |= convert(UInt64, v) << offset(i)
+    @inbounds x.data[index(i)] = bits
+end
+    
+@inline function Base.setindex!(x::DiBitVector, v::Integer, i::Int)
+    v & 3 == v || throw(DomainError("Can only contain 0:3 (tried $v)"))
+    @boundscheck checkbounds(x, i)
+    unsafe_setindex!(x, convert(UInt64, v), i)
+end
+
+@inline function Base.push!(x::DiBitVector, v::Integer)
+    len = length(x)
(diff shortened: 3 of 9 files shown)


eulerkochy:
@sbromberger please note the CI failures. The PR looks good otherwise.

sbromberger:
The failures seem to be in priorityqueue. Am I missing something?

https://travis-ci.org/github/JuliaCollections/DataStructures.jl/jobs/685830700#L333

eulerkochy:
Can you close and re-open the PR? Let the CI run again. Strange though.

sbromberger:
Closing and reopening doesn't trigger travis. Can you do it manually?

eulerkochy:
@sbromberger check the issue with the Appveyor CI run.
[It errors on x86 builds](https://ci.appveyor.com/project/kmsquire/datastructures-jl/builds/32878142/job/a1n6nwbi719l9b9d#L146)

sbromberger:
I'm stuck. I don't have a 32-bit system to test on.

sbromberger:
Closing this out in frustration. I can't keep trying to guess how to make this work on 32-bit systems. Sorry.

src/dibit_vector.jl
```diff
@@ -0,0 +1,74 @@
...
+    len::UInt
+
+    function DiBitVector(n::Integer, v::Integer)
+        if !(Int(v) in 0:3)
+            throw(ArgumentError("v must be in 0:3"))
+        end
+        fv = (0x0000000000000000, 0x5555555555555555,
+        0xaaaaaaaaaaaaaaaa, 0xffffffffffffffff)[v + 1]
+        vec = Vector{UInt}(undef, cld(n, 32))
+        fill!(vec, fv)
```
rfourquet, line 24: This can't work if `UInt == UInt32`, as e.g. `0x5555555555555555` can't be losslessly converted to `UInt32`. You could use `fv % UInt`, but I would personally just use `UInt64` everywhere for simplicity, even if this means it's less efficient on 32-bits machines. This is e.g. what `BitSet` and `BitArray` use.

eulerkochy:
@sbromberger I'm reopening and let's pull it out together. There HAS to be a way.

src/dibit_vector.jl
eulerkochy: I guess the problem lies in how do define it here @sbromberger . Can you split it case-wise. Like one for `UInt32` and other for `UInt64`.

sbromberger: OK, I think I did it.

sbromberger: Actually, I like the idea of just making things UInt64.

eulerkochy:
@sbromberger do we need a version bump for this right now?

sbromberger:
> do we need a version bump for this right now?

It'd be nice so I could start using it in LightGraphs....

eulerkochy:
@sbromberger could you bump the version. I'll merge and register.

docs/src/dibit_vector.md
```diff
@@ -0,0 +1,47 @@
+```@meta
+DocTestSetup = :(using DataStructures)
+```
+
+# DiBitVector
+
+`DiBitVector` provides a memory-efficient vector of elements that represent four different values from `0` to `3`. This structure is comparable to a `SparseArrays.BitVector` in its performance and memory characteristics.
```
rfourquet, line 7:
```suggestion
`DiBitVector` provides a memory-efficient vector of elements that represent four different values from `0` to `3`. This structure is comparable to a `BitVector` in its performance and memory characteristics.
```

src/dibit_vector.jl
```diff
@@ -0,0 +1,70 @@
...
+
+A bitvector whose elements are two bits wide, allowing
+storage of integer values between 0 and 3. Optimized
+for performance and memory savings.
+"""
+mutable struct DiBitVector <: AbstractVector{UInt8}
+    data::Vector{UInt64}
+    len::UInt
+
+    function DiBitVector(n::Integer, v::Integer)
```
rfourquet, line 12: I think a docstring for the constructors is warranted, in particular this one which is unlike what is available for other `AbstractVector`s. By the way, it might even appear backward with some of `Array` constructors, like in `Array{Union{Int,Nothing}}(nothing, 2)` where the initializer is given before the dimension.

src/dibit_vector.jl
```diff
@@ -0,0 +1,70 @@
+"""
+    DiBitVector <: AbstractVector{UInt8}
+
+A bitvector whose elements are two bits wide, allowing
+storage of integer values between 0 and 3. Optimized
+for performance and memory savings.
+"""
+mutable struct DiBitVector <: AbstractVector{UInt8}
```
rfourquet, line 8: What about `<: AbstractVector{Int8}` ? (not that I prefer `Int8`, but was just curious why you chose `UInt8`).

sbromberger: I chose `UInt8` because of my use case: a coloring from 0 to 3.  I suppose we could change it to `Int8` for the folks who want -2 to +1... but I don't know of any use cases for that offhand.

src/dibit_vector.jl
sbromberger: How do you write a docstring for an inner constructor? Do you just put it in the main struct docstring? (I tried doing it above the inner and it doesn't show up.)

sbromberger: Got it.

```
help?> DiBitVector
search: DiBitVector
  DiBitVector(n::Integer, v::Integer)
  Create a DiBitVector with n elements preinitialized to a value v from 0 to 3, inclusive.
  A DiBitVector is a vector whose elements are two bits wide, allowing storage of integer values between 0 and 3. This structure is optimized for
  performance and memory savings for large numbers of elements.
  ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  DiBitVector(n::Integer)
  Create a DiBitVector with n elements set to 0.
```

src/dibit_vector.jl
rfourquet: Ok, but I was more thinking `0` to `3` (it would just be a matter of adding `% Int8` in `getindex`.

src/dibit_vector.jl
rfourquet: Thinking about this a bit more... How likely it is that this will be extended to `DiBiArray` with multiple dimensions? In this case, putting the dimensions last would be more necessary (e.g. `DiBiArray(value, d1, d2)`, like for some arrays constructors, and `fill`, and `rand` etc.) So in any I would suggest to switch the order here in `DiBiVector` for consistency (but it's just my outsider's opinion).

src/dibit_vector.jl
sbromberger: My thought was because it is a subtype of `AbstractVector{UInt8}` it should output `UInt8` as the eltype.

I made it a subtype of `UInt8` because, essentially, that's what it is. (well, 8 of them packed into a `UInt64`, but still - unsigned.)

src/dibit_vector.jl
sbromberger: I have no plans on extending this to a matrix or n-dimentions. That said, I'm not averse to changing the constructor syntax. It means that we will have an inconsistency in position between
```
d = DiBitVector(2, 10)   # fill a 10-element DBV with values of 2
```
and
```
d = DiBitVector(10)  # zero-fill a 10-element DBV
```
but if that's the consensus I'll change it.

rfourquet: Yes that's not ideal either. I'm not aware of whether there is a common practice in the package ecosystem for this case. If `DiBitVector(10)` is kept, I guess I prefer your current version for specifying the value, i.e. `DiBiVector(10, 2)`.

src/dibit_vector.jl
rfourquet: Sure if it's a subtype of `AbstractVector{UInt8}`, it must return `UInt8`. But this would work equally well with `Int8`. I guess it depends on the use-case. Setting that as a type parameter would be an option (`mutable struct DiBitVector{T<:Integer} <: AbstractVector{T}`), but might be overkill, I don't know.

src/dibit_vector.jl
sbromberger: `DiBitVector(10)` is useful; and I might even propose at some point that it allocate `undef` memory (since we always have `DiBitVector(10, 0)` to get zeroed memory).

Anyway, I'll mark this conversation resolved. We can always change things later based on wider feedback.

src/dibit_vector.jl
sbromberger: I think it's overkill. `BitVector` is `<: AbstractArray{Bool, 1}`, which makes sense. The smallest valid type for a `DiBitVector` is `UInt8`. (If we had nybbles, I'd use that.) `Int8` is the same width but is ambiguous: do we range from -2 to +1, or do we range from 0 to 3?

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.15"
+version = "0.17.16"
```
eulerkochy, line 3:
```suggestion
version = "0.17.17"
```

sbromberger:
Thanks for all the feedback and help in getting this merged.

PR #624, opened 2020-05-14
Remove Primes from dependencies

nico202:
Primes does not seem to be used, so remove it from dependencies

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.17.15"
+version = "0.17.16"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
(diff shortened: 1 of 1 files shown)


kmsquire:
Hi @nico202, thanks for this PR. However, this isn’t quite the right fix.

You’re correct that Primes.jl isn’t used by the main package, but it is used in the tests. I believe the correct fix would be to create a `Package.toml` file inside the tests directory, and remove `tests/REQUIRE`. The latter file used to handle this extra dependency by itself, but that is probably not true now.

Do you think you can update this PR to do that?

nico202:
Do you know where primes is used in tests? I cannot find any reference

```bash
DataStructures.jl$ rg Primes
Changelog.md
69:     * Tests: primes moved from Base to Primes.jl for

Project.toml
14:Primes = "27ebfcd6-29c5-5fa9-bf4b-fb8fc14df3ae"
20:test = ["Test", "Primes", "Random", "Serialization"]

test/REQUIRE
1:Primes 0.1.1
```

oxinabox:
This change has now been mafe by another PR

PR #631, opened 2020-06-05
Base.IntSet -> Base.BitSet

cmcaine:
Base doesn't have an IntSet any more. Reading the description, sounds like you're describing Base.BitSet, so I've updated the reference.

If this is correct, I'll check for other instances of this to replace, lmk.


oxinabox:
Yep, it was renamed in julia 0.7

cmcaine:
I just git grepped and this looks like the only instance. I think this is done and can be merged now.

rfourquet:
Is the only reason to keep this data-structure the `complement` feature?
Also, note that this is not a fully drop-in replacement for `Base.BitSet` because of negative integers (I'm not arguing against the renaming though!)

oxinabox:
> Is the only reason to keep this data-structure the `complement` feature?
> Also, note that this is not a fully drop-in replacement for `Base.BitSet` because of negative integers (I'm not arguing against the renaming though!)

no idea of the history here.
If someone wants to look into it and see if we can just remove this that would be good but is beyond the scope of this PR

PR #636, opened 2020-06-24
remove redundant iterate method

KristofferC:
the fallback method seems to work just as well.
This method also causes some invalidations so good to get rid of it

cc @timholy (playing around with new SnoopCompile)


oxinabox:
thanks

PR #637, opened 2020-06-25
Add Base.copy, Base.copymutable for Sorted containers

NHDaly:
Per our notes in https://github.com/JuliaCollections/DataStructures.jl/pull/571#issuecomment-594125382, the sorted containers currently don't implement `Base.copy(x)`, which is "Strongly Recommended," but not "Required."

However, they do all already provide `packcopy(m)`, and as far as i can tell, that seems like a perfectly reasonable implementation of copy: it returns a new container with all the same elements but not necessarily the same structure. ✔️ 

So this PR simply aliases `copy` and `copymutable` to `packcopy`. Does that sound good? :)
Thanks!
~Nathan

### DIFF
--- a/test/test_sorted_containers.jl
+++ b/test/test_sorted_containers.jl
@@ -1717,4 +1717,10 @@ end
     @test pop!(s,50, nothing) == nothing
     @test isempty(s)
 
+    # Test AbstractSet/AbstractDict interface
+    for m in [SortedSet([1,2]), SortedDict(1=>2, 2=>3), SortedMultiDict(1=>2, 1=>3)]
+        # copy()
+        @test isequal(copy(m), m)
+        @test isequal(Base.copymutable(m), m)
+    end
 end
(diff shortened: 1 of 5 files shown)


NHDaly:
Request for review: @oxinabox

test/test_sorted_containers.jl
```diff
@@ -1717,4 +1717,10 @@ end
     @test pop!(s,50, nothing) == nothing
     @test isempty(s)
 
+    # Test AbstractSet/AbstractDict interface
+    for m in [SortedSet([1,2]), SortedDict(1=>2, 2=>3), SortedMultiDict(1=>2, 1=>3)]
+        # copy()
+        @test isequal(copy(m), m)
+        @test isequal(Base.copymutable(m), m)
```
oxinabox, line 1724: I think we want to also check the type on these.
Since dictionaries are equal based on content.
But an important thing about `copymutable` is that they preserve the type,

oxinabox:
can you also bump the version number?

NHDaly:
Thanks Lyndon; done with both! :)

test/test_sorted_containers.jl
NHDaly: Good call. Done.

NHDaly:
:) Thanks Lyndon!

PR #638, opened 2020-06-26
[WIP] Red Black Tree

eulerkochy:
Part of GSoC'2020 evaluation.
This PR is for implementation of Red Black Tree in Julia. Mainly, to be used as a `tag` for sorted containers. C++ uses `rb_tree` for it's `set` and `map`.

The only thought is whether to `export` it, or not, because `src/balanced_tree.jl` isn't exported.

- [ ] Implement delete!
- [ ] Write tests
- [ ] Documentation

### DIFF
--- a/src/red_black_tree.jl
+++ b/src/red_black_tree.jl
@@ -1,0 +1,173 @@
+
+# color is true if it's a Red Node, else it's false
+mutable struct RBTreeNode{K}
+	color::Bool 
+	data::K
+	leftChild::Union{Nothing, RBTreeNode{K}}
+	rightChild::Union{Nothing, RBTreeNode{K}}
+	parent::Union{Nothing, RBTreeNode{K}}
+
+	RBTreeNode{K}() where K = new{K}(true, nothing, nothing, nothing, nothing)
+	
+	RBTreeNode{K}(d::K) where K = new{K}(true, d, nothing, nothing, nothing)
+end
+
+RBTreeNode() = RBTreeNode{Any}()
+RBTreeNode(d) = RBTreeNode{Any}(d)
+
+mutable struct RBTree{K}
+	root::RBTreeNode{K}
+
+	RBTree{K}() where K = new{K}(RBTreeNode{K}())
+end
+
+RBTree() = RBTree{Any}()
+
+function insert(root::Union{Nothing, RBTreeNode}, node::RBTreeNode)
+	if root == nothing || root.data == nothing
+		return node
+	else
+		if (node.data > root.data)
+			root.rightChild = insert(root.rightChild, node)
+			root.rightChild.parent = root
+		else
+			root.leftChild = insert(root.leftChild, node)
+			root.leftChild.parent = root
+		end
+		return root
+	end
+end
+
+function left_rotate!(root::Union{Nothing, RBTreeNode}, node_x::RBTreeNode)
+	node_y = node_x.rightChild
+	node_x.rightChild = node_y.leftChild
+	if node_x.rightChild != nothing
+		node_x.rightChild.parent = node_x
+	end
+	node_y.parent = node_x.parent
+	if (node_x.parent == nothing)
+		root = node_y
+	elseif (node_x == node_x.parent.leftChild)
+		node_x.parent.leftChild = node_y
+	else
+		node_x.parent.rightChild = node_y
+	end
+	node_y.leftChild = node_x
+	node_x.parent = node_y
+end	
+
+function right_rotate!(root::Union{Nothing, RBTreeNode}, node_x::RBTreeNode)
+	node_y = node_x.leftChild
+	node_x.leftChild = node_y.rightChild
+	if node_x.leftChild != nothing
+		node_x.leftChild.parent = node_x
+	end
+	node_y.parent = node_x.parent
+	if (node_x.parent == nothing)
+		root = node_y
+	elseif (node_x == node_x.parent.leftChild)
+		node_x.parent.leftChild = node_y
+	else
+		node_x.parent.rightChild = node_y
+	end
+	node_y.rightChild = node_x
+	node_x.parent = node_y
+end	
+
+function balance!(root::Union{Nothing, RBTreeNode}, node::RBTreeNode)
+	parent = nothing
+	grand_parent = nothing
+	# for root node, we need to change the color to black
+	# other nodes, we need to maintain the property such that
+	# no two adjacent nodes are red in color
+	while (node != root && node.color && node.parent.color)
+		parent = node.parent
+		grand_parent = parent.parent
+
+		# parent is the leftChild of grand_parent
+		if (parent == grand_parent.leftChild)
+			uncle = grand_parent.rightChild
+
+			# uncle is red in color
+			if (uncle != nothing && uncle.color)
+				grand_parent.color = true
+				parent.color = false
+				uncle.color = false
+				node = grand_parent
+			# uncle is black in color
+			else 
+				# node is rightChild of it's parent
+				if (node == parent.rightChild)
+					left_rotate!(root, parent)
+					node = parent
+					parent = node.parent
+				end
+				# node is leftChild of it's parent
+				right_rotate!(root, grand_parent)
+				parent.color, grand_parent.color = grand_parent.color, parent.color
+				node = parent
+			end
+		# parent is the rightChild of grand_parent
+		else 
+			uncle = grand_parent.leftChild
+
+			# uncle is red in color
+			if (uncle != nothing && uncle.color)
+				grand_parent.color = true
+				parent.color = false
+				uncle.color = false
+				node = grand_parent
+			# uncle is black in color
+			else 
+				# node is leftChild of it's parent
+				if (node == parent.leftChild)
+					right_rotate!(root, parent)
+					node = parent
+					parent = node.parent
+				end
+				# node is rightChild of it's parent
+				left_rotate!(root, grand_parent)
+				parent.color, grand_parent.color = grand_parent.color, parent.color
+				node = parent
+			end
+		end
+	end
+	root.color = false
+end
+
+function insert!(tree::RBTree{K}, d::K) where K
+	node = RBTreeNode(d)
+	tree.root = insert(tree.root, node)
+	balance!(tree.root, node)
+	return tree
+end
+
+function print_tree_inorder(node::Union{Nothing, RBTreeNode})
+	if (node != nothing)
+		print_tree_inorder(node.leftChild)
+		println("value -> ", node.data)
+		print_tree_inorder(node.rightChild)
+	end
+end
+
+function print_tree_levelorder(node::Union{Nothing, RBTreeNode{K}}) where K
+	level = Dict{Int, Vector{K}}()
+	function traverse_tree(node::Union{Nothing, RBTreeNode}, lvl = 1)
+		if (node != nothing)
+			if (haskey(level, lvl))
+				push!(level[lvl], node.data)
+			else
+				level[lvl] = K[node.data]
+			end
+			traverse_tree(node.leftChild, lvl+1)
+			traverse_tree(node.rightChild, lvl+1)
+		end
+	end
+
+	traverse_tree(node)
+
+	for (k, v) in level
+		println("Level -> ", k)
+		println(v)
+	end
+end
--- a/src/red_black_tree.jl
+++ b/src/red_black_tree.jl
@@ -23,6 +23,29 @@ end
 
 RBTree() = RBTree{Any}()
 
+function search_node(tree::RBTree{K}, d::K) where K
+	node = tree.root
+	(node == nothing) && return nothing
+	while (node != nothing)
+		if (d < node.data)
+			(node.leftChild == nothing) && break
+			node = node.leftChild
+		elseif (d > node.data)
+			(node.rightChild == nothing) && break
+			node = node.rightChild
+		else
+			break
+		end
+	end
+	return node
+end
+
+function search_key(tree::RBTree{K}, d::K) where K 
+	node = search_node(tree, d)
+	(node == nothing) && return false
+	return (node.data == d)
+end
+
 function insert(root::Union{Nothing, RBTreeNode}, node::RBTreeNode)
 	if root == nothing || root.data == nothing
 		return node
@@ -38,7 +61,7 @@ function insert(root::Union{Nothing, RBTreeNode}, node::RBTreeNode)
 	end
 end
 
-function left_rotate!(root::Union{Nothing, RBTreeNode}, node_x::RBTreeNode)
+function left_rotate!(tree::RBTree, node_x::RBTreeNode)
 	node_y = node_x.rightChild
 	node_x.rightChild = node_y.leftChild
 	if node_x.rightChild != nothing
@@ -46,7 +69,7 @@ function left_rotate!(root::Union{Nothing, RBTreeNode}, node_x::RBTreeNode)
 	end
 	node_y.parent = node_x.parent
 	if (node_x.parent == nothing)
-		root = node_y
+		tree.root = node_y
 	elseif (node_x == node_x.parent.leftChild)
 		node_x.parent.leftChild = node_y
 	else
@@ -56,7 +79,7 @@ function left_rotate!(root::Union{Nothing, RBTreeNode}, node_x::RBTreeNode)
 	node_x.parent = node_y
 end	
 
-function right_rotate!(root::Union{Nothing, RBTreeNode}, node_x::RBTreeNode)
+function right_rotate!(tree::RBTree, node_x::RBTreeNode)
 	node_y = node_x.leftChild
 	node_x.leftChild = node_y.rightChild
 	if node_x.leftChild != nothing
@@ -64,7 +87,7 @@ function right_rotate!(root::Union{Nothing, RBTreeNode}, node_x::RBTreeNode)
 	end
 	node_y.parent = node_x.parent
 	if (node_x.parent == nothing)
-		root = node_y
+		tree.root = node_y
 	elseif (node_x == node_x.parent.leftChild)
 		node_x.parent.leftChild = node_y
 	else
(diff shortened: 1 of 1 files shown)


StephenVavasis:
Years ago, I carried out some simple performance tests of balancedtree23.jl, and it seemed that the bottleneck is cache misses.  A red-black tree would likely not improve on cache-behavior.  Rather, you would need the more complicated B-tree structure of Bender et al.:

https://erikdemaine.org/papers/CacheObliviousBTrees_SICOMP/paper.pdf

eulerkochy:
@StephenVavasis  I have B-Tree as a part of my proposal too and I plan on implementing the cache-friendly version, and yes, the main motivation for that came for Eric Demaine's lecture . I would go through the paper in the meanwhile.

src/red_black_tree.jl
```diff
@@ -0,0 +1,199 @@
...
+					node = parent
+					parent = node.parent
+				end
+				# node is rightChild of it's parent
+				left_rotate!(tree, grand_parent)
+				parent.color, grand_parent.color = grand_parent.color, parent.color
+				node = parent
+			end
+		end
+	end
```
oxinabox, line 160: this file is using tabs instead of spaces.

eulerkochy: resolved in the new commit

eulerkochy:
@timholy can you guide on how to use `print_tree(t)` defined in `AbstractTrees.jl` for printing this tree? I've written a pretty basic one, for debugging purposes. It would be nice, if I can reuse code already written :stuck_out_tongue:

eulerkochy:
@oxinabox it's ready for review. Only a minor thing to be noted is that I'm not planning to export it. It would be used for underlying SortedDict. 
Plan is to work on getting `rb_tree_tag` like structure for the SortedDict API. Hence there is no parameterised initialisation of Value `V` parameter. So that the `RBTree{K, V}` can reflect a key-value pair. It can be added without major hiccups. Waiting for your call on this.

src/red_black_tree.jl
```diff
@@ -0,0 +1,327 @@
...
+
+function create_null_node(K::Type)
+    node = RBTreeNode{K}()
+    node.color = false
+    return node
+end 
+
+mutable struct RBTree{K}
+    root::RBTreeNode{K}
+    Nil::RBTreeNode{K}
```
oxinabox, line 36: why is this in TitleCase ?
```suggestion
    nil::RBTreeNode{K}
```

src/red_black_tree.jl
```diff
@@ -0,0 +1,327 @@
...
+
+Base.:(!=)(x::RBTreeNode{K}, y::RBTreeNode{K}) where K = !isequal(x, y)
+
+function create_null_node(K::Type)
+    node = RBTreeNode{K}()
+    node.color = false
+    return node
+end 
+
+mutable struct RBTree{K}
```
oxinabox, line 34: Should this struct be mutable, or should it just be changed by mutating the root node?

src/red_black_tree.jl
```diff
@@ -0,0 +1,327 @@
...
+RBTreeNode() = RBTreeNode{Any}()
+RBTreeNode(d) = RBTreeNode{Any}(d)
+
+function Base.isequal(x::RBTreeNode{K}, y::RBTreeNode{K})where K
+   return (x.color == y.color && x.data == y.data)
+end
+
+Base.:(==)(x::RBTreeNode{K}, y::RBTreeNode{K}) where K = isequal(x, y)
+
+Base.:(!=)(x::RBTreeNode{K}, y::RBTreeNode{K}) where K = !isequal(x, y)
```
oxinabox, line 26: According to docs, unless handling floating point specially one only should do `==`

> 
> ## isequal(x, y)
> 
>   Similar to ==, except for the treatment of floating point numbers and of missing values. isequal
>   treats all floating-point NaN values as equal to each other, treats -0.0 as unequal to 0.0, and
>   missing as equal to missing. Always returns a Bool value.
> 
>   Implementation
>   ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡
> 
>   The default implementation of isequal calls ==, so a type that does not involve floating-point
>   values generally only needs to define ==.
> 
>   isequal is the comparison function used by hash tables (Dict). isequal(x,y) must imply that
>   hash(x) == hash(y).
> 
>   This typically means that types for which a custom == or isequal method exists must implement a
>   corresponding hash method (and vice versa). Collections typically implement isequal by calling
>   isequal recursively on all contents.
> 
>   Scalar types generally do not need to implement isequal separate from ==, unless they represent
>   floating-point numbers amenable to a more efficient implementation than that provided as a
>   generic fallback (based on isnan, signbit, and ==).
> 
> ## !=(x, y)
>   ≠(x,y)
> 
>   Not-equals comparison operator. Always gives the opposite answer as ==.
> 
>   Implementation
>   ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡
> 
>   New types should generally not implement this, and rely on the fallback definition !=(x,y) =
>   !(x==y) instead.


So one should just implement `==` here.

and one should implement `hash` if ever implementing `==` (or `isequals`).
I suggest an approproate hash is to just return the hash of the `data`.
The colour won't provide enough gain i think. Worst case is if something with each color is put into the same hashtable, and then is still resolved with a single comparison.

src/red_black_tree.jl
```diff
@@ -0,0 +1,327 @@
...
+    function RBTree{K}() where K 
+        new{K}(create_null_node(K), create_null_node(K))
+    end
+end
+
+RBTree() = RBTree{Any}()
+
+function search_node(tree::RBTree{K}, d::K) where K
+    node = tree.root
+    while node != tree.Nil && d != node.data
```
oxinabox, line 47: `node != tree.Nil`
is the same as:
```
(node.color === false && node.data === node.parent===node.left_child === node.right_child === nothing)
```
Potentially instead of storing a explict `Nil` in the tree to compare to we can just define a `is_nil` function that does that check.

Or conversly, we could always make sure the node is a reference to the tree's `Nil`,
and then do `!==` which would only do a pointer equality check.


if we did get rid of the `Nil` element, then we could also get rid of the `Tree` type and just have the `Node` type, since the tree would be described completely by its root.

src/red_black_tree.jl
```diff
@@ -0,0 +1,327 @@
...
+                    node = parent
+                    left_rotate!(tree, node)
+                end
+                # node is leftChild of it's parent
+                node.parent.color = false
+                node.parent.parent.color = true
+                right_rotate!(tree, node.parent.parent)
+            end
+        # parent is the rightChild of grand_parent
+        else
```
oxinabox, line 154: I am unclear as to if this comment belongs to the block following the else or the block preciding it.
(Similar for eom other comments in this section)
If it is the block following then it can be clearer if put on same line.
If not then it should be indented
```suggestion
        else  # parent is the rightChild of grand_parent
```

src/red_black_tree.jl
```diff
@@ -0,0 +1,327 @@
...
+        end
+
+        if d < node.data
+            node = node.leftChild
+        else
+            node = node.rightChild
+        end
+    end
+
+    (z == tree.Nil) && throw(KeyError(d))
```
oxinabox, line 296: `delete!` on dicts for items that are not present does not error in julia.
So maybe it should be the same for trees?

```julia
julia> x = Dict(1=>3)
Dict{Int64,Int64} with 1 entry:
  1 => 3

julia> delete!(x, 2)
Dict{Int64,Int64} with 1 entry:
  1 => 3
```

test/test_red_black_tree.jl
```diff
@@ -0,0 +1,71 @@
...
+    	delete!(t, i)
+    end
+
+    for i = 1:10000
+    	if iseven(i)
+    		@test search_key(t, i)
+    	else
+    		@test !search_key(t, i)
+    		@test_throws KeyError delete!(t, i)
+    	end
```
oxinabox, line 25: the indentation in this bit is all messed up.
it is a mix of 4 spaces, 2 spaces, and tabs

src/red_black_tree.jl
eulerkochy: It can be done. But let's think of it in this way. This file is mostly to be used as an underlying container for `SortedDict` right? Throwing a `KeyError` in the underlying container doesn't hurt, we can catch this and process accordingly.

src/red_black_tree.jl
eulerkochy: I think, it should be `mutable`. Althought the `nil` element is fixed, the `root` keeps on changing.

src/red_black_tree.jl
eulerkochy: The implementation of Red-Black Tree becomes complicated when `T.nil` is not used, especially in cases related to `delete`. Let's keep this for the this now. If I find a good workaround, I'll put a PR later.

test/test_red_black_tree.jl
eulerkochy: Oops.

src/red_black_tree.jl
```diff
@@ -0,0 +1,327 @@
...
+# rightChild has keys which are greater than the node
+# color is true if it's a Red Node, else it's false
+mutable struct RBTreeNode{K}
+    color::Bool 
+    data::Union{K, Nothing}
+    leftChild::Union{Nothing, RBTreeNode{K}}
+    rightChild::Union{Nothing, RBTreeNode{K}}
+    parent::Union{Nothing, RBTreeNode{K}}
+
+    RBTreeNode{K}() where K = new{K}(true, nothing, nothing, nothing, nothing)
```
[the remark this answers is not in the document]
eulerkochy: `search_key` uses the fact that `tree.nil`'s `data` field is `nothing`. So, it is done for the sake of ease of implementation,

src/red_black_tree.jl
oxinabox: Can we fake that by being able to just mutate the content of the root node?

src/red_black_tree.jl
oxinabox: Can we do this then:

> Or conversly, we could always make sure the node is a reference to the tree's Nil,
and then do !== which would only do a pointer equality check.

src/red_black_tree.jl
oxinabox: Consistency is important to make it easier to learn.

Exception catching is expensive.
As you say `SortedDict` would just supress that error and return the `Dict` unmodified anyway.

src/DataStructures.jl
```diff
@@ -113,4 +113,6 @@ module DataStructures
 
     include("dibit_vector.jl")
     include("deprecations.jl")
+
+    include("red_black_tree.jl")
```
oxinabox, line 117: I feel like deprecations being last is a good idea?
```suggestion
        include("red_black_tree.jl")
        
        include("deprecations.jl")
```

src/red_black_tree.jl
```diff
@@ -0,0 +1,380 @@
...
+        else # parent is the rightChild of grand_parent
+            uncle = grand_parent.leftChild
+
+            if (uncle.color) # uncle is red in color
+                grand_parent.color = true
+                parent.color = false
+                uncle.color = false
+                node = grand_parent
+            else  # uncle is black in color
+                if (node == parent.leftChild) # node is leftChild of it's parent
```
oxinabox, line 192:
```suggestion
                if (node == parent.leftChild) # node is leftChild of its parent
```

src/red_black_tree.jl
```diff
@@ -0,0 +1,380 @@
...
+                    node = parent
+                    right_rotate!(tree, node)
+                end
+                # node is rightChild of it's parent
```
oxinabox, line 196:
```suggestion
                # node is rightChild of its parent
```

src/red_black_tree.jl
```diff
@@ -0,0 +1,380 @@
...
+        if (parent == grand_parent.leftChild) # parent is the leftChild of grand_parent
+            uncle = grand_parent.rightChild
+
+            if (uncle.color) # uncle is red in color
+                grand_parent.color = true
+                parent.color = false
+                uncle.color = false
+                node = grand_parent
+            else  # uncle is black in color
+                if (node == parent.rightChild) # node is rightChild of it's parent
```
oxinabox, line 174:
```suggestion
                if (node == parent.rightChild) # node is rightChild of its parent
```

https://www.grammarly.com/blog/its-vs-its/

src/red_black_tree.jl
```diff
@@ -0,0 +1,380 @@
...
+                    node = parent
+                    left_rotate!(tree, node)
+                end
+                # node is leftChild of it's parent
```
oxinabox, line 178:
```suggestion
                # node is leftChild of its parent
```

src/red_black_tree.jl
```diff
@@ -0,0 +1,380 @@
...
+    end
+    v.parent = u.parent
+end
+
+"""
+   minimum_node(tree::RBTree, node::RBTreeNode) 
+
+Returns the RBTreeNode with minimum value in subtree of `node`. 
+"""
+function minimum_node(tree::RBTree, node::RBTreeNode)
```
oxinabox, line 319: Should this overload `Base.minimum` ?
Or does "minimum value" not mean what i think it does?

test/test_red_black_tree.jl
```diff
@@ -0,0 +1,77 @@
+include("../src/red_black_tree.jl")
+@testset "RBTree" begin
+    t = RBTree{Int}()
+    for i = 1:10000
```
oxinabox, line 4: The rest of the package fairky  consistently uses
```suggestion
    for i in 1:10000
```

test/test_red_black_tree.jl
```diff
@@ -0,0 +1,77 @@
...
+        insert!(t, i)
+    end
+    for i = 1:10000
+        @test search_key(t, i)
+    end
+
+    for i = 10001:20000
```
oxinabox, line 11: Why are these numbers so large?
Seems like it will not improve our actually test coverage and will make the tests slower and harder to read if they fail?

src/red_black_tree.jl
```diff
@@ -0,0 +1,380 @@
...
+end
+
+"""
+    search_key(tree, key)
+
+Returns true if `key` is present in the `tree`, else returns false.    
+"""
+search_key(tree, key)
+
+function search_key(tree::RBTree{K}, d::K) where K
```
oxinabox, line 71: Can we use `Base.haskey` instead?

src/red_black_tree.jl
```diff
@@ -0,0 +1,380 @@
...
+RBTree() = RBTree{Any}()
+
+"""
+    search_node(tree, key)
+
+Returns the last visited node, while traversing through in binary-search-tree fashion looking for `key`.
+"""
+search_node(tree, key)
+
+function search_node(tree::RBTree{K}, d::K) where K
```
oxinabox, line 52: Can we use this to provide:
`Base.in` and `Base.getindex`

src/red_black_tree.jl
```diff
@@ -0,0 +1,380 @@
...
+    node = search_node(tree, d)
+    return (node.data == d)
+end
+
+"""
+    insert_node!(tree::RBTree, node::RBTreeNode)
+
+Inserts `node` at proper location by traversing through the `tree` in a binary-search-tree fashion.
+"""
+function insert_node!(tree::RBTree, node::RBTreeNode)
```
oxinabox, line 81: Can we use this to provide `Base.push!(tree, value}` where `value` is some time type we can convert to `K`

test/test_red_black_tree.jl
```diff
@@ -0,0 +1,77 @@
...
+
+    # for handling cases related to insert!
+    t3 = RBTree()
+    for num in nums
+        insert!(t3, num)
+    end
+
+    for i in visited
+        @test search_key(t3, i)
+    end
```
oxinabox, line 75: We should break each section into its own sub-testset.

StephenVavasis:
Could I ask what is the mission for this data structure?  If it is to serve as the substrate for SortedDict, etc., then it appears to be incomplete because it lacks all the machinery for iteration (tokens).  In any case, the author wrote earlier about plans to work on a B-tree data structure.  A B-tree has the possibility of better performance because of better data locality than a red-black tree.

eulerkochy:
@StephenVavasis what do you mean by `machinery for iteration`? I'm working on similar tree-based DS and I've a bunch of PRs lined up. I want to get this right. It would really nice, if you can elaborate a bit more. And the others would fall into place. And yes, I've a branch for `BTree` and am working on it. (I'm stuck on cases related to delete.)

oxinabox:
> Could I ask what is the mission for this data structure?  If it is to serve as the substrate for SortedDict, etc., then it appears to be incomplete because it lacks all the machinery for iteration (tokens).  In any case, the author wrote earlier about plans to work on a B-tree data structure.  A B-tree has the possibility of better performance because of better data locality than a red-black tree.

Primarily, I see it as an end-onto-itself.
Its a canonical data structure taught in most CS undergrad courses.
Simply having it so people can try it out and look at the source is useful.
Maybe for some particular niche applications it is better, maybe not.

I am not inclined to mess with `SortedDict` since that is so well optimised and tested right now.

SyxP:
Perhaps such data structures should stand alone in their own package with emphasis on readability and comments and not just optimised for speed?

StephenVavasis:
@oxinabox already mentioned in a different response that there is no plan to replace the current 2-3 tree substrate of the sorted containers with red-black trees so perhaps this is not so important.

But if you wish to pursue it: tokens and semitokens in the existing sorted containers are handles that allow O(1) access to a (key,value) pair in the structure.  The (key,value) pairs are all stored in an array, and the semitoken is the index into this array of the data item.  The full token is a tuple: (data_structure, semitoken).  The point of tokens and semitokens is that they can be incremented or decremented making it possible to iterate over the items in the data structure in forward or reverse order, or over a subrange.  The code to increment a semitoken chases the leaf node up to the lowest ancestor level in which is *not* the rightmost child, then going to the next sibling of that ancestor, and then following this back to another leaf.  Thus, incrementing a token requires O(log n) operations, where n is the number of entries in the data structure.

The distinction between tokens and semitokens is that tokens require heap allocation (there may be some new features in 1.5 to address this matter, but I’m not sure), whereas semitokens are just immutable integers and hence can be put into arrays or other data structures cheaply.

src/red_black_tree.jl
eulerkochy: I think, it shouldn't overload `Base.minimum`. Minimum value in a subtree of a node isn't that obvious, so a separate method, should be better.

PR #640, opened 2020-07-06
AVL Tree

eulerkochy:
Part of GSoC'20 evaluation
This PR implements AVL Tree.


src/avl_tree.jl
```diff
@@ -0,0 +1,259 @@
+# it has unique keys
+# leftChild has keys which are less than the node
+# rightChild has keys which are greater than the node
+# height stores the height of the subtree.
+mutable struct AVLTreeNode{K}
+    height::Int8
```
oxinabox, line 6: Is Int8 big enough?
that means max height is 128.
Is it so self-balanced that that means 2^128ish items (which is more than enough)?

src/avl_tree.jl
```diff
@@ -0,0 +1,259 @@
...
+# balance is the difference of height between leftChild and rightChild of a node.
+function get_balance(node::Union{AVLTreeNode, Nothing})
+    if node == nothing
+        return 0
+    else
+        return get_height(node.leftChild) - get_height(node.rightChild)
+    end
+end
+
+fix_height(node::AVLTreeNode) = 1 + max(get_height(node.leftChild), get_height(node.rightChild))
```
oxinabox, line 40: What does this do?
It doesn't fix the hight since it doesn't mutate.

Perhaps it should be called:
`compute_hight` and have a comment that it computes the hight rather than using the variable?

or make it should be called:
`recompute_height!(node)` and it should update the hight in place?

src/avl_tree.jl
```diff
@@ -0,0 +1,259 @@
...
+            return nothing
+        elseif key < node.data
+            node.leftChild = delete_node(node.leftChild, key)
+        elseif key > node.data
+            node.rightChild = delete_node(node.rightChild, key)
+        else
+            if node.leftChild == nothing
+                temp = node.rightChild
+                node = nothing
+                return temp
```
oxinabox, line 194: consider rather than calling this `temp`
a more explictit name like `result`, `return_value` or similar.

src/avl_tree.jl
```diff
@@ -0,0 +1,259 @@
...
+Inserts `key` in the `tree` if it is not present.
+"""
+function Base.push!(tree::AVLTree{K}, key0) where K
+    key = convert(K, key0)
+    insert!(tree, key)
+end
+
+function Base.delete!(tree::AVLTree{K}, d::K) where K
+
+    function delete_node(node::Union{AVLTreeNode, Nothing}, key)
```
oxinabox, line 183: Should this have a `!`?

src/avl_tree.jl
```diff
@@ -0,0 +1,259 @@
...
+    tree.count -= 1
+    return tree
+end
+
+"""
+    getindex(tree, ind)
+
+Gets the key present at index `ind` of the tree. Indexing is done in increasing order of key.
+"""
+getindex(tree, ind)
```
oxinabox, line 244: Do we need this docstring?
Or is the docstring that is attached to the fully generic `Base.getindex` clear enough?

This one in parocular is too generically typed I think and will cause problems like:
https://github.com/JuliaCollections/DataStructures.jl/issues/663

src/avl_tree.jl
```diff
@@ -0,0 +1,259 @@
...
+
+function Base.getindex(tree::AVLTree{K}, ind) where K 
+    @boundscheck (1 <= ind <= tree.count) || throw(ArgumentError("$ind should be in between 1 and $(tree.count)"))
+    function traverse_tree_inorder(node::Union{AVLTreeNode, Nothing})
+        if (node != nothing)
+            left = traverse_tree_inorder(node.leftChild)
+            right = traverse_tree_inorder(node.rightChild)
+            append!(push!(left, node.data), right)
+        else
+            return K[]
```
oxinabox, line 254: Can this ever happen?
I don't think it can since we did a check

src/avl_tree.jl
```diff
@@ -0,0 +1,259 @@
...
+
+"""
+    getindex(tree, ind)
+
+Gets the key present at index `ind` of the tree. Indexing is done in increasing order of key.
+"""
+getindex(tree, ind)
+
+function Base.getindex(tree::AVLTree{K}, ind) where K 
+    @boundscheck (1 <= ind <= tree.count) || throw(ArgumentError("$ind should be in between 1 and $(tree.count)"))
```
oxinabox, line 247: Should we use  a `BoundsError` here?

oxinabox (approved):
A few comments, 
looks basically right to me.

src/avl_tree.jl
eulerkochy: Well, it is the smallest `Integer` type right ?

src/avl_tree.jl
eulerkochy: Yes. It can. `node` might not be `nothing`, but it's `leftChild` and `rightChild` can be.

src/avl_tree.jl
eulerkochy: Removed !!

src/avl_tree.jl
oxinabox: this will cause a out of bounds error a few lines later right?
Should we just throw on directly here?

eulerkochy: Ya, we could. Even I was thinking about that. 
Once `traverse_tree_inorder` is called, we can be sure that, there is a valid index. So, it will come from the `leftChild`, or the `rightChild`. 
Let `idx` be the index we are interested in. 
`L` = size of the `left` array
`R` = size of the `right` array.
```julia
if idx <= L
    return left[idx]
else
    return right[idx-L]
```

eulerkochy:
Posting as a comment to describe the change in the latest commit.
Previously, `getindex` worked in by traversing through `log(n)` nodes but allocating `O(n)` memory. So, the effectiveness of maintaining a balanced BST was lost. Order-statistics worked in `O(n)`, instead of `O(log n)`.
So, here the change, adding a field `subsize` to the nodes, can help in answering order statistics in `O(log n)`.
```
subsize = number of nodes, including itself, that is present in the subtree of the node
```
And, here's the pseudo-code for the proposed algorithm.
```
function traverse_tree(node, idx)
    if node != nothing
        L = (node.leftChild != nothing) ? node.leftChild.subsize : 0
        R = (node.rightChild != nothing) ? node.rightChild.subsize : 0
        if idx <= L
            traverse_tree(node.leftChild, idx)
        else if idx == L+1
            return node.data
        else
            traverse_tree(node.rightChild, idx-L-1)
```

src/avl_tree.jl
oxinabox: right.

src/DataStructures.jl
```diff
@@ -60,7 +60,7 @@ module DataStructures
     export DiBitVector
 
     export RBTree, search_node, minimum_node
-    export AVLTree
+    export AVLTree, rank
```
oxinabox, line 63: The name `rank` is also used by `LinearAlgebra`.
Are these the same concept? Should we overload `LInearAlgebra.rank`?

eulerkochy: `rank` in Linear Algebra, means something fundamentally from what is present here.
In linear algebra,`rank` means the maximum number of linearly independent vectors in the vector space generated by the columns/rows.
Here, it correspondings to the order/position in which the keys appear in the sorted sequence.
For example, 
if the tree consists of ` keys = [2, 5, 7, 10]`, 
```
rank(2) = 1
rank(5) = 2
rank(7) = 3
rank(10) = 4
```

oxinabox: ok, fair enough.
I wonder if we can name it something different to avoid a name conflict with LinearAlgebra.
Like `sort_rank` or soething

PR #641, opened 2020-07-08
CircularBuffer: Add benchmarks and improve performance

vyu:
This PR adds a set of benchmarks for `CircularBuffer` and also improves its performance.

Here are some `PkgBenchmark` reports using the benchmarks and comparing the new performance:
- [Comparison with baseline](https://gist.github.com/vyu/51705d29b4aa65fd52162ac8fa6c07b9) on Julia 1.4.
- [Comparison with baseline](https://gist.github.com/vyu/4bbc48fd38cdbf764e8c5b660e985943) on Julia 1.0.
- [Timings on Julia 1.4](https://gist.github.com/vyu/2b64c9e3edaf00f45845a30213c21c46), on an N2D instance on GCP.

The benchmark file can be called directly by `PkgBenchmark` using the `script` keyword. For example, `PkgBenchmark.benchmarkpkg("DataStructures"; script="benchmark/bench_circular_buffer.jl")`.

I'll add commit comments here to explain the rationale behind some of the changes.

There was some inconsistency in the code over whether to use accessor functions (`capacity(cb)`) or dot notation (`cb.capacity`) to read field values. For consistency within the file, I've modified them to use accessor functions. They generate identical code since the functions get inlined.

I haven't worked on `append!` and `fill!` yet because improving those is more complicated. I might get to them later if this PR goes through.

Thanks!

### DIFF
--- a/benchmark/bench_circular_buffer.jl
+++ b/benchmark/bench_circular_buffer.jl
@@ -1,0 +1,145 @@
+module BenchCircularBuffer
+
+using DataStructures
+using BenchmarkTools
+using Random
+
+const SUITE = BenchmarkGroup()
+const CAPS = [1024, 1024^2]
+
+# Empirically, 1 second is sufficient for consistent timings here.
+BenchmarkTools.DEFAULT_PARAMETERS.seconds = 1
+
+function init_cb(a)
+    len = length(a)
+    cb = CircularBuffer{Int}(len)
+    cb.first = 1 + div(len, 2)
+    return append!(cb, a)
+end
+
+function perf_chained_getindex(cb)
+    i = 1
+    for _ in eachindex(cb)
+        i = cb[i]
+    end
+    return i
+end
+
+function perf_first(cb)
+    total = 0
+    for i in 1:capacity(cb)
+        cb.first = i
+        total += first(cb)
+    end
+    return total
+end
+
+function perf_last(cb)
+    total = 0
+    for i in 1:capacity(cb)
+        cb.first = i
+        total += last(cb)
+    end
+    return total
+end
+
+g = addgroup!(SUITE, "access")
+
+# Creating a new copy for each sample seems to eliminates an unknown but substantial source
+# of timing variation on some machines. This is why each benchmark has `setup=(cb = deepcopy($cb))`.
+for cap in CAPS
+    Random.seed!(0)
+    cb = init_cb(randcycle(cap))
+
+    g["chained_getindex", cap] = @benchmarkable perf_chained_getindex(cb) setup=(cb = deepcopy($cb))
+
+    g["sum", cap] = @benchmarkable sum(cb) setup=(cb = deepcopy($cb))
+    g["foldl", cap] = @benchmarkable foldl(+, cb) setup=(cb = deepcopy($cb))
+    g["foldr", cap] = @benchmarkable foldr(+, cb) setup=(cb = deepcopy($cb))
+
+    g["first", cap] = @benchmarkable perf_first(cb) setup=(cb = deepcopy($cb))
+    g["last", cap] = @benchmarkable perf_last(cb) setup=(cb = deepcopy($cb); cb.length = div($cap, 2))
+
+    g["convert", cap] = @benchmarkable convert(Array, cb) setup=(cb = deepcopy($cb))
+end
+
+function perf_setindex!(cb, indices)
+    for i in indices
+        cb[i] = i
+    end
+    return cb
+end
+
+function perf_push!(cb)
+    for i in 1:capacity(cb)
+        push!(cb, i)
+    end
+    return cb
+end
+
+function perf_pushfirst!(cb)
+    for i in 1:capacity(cb)
+        pushfirst!(cb, i)
+    end
+    return cb
+end
+
+g = addgroup!(SUITE, "mutate")
+
+for cap in CAPS
+    cb = init_cb(-1:-1:-cap)
+
+    Random.seed!(0)
+    g["random_setindex!", cap] = @benchmarkable perf_setindex!(cb, $(randperm(cap))) setup=(cb = deepcopy($cb))
+    g["forward_setindex!", cap] = @benchmarkable perf_setindex!(cb, 1:$cap) setup=(cb = deepcopy($cb))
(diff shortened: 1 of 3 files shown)


src/circular_buffer.jl
```diff
@@ -18,6 +18,8 @@ end
 
 CircularBuffer(capacity) = CircularBuffer{Any}(capacity)
 
+Base.IndexStyle(::Type{<:CircularBuffer}) = IndexLinear()
+
```
vyu, line 22: `IndexLinear` (as opposed to the default `IndexCartesian` for `AbstractArrays`) allows functions to use an optimized `Base._mapreduce`. This, for example, speeds up `sum`.

src/circular_buffer.jl
```diff
@@ -28,17 +30,11 @@ function Base.empty!(cb::CircularBuffer)
     return cb
 end
 
-Base.@propagate_inbounds function _buffer_index_checked(cb::CircularBuffer, i::Int)
-    @boundscheck if i < 1 || i > cb.length
-        throw(BoundsError(cb, i))
-    end
-    _buffer_index(cb, i)
-end
```
vyu, line 36: This can be collapsed into `_buffer_index`, which can be simplified without loss of performance. Callers can use `@inbounds` to enable or disable bounds checking.

src/circular_buffer.jl
```diff
@@ -49,17 +45,19 @@ Get the i-th element of CircularBuffer.
 * `cb[1]` to get the element at the front
 * `cb[end]` to get the element at the back
 """
-@inline Base.@propagate_inbounds function Base.getindex(cb::CircularBuffer, i::Int)
-    cb.buffer[_buffer_index_checked(cb, i)]
+Base.@propagate_inbounds function Base.getindex(cb::CircularBuffer, i::Int)
+    j = _buffer_index(cb, i)
+    @inbounds return cb.buffer[j]
 end
```
vyu, line 51: If `i` is inbounds for `cb`, then `_buffer_index(cb, i)` is guaranteed inbounds for `cb.buffer`. We can separate these into two lines and annotate the second with `@inbounds`.

Also, `@inline` here is redundant because `@propagate_inbounds` implies `@inline`.

src/circular_buffer.jl
```diff
@@ -154,25 +154,21 @@ end
 
 Return the number of elements currently in the buffer.
 """
-Base.length(cb::CircularBuffer) = cb.length
-
-Base.eltype(::Type{CircularBuffer{T}}) where T = T
```
vyu, line 159: This is already defined for the `AbstractArray{T}` supertype.

src/circular_buffer.jl
```diff
@@ -154,25 +154,21 @@ end
 
 Return the number of elements currently in the buffer.
 """
-Base.length(cb::CircularBuffer) = cb.length
```
vyu, line 157: `Base.length(::AbstractArray)` is already a generic function that calls `size`. For consistency with Julia `Base`, we can define just `size` instead.

src/circular_buffer.jl
```diff
@@ -154,25 +154,21 @@ end
...
+Base.length
 
 """
     size(cb::CircularBuffer)
 
 Return a tuple with the size of the buffer.
 """
-Base.size(cb::CircularBuffer) = (length(cb),)
-
-Base.convert(::Type{Array}, cb::CircularBuffer{T}) where {T} = T[x for x in cb]
```
vyu, line 168: This `convert` was slower than the generic `convert`.

src/circular_buffer.jl
```diff
@@ -186,23 +182,21 @@ capacity(cb::CircularBuffer) = cb.capacity
...
+
 """
     last(cb::CircularBuffer)
 
 Get the last element of CircularBuffer.
 """
-Base.@propagate_inbounds function Base.last(cb::CircularBuffer)
-    @boundscheck (cb.length == 0) && throw(BoundsError(cb, 1))
-    return cb.buffer[_buffer_index(cb, cb.length)]
-end
```
vyu, line 208: This was no faster than the generic `last` function. The situation is different for `first` because the generic `first` does not know about `cb.first`, so the specialized `first` here is a shortcut. There is no such shortcut for `last` because we don't track the last index directly.

benchmark/bench_circular_buffer.jl
```diff
@@ -0,0 +1,145 @@
...
+end
+
+perf_empty_push!(cb) = (empty!(cb); perf_push!(cb))
+perf_empty_pushfirst!(cb) = (empty!(cb); perf_pushfirst!(cb))
+
+function perf_pop!(cb)
+    cap = capacity(cb)
+    cb.length = cap
+    total = 0
+    for _ in 1:cap
```
eulerkochy, line 108: In this codebase, we use the convention `for i = 1:num` for iterating over ranges. Also, I know it's trivial, but consider replacing `_`. It's there in few other places too.

benchmark/bench_circular_buffer.jl
```diff
@@ -0,0 +1,145 @@
...
+        total += pop!(cb)
+    end
+    return total
+end
+
+function perf_popfirst!(cb)
+    cap = capacity(cb)
+    cb.length = cap
+    total = 0
+    for _ in 1:cap
```
eulerkochy, line 118: Here, too!

vyu:
Thanks! I've modified lines with range iteration to follow your suggestion.

vyu:
I just noticed that `setindex(cb::CircularBuffer, item, i)` returns `cb`. This has been the case since the beginning (when `CircularBuffer` was added in v0.4.3) and this PR retains that behavior. But per `setindex` in Julia `Base`, it should really return `item`, not `cb`. I think this is a bug that ought to be fixed, but fixing it is potentially breaking if there is code out there that depends on this unconventional behavior. I can open a separate PR for it.

src/circular_buffer.jl
vyu: Reverted and replaced with a shorter definition. I was mistaken. Even though benchmarks showed no benefit, it might still be helpful on some architectures or code because the specialized method allows the bounds check to be elided by `@inbounds`, whereas the generic `last` does not.

oxinabox (approved):
Sorry for losing track of this. This PR looks good. 
Can it be rebased, then we can merge it.

PR #642, opened 2020-07-09
Don't explictly inline things in SparseIntSet

oxinabox:
This was causing time to first plot slow down.

This undoes https://github.com/JuliaCollections/DataStructures.jl/pull/540
which added them
I am not completely sure we want to do that.

But here is a branch that does do that so we can try it out


oxinabox:
Jameson says the impact of this didn't come out to be too much, so closing this for now

Issue #479, opened 2019-01-09
Maturing out the package (Path to DataStructures.jl v1.0)


oxinabox:
What do we need to do before we would be happy to tag DataStructures.jl v1.0?
I mean, we are one of the most common dependencies in the package ecosystem,
and julia 1.0 is out.

My thoughts;
 
 - [ ] Search the code base for anything written in outmoded styles (e.g. #472)
 - [ ] Get rid of anything deprecated #477 (but that will need to be done again after)
 - [ ] Split up the package #310, but noting that all the packages sliced out that way will want to hit 1.0 before (or at the same time as) DataStructures.jl does.

It hurts people when we break things, and while we are good at deprecating,
getting on the the "strict" side of semver is something we need to do for the sake of the ecossytem

oxinabox:
People keep asking when will DataStructures 1.0 come out.
The answer is as soon as we completely audit the package to make sure there is nothing that needs to be removed/corrected to follow modern julia conventions.
And have actioned those corrections.
A lot of the code in the package is pretty old, and follows e.g. conventions from julia 0.3.
We need to identify and correct these.

I will post a list of files at the bottom of this post.
Then people can go through this and use the checklist to make sure everything is good, one file at a time.




How to audit a file to check if 1.0 ready
---

 - Answer the question of **should we even keep this in the package**. E.g. no data structure that somone just kind of made up (like the now removed `Classifier Collection)
 - **Type Names Correct/Sensible:** for example make sure we don't have a type called `MagicVectors` that is a subtype of vector. That should be called `MagicVector` (singular).
 - **No constructors taking types as positional arguments.** For example there should not be `MagicVector(::Type{T}) where T`, instead there should just be `MagicVector{T}()`.
 - **No functions that duplicate constructors**. For example there should not be a function `magic_vector(xs)`, only `MagicVector(xs)`.
 - **No functions that are improperly cased and pretend to be constructors**. 
    - E.g. no `MagicVector(xs) = Vector(xs) .* "Magic"`, that should just be a normal function `magic_vector(xs) =  Vector(xs) .* "Magic"`
    - E.g. no `MagicVector(xs)=Vector{Magic}(xs)`, that should be a `const` type alias `const MagicVector=Vector{Magic}`
 - **No exported functions that could be just done with overloads of `Base`**. E.g. don't need `top`, `peak`, `head` all of which act the same as `first` but on a particular data structure.
[...]

PR #643, opened 2020-07-11
Mark as 0.18-DEV and remove deprecations

### LINKED ISSUE #496: Deprecate then Remove ClassifierCollections

### LINKED ISSUE #577: Prepare for release 0.18
#578 requires release of a breaking version.

List of other things we want to do before releasing 0.18:
 - [x] Delete deprecated `ClassifierCollection` 
 - [x] Delete all things in `deprecations`
 - [ ] deprecate `find_root` (see #603 )

oxinabox:
closes #577 

closes #496 

Follows up on https://github.com/JuliaCollections/DataStructures.jl/pull/603#issuecomment-612401679
and deprecates `find_root`

Required to be merged before #547


davidanthoff:
PLEASE, make it 1.0.0! So that going forward you have the option to do minor, patch and major releases.

oxinabox:
@davidanthoff  see https://github.com/JuliaCollections/DataStructures.jl/issues/479

Can't do it this release as it still has deprecations added.
Further there are pending breaking issues still open,
and we haven't completed the audit of the code base to be sure we don't have any Julia 0.3 style constructors, or other issues from not following the Julia 1.0 APIs. (like `push!` returning the value rather than the collection etc).
I am keen to get to 1.0 also, but not if I know that i will be tagging 2.0 soon after.

davidanthoff:
I think there is zero harm in a 2.0 shortly after a 1.0, who cares? But going 1.0 now allows you to do minor and patch releases in the meantime, which is a great benefit to downstream packages because it becomes clear what is what. I think this fixation on 1.0 as a somewhat special version really is not a good match with the fact that semver plays an important role in the package resolution story in Julia.

oxinabox:
While I do agree, in general and personally, I just am not willing to commit to that position on DataStructures.jl.
Because it is such a public package, and that is not the standard the community follows. Not yet at least.
Make the case on any other package I maintained and I would be seriously considering it, but not DataStructures.

(When we started talking about general registry, I argued that it should not accept packages before they tagged 1.0)

PR #645, opened 2020-07-22
SplayTree

eulerkochy:
Part of GSoC'20 implementation.

SplayTree has no additional memory footprint as compared to Red-Black Tree or AVL Tree. It offers `O(log n)` amortized time complexity for the operations `search`, `insert` and `delete` 

TO-DO

- [x] writes tests

### DIFF
--- a/src/splay_tree.jl
+++ b/src/splay_tree.jl
@@ -216,7 +216,7 @@ function Base.push!(tree::SplayTree{K}, key0) where K
 end
 
 function Base.getindex(tree::SplayTree{K}, ind) where K 
-    @boundscheck (1 <= ind <= tree.count) || throw(BoundsError("$ind should be in between 1 and $(tree.count)"))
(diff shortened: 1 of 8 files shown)


src/splay_tree.jl
```diff
@@ -0,0 +1,231 @@
...
+    if !isa(s.leftChild, Nothing)
+        s.leftChild.parent = nothing
+    end
+
+    tree.root = _join(tree, s.leftChild, t)
+    tree.count -= 1
+    return tree
+end
+
+function Base.insert!(tree::SplayTree{K}, d::K) where K
```
oxinabox, line 181: We probably should not have 2 argument `insert!`,  just `push!`.
Base only provides a 3 argument `insert` for things with seperate indexes and values.
Base provides it for `Vector` but not for `Set`

oxinabox: Possibly we need to deprecate it out of other trees also.
I should have spotted this before

oxinabox (approved):
A few comments, once those are address should be good to merge

PR #646, opened 2020-08-03
Use Base.Ordering with heaps

jlumpe:
Note: this is basically a duplicate of #547, which still seems active. I wrote all this before finding that one, so I'm submitting it anyways just in case any of it is still useful.

Changes made:

* `comp::Comp` arguments and fields replaced with `order::Base.Ordering` everywhere
  * Heaps now support arbitrary orderings
  * `LessThan` and `GreaterThan` singleton types and `compare` function replaced with `Base.Forward`, `Base.Reverse`, and `Base.Order.lt`.
* Heap constructors now take an `Ordering` as a constructor argument rather than a type parameter, because in general we can't get an instance from just the type (exception follows)
* Min/max type aliases are now `const BinaryMinHeap{T} = BinaryHeap{T, typeof(Forward)}` and `const BinaryMaxHeap{T} = BinaryHeap{T, typeof(Reverse)}`, and similarly for mutable version
  * These required adding additional constructors to `BinaryHeap`/`MutableBinaryHeap` for these two special cases where the order is a singleton: `BinaryHeap{T, typeof(Forward)}(...) = BinaryHeap{T}(Forward, ...)` etc
* `BinaryHeap`, `MutableBinaryHeap`, `nsmallest`, and `nlargest` now have methods with the `lt`, `by`, and `rev` keyword arguments like in `sort`, which is more user-friendly than requiring an `Ordering` instance.

This introduces similar breaking changes as in #547 in that `isless` is now used for comparison by default instead of `<`, which may yield different results where `NaN`s are involved (although, I would say that the previous behavior was undefined).
I haven't updated or run the benchmark script yet so I don't know if there are any performance regressions.

### DIFF
--- a/docs/src/heaps.md
+++ b/docs/src/heaps.md
@@ -95,7 +95,10 @@ array without sorting the entire array first:
 ```julia
 nlargest(3, [0,21,-12,68,-25,14]) # => [68,21,14]
 nsmallest(3, [0,21,-12,68,-25,14]) # => [-25,-12,0]
+nlargest(3, [0,21,-12,68,-25,14], by=abs) # => [68,-25,21]
 ```
 
-`nlargest(n, a)` is equivalent to `sort(a, lt = >)[1:min(n, end)]`, and
-`nsmallest(n, a)` is equivalent to `sort(a, lt = <)[1:min(n, end)]`.
+Both functions support the same `lt`, `by`, and `order` keyword arguments as
+`sort` to define how elements are ordered.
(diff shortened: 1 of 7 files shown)


oxinabox:
Redundant with #547

jlumpe:
I think the one thing not included is the ability to use custom orderings which are not singletons. I can create a new PR for that.

jlumpe:
@oxinabox new PR is here: #673

PR #652, opened 2020-08-13
[WIP] Tried to add temporary immutable struct

AliMalik9599:
Hi, I'm relatively new to Julia and was interested in this problem. I attempted to add a temporary way to make queue immutable, but definitely a work in progress and needs changes. Any tips are appreciated, thanks.


oxinabox:
making Queue a immutable struct should be just a matter of removing the `mutable`, and making sure all operations just mutate the backing `Deque` instead.
That would be a good idea.

Issue #655, opened 2020-08-17
Shouldn't iterate(::MultiDict) return each pair, like it was inserted?


NHDaly:
This feels wrong and surprising to me:
```julia
julia> d = DataStructures.MultiDict{Int,Int}()
MultiDict{Int64,Int64}(Dict{Int64,Array{Int64,1}}())

julia> insert!(d, 1,1)
MultiDict{Int64,Int64}(Dict(1 => [1]))

julia> insert!(d, 1,2)
MultiDict{Int64,Int64}(Dict(1 => [1, 2]))

julia> collect(d)
1-element Array{Pair{Int64,Array{Int64,1}},1}:
 1 => [1, 2]
```

I would've expected iterating `d` to return the two pairs that I inserted! That is, I expected `[1=>1, 1=>2]`.

Can we change it to this? Does this match others' expectations?

NHDaly:
Very related, this seems wrong to me?:
```julia
julia> (2=>3) in Dict(2=>3)
true

julia> (2=>3) in DataStructures.MultiDict(2=>3)
false
```
It's because it is looking for this:
```julia
julia> (2=>[3]) in DataStructures.MultiDict(2=>3)
true
```

NHDaly:
Also, related, this constructor should probably work too?:
```julia
julia> MultiDict(1=>2, 2=>3)
MultiDict{Int64,Int64}(Dict(2 => [3],1 => [2]))

julia> MultiDict(x=>x+1 for x in [1,2])
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type Array{Any,1}
Closest candidates are:
  convert(::Type{T}, ::AbstractArray) where T<:Array at array.jl:554
  convert(::Type{T}, ::T) where T<:AbstractArray at abstractarray.jl:14
  convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/factorization.jl:55
  ...
Stacktrace:
 [1] setindex!(::Dict{Any,Array{Any,1}}, ::Int64, ::Int64) at ./dict.jl:380
 [2] Dict at ./dict.jl:103 [inlined]
 [3] MultiDict at /Users/nathandaly/.julia/packages/DataStructures/iWRBf/src/multi_dict.jl:12 [inlined]
[...]

Issue #654, opened 2020-08-17
Add a delete!(::MultiDict, k, v)?


NHDaly:
Does it make sense to add a `delete!(::MultiDict, k, v)` function where you can delete a _specific_ `key=>value` pair?

I was expecting this, and was surprised it's not offered.

NHDaly:
Like, maybe something like this?:
```julia
function delete_pair!(d::MultiDict, key, val)
    vs = get(d, key, Base.secret_table_token)
    if vs !== Base.secret_table_token
        filter!(x->x≠val, vs)
    end
    isempty(vs) && delete!(d, key)
    return d
end
```

NHDaly:
Unfortunately, this does a linear scan of the values vector to find the target, which isn't great. But to fix this I guess we'd have to use a different datastructure there, like a Set{V}..

NHDaly:
Indeed, the golang vector-based MultiDict does exactly the above, including the terrible worst-case O(n^2) performance if you delete all pairs from the dict with the same keys, one at a time.
https://github.com/jwangsadinata/go-multimap/blob/c29f3d7f33b68d8bcd82880a558e9f8eed58e2b3/slicemultimap/slicemultimap.go#L76-L88

oxinabox:
I think this makes sense

PR #656, opened 2020-08-19
RFC: Change MultiDict constructors to more closely match the abstraction: accept _multiple pairs_

NHDaly:
From my understanding, the MultiDict is like a Dict, except that it can hold multiple pairs that have the same key. Much like bag-semantics in SQL.

And the current constructors do imply this as well:
```julia
julia> MultiDict(1=>1, 1=>2)
MultiDict{Int64,Int64}(Dict(1 => [1, 2]))
```

But before this PR, the constructors actually _sometimes_ accept pairs of {K,V}, in some contexts, but they also _sometimes_ accept pairs of {K,Vector{V}}, in other contexts. But neither would work in all contexts.

Here are some examples of the failures before this PR:

```julia
julia> MultiDict(1=>1, 1=>2)
MultiDict{Int64,Int64}(Dict(1 => [1, 2]))

julia> MultiDict(1=>[2])   # without type args, [2] is treated as an array value when it's a pair
MultiDict{Int64,Array{Int64,1}}(Dict(1 => [[2]]))

julia> MultiDict([1=>[2]])   # Same for arrays of pairs.
MultiDict{Int64,Array{Int64,1}}(Dict(1 => [[2]]))

julia> MultiDict([(1,[2])])   # but when it's an array of *tuples*, an array is treated as _all the possible values for the given key!_
MultiDict{Int64,Int64}(Dict(1 => [2]))

julia> MultiDict([(1,2), (1,3)])   # But it's not required.
MultiDict{Int64,Int64}(Dict(1 => [2, 3]))

julia> MultiDict{Int,Int}(1=>[2]) # But WITH type args, the second element IS required to be a vector of all values
MultiDict{Int64,Int64}(Dict(1 => [2]))

julia> MultiDict{Int,Int}(1=>2, 2=>3)  # Can't construct with pairs if you've included the types on the dict.
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type Array{Int64,1}
Closest candidates are:
  convert(::Type{T}, ::AbstractArray) where T<:Array at array.jl:533
  convert(::Type{T}, ::T) where T<:AbstractArray at abstractarray.jl:14
  convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/factorization.jl:55
  ...
Stacktrace:
 [1] setindex!(::Dict{Int64,Array{Int64,1}}, ::Int64, ::Int64) at ./dict.jl:380
 [2] Dict{Int64,Array{Int64,1}}(::Pair{Int64,Int64}) at ./dict.jl:107
 [3] MultiDict{Int64,Int64}(::Pair{Int64,Int64}) at /Users/nathandaly/.julia/packages/DataStructures/iWRBf/src/multi_dict.jl:12
 [4] top-level scope at REPL[17]:1

julia> xs = [1,1,2];

julia> MultiDict(x=>x+1 for x in xs)  # Also, in Generators, pairs are also required to point to vectors, **with or without types**
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type Array{Any,1}
Closest candidates are:
  convert(::Type{T}, ::AbstractArray) where T<:Array at array.jl:533
  convert(::Type{T}, ::T) where T<:AbstractArray at abstractarray.jl:14
  convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/factorization.jl:55
  ...
Stacktrace:
 [1] setindex!(::Dict{Any,Array{Any,1}}, ::Int64, ::Int64) at ./dict.jl:380
 [2] Dict at ./dict.jl:103 [inlined]
 [3] MultiDict at /Users/nathandaly/.julia/packages/DataStructures/iWRBf/src/multi_dict.jl:12 [inlined]
 [4] multi_dict_with_eltype at /Users/nathandaly/.julia/packages/DataStructures/iWRBf/src/multi_dict.jl:28 [inlined]
 [5] MultiDict(::Base.Generator{Array{Int64,1},var"#19#20"}) at /Users/nathandaly/.julia/packages/DataStructures/iWRBf/src/multi_dict.jl:18
 [6] top-level scope at REPL[13]:1

julia> MultiDict{Int,Int}(x=>x+1 for x in xs)   # Also, in Generators, pairs are also required to point to vectors, **with or without types**
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type Array{Int64,1}
Closest candidates are:
  convert(::Type{T}, ::AbstractArray) where T<:Array at array.jl:533
  convert(::Type{T}, ::T) where T<:AbstractArray at abstractarray.jl:14
  convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/factorization.jl:55
  ...
Stacktrace:
 [1] setindex!(::Dict{Int64,Array{Int64,1}}, ::Int64, ::Int64) at ./dict.jl:380
 [2] Dict at ./dict.jl:103 [inlined]
 [3] MultiDict{Int64,Int64}(::Base.Generator{Array{Int64,1},var"#27#28"}) at /Users/nathandaly/.julia/packages/DataStructures/iWRBf/src/multi_dict.jl:12
 [4] top-level scope at REPL[31]:1

julia> MultiDict(x=>[x+1] for x in xs)  # These work, but they're also super weird because they're different than providing a single pair.
MultiDict{Any,Any}(Dict{Any,Array{Any,1}}(2 => [3],1 => [2]))

julia> MultiDict{Int,Int}(x=>[x+1] for x in xs)  # same here
MultiDict{Int64,Int64}(Dict(2 => [3],1 => [2]))
```

---------------------------------

## New Proposal

I tried fixing just some of those weird constructors, but the inconsistencies were too hard to grapple with. I think the new rules I'm proposing are quite logical, concise, easy to understand, and keep close to the core concept. But they will be a breaking change in the API. Is that okay?

Here's what I'm proposing and what this PR implements:
1. MultiDict constructors expect pairs or tuples or generators of {K,V}:
    - `MultiDict(1=>1,1=>2)`, `MultiDict{Int,Int}(1=>2,2=>3)`, `MultiDict{Int,Int}([(1,1), (1,2)])`, etc
2. To manually construct a MultiDict with {K,Vector{V}), you can construct with the _underlying Dict_ directly.
    - `MultiDict(Dict(1=>[1,2])`, `MultiDict{Int,Int}(Dict(1=>[2],2=>[3]))`, etc

The tests used the constructors that passed all values in a vector in a few places, so I've changed those to either enumerate the pairs, or to use the inner Dict constructor.

-------------------

Does this seem reasonable? I think it makes the type much easier to use, more consistent, and more flexible. As a user, it took me a while before finding a constructor that actually did what I wanted, and so i think this will be helpful to others! :)

If we do go with this, should I bump the minor release number? (0.18 => 0.19)

Let me know what you think! Thanks! :)


NHDaly:
As an example, I left this comment here because i thought that the generator constructor was just broken:
https://github.com/JuliaCollections/DataStructures.jl/issues/655#issuecomment-675559766
 I had no idea that they value was supposed to be a vector containing all values, considering that that's violating the core metaphor of this container.

src/multi_dict.jl
```diff
@@ -127,3 +141,32 @@ function iterate(e::EnumerateAll, s)
     v, vst = vstate
     return ((k, v), (dstate, k, vs, vstate))
 end
+
+# grow_to! copied from Base -- needed for abstract generator constructor
+function Base.grow_to!(dest::MultiDict{K, V}, itr) where V where K
+    y = iterate(itr)
+    y === nothing && return dest
+    ((k,v), st) = y
```
eulerkochy, line 149:
```suggestion
    ((k, v), st) = y
```

src/multi_dict.jl
```diff
@@ -127,3 +141,32 @@ function iterate(e::EnumerateAll, s)
...
+    insert!(dest2, k, v)
+    Base.grow_to!(dest2, itr, st)
+end
+
+# this is a special case due to (1) allowing both Pairs and Tuples as elements,
+# and (2) Pair being invariant. a bit annoying.
+function Base.grow_to!(dest::MultiDict{K,V}, itr, st) where V where K
+    y = iterate(itr, st)
+    while y !== nothing
+        (k,v), st = y
```
eulerkochy, line 160:
```suggestion
        (k, v), st = y
```

src/multi_dict.jl
```diff
@@ -127,3 +141,32 @@ function iterate(e::EnumerateAll, s)
...
+        if isa(k,K) && isa(v,V)
```
eulerkochy, line 161:
```suggestion
        if isa(k, K) && isa(v, V)
```

src/multi_dict.jl
```diff
@@ -127,3 +141,32 @@ function iterate(e::EnumerateAll, s)
...
+            insert!(dest, k, v)
+        else
+            new = empty(dest, promote_typejoin(K,typeof(k)), promote_typejoin(V,typeof(v)))
```
eulerkochy, line 164: It's trivial but use a better variable name than `new`. IMHO, it's better to avoid keywords as variable names. Probably, `new_md` would be much better.

eulerkochy (changes requested):
I've reviewed the PR, and requested some stylistic changes. What you're proposing is a breaking change, so I'll like to hear @oxinabox 's opinion about this.

src/multi_dict.jl
NHDaly: Eww gross, completely agreed. I copied this whole function from Base; sorry i didn't read it very closely.. Lemme make all your edits! 👍

NHDaly: Done, thanks! :)

NHDaly:
Also, some of the other MultiDict changes I've proposed: #654, #655 would also be breaking changes.. Maybe it makes sense to do them all at once, in one release?

oxinabox:
yes, we do want to batch up the breaking changes into a single release.
It is unfortunate that these PRs were not make 2 weeks ago, as we were making a breaking release at the time.
it would be annoying to make another breaking release in the next 2 months.

NHDaly:
😭 oh no! Bummer. I'm fine leaving all of these on like one big "multi-dict" branch or something that we can wait to merge until your comfortable with another breaking change? Does that work?

I guess, the kind of nice thing about this package is that probably most people are using it for only a few container types, so hopefully each breaking API change only affects some (hopefully different) people?

oxinabox:
It we can't workout a way to deprecate them safely then yes, putting it onto a branch would be best.
Then we can look at that branch and see if at the end its all worth making a breaking release over.
or if we want to hold-off

NHDaly:
Okay, makes sense! If you can create a new branch (maybe `MultiDict`? Or `MultiDict-changes` or something?), I'll switch the PR base over to it!

NHDaly:
Or I guess you're saying maybe we can still support the older constructors but deprecated?

Unfortunately I don't think so, without the behavior remaining ambiguous and confusing. For example, the fact that whether or not you add the type parameters changes the meaning of the pair arguments seems completely broken to me:
```julia
julia> MultiDict(1=>[2,3])  # Treated as a single pair
MultiDict{Int64,Array{Int64,1}}(Dict(1 => [[2, 3]]))

julia> MultiDict{Int,Int}(1=>[2,3])  # Treated as a normalized representation of all pairs
MultiDict{Int64,Int64}(Dict(1 => [2, 3]))
```

I think that in order to make the constructors behave consistently, we'll need to break the old behavior.

I'm very open to suggestions for how to do this change in a safe, backwards compatible, deprecation way though if you see such a potential path!! :) Thanks again for your time :)

NHDaly:
Thanks friends! Sorry it took me so long to get to this.

Now that i have the committer bit on this repo, i've created a new [MultiDict-reboot](https://github.com/JuliaCollections/DataStructures.jl/tree/MultiDict-reboot) branch, and adjusted this PR to point to it. i'm going to do that with all the MultiDict PRs that have breaking changes, so we can do them all together.

I still think it's nice to review and merge this as individual PRs though, to make reviewing easier. :) Anyone wanna do a final review and/or approval? ❤️

PR #659, opened 2020-08-19
deprecate nil and cons

oxinabox:
Slowly getting rid of julia 0.3isms.

Possibly should also deprecate `head` for `first`?


nickrobinson251 (approved):

oxinabox:
We are just going to deprecate this type outright I think

PR #668, opened 2020-08-29
Replace `import` with `using`

nickrobinson251:
- This changes the internal code style to make module qualification required for adding methods to a function.
- This is related to #479 
  - trying to read the code - as is needed for #479 - was made much more difficlut by not being able to easily see what functin declaration were extending function from other packages (mainly Base), especially since some methods _were_ using with module qualification e.g. `Base.show` but others were not e.g. just `show`.
  - This changes method extension to always be explicit e.g. `Base.show` not `show`
- This should mostly be safe, since tests should catch any method definition for `foo` that should now be `Base.foo` (and throw e.g. `ERROR: LoadError: LoadError: LoadError: MethodError: no method matching foo(::Array{Any,1})
You may have intended to import Base.foo`
- But the slight risk (of this PR introducing bugs) is if a previously `import`ed Base function (that's also not exported by Base) was not extended but instead used, and used in a code path that's not covered by tests. 
  - I've tried to look out for that
  - and tests coverage in 95%


src/disjoint_set.jl
```diff
@@ -84,7 +84,7 @@ in_same_set(s::IntDisjointSets{T}, x::T, y::T) where {T<:Integer} = find_root!(s
 Merge the subset containing x and that containing y into one
 and return the root of the new set.
 """
-function union!(s::IntDisjointSets{T}, x::T, y::T) where {T<:Integer}
+function Base.union!(s::IntDisjointSets, x::Integer, y::Integer)
```
nickrobinson251, line 87:
```suggestion
function Base.union!(s::IntDisjointSets{T}, x::T, y::T) where {T<:Integer}
```

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.18.2"
+version = "0.18.3"
```
oxinabox, line 3: since this is a completely invisable change to the user, lets leave the version number as is

```suggestion
version = "0.18.2"
```

oxinabox (approved):
This is great!

PR #670, opened 2020-08-30
Deprecate `reverse_iter` to `Iterators.reverse`

nickrobinson251:
- trying out replacing `reverse_iter` with `Iterators.reverse` (not `reverse` since we don't promise to make a copy) -- thoughts?
- in the spirit of "No exported functions that could be just done with overloads of Base" (#479) 
- as commented in #479, I don't _think_ the `Iterators.reverse(::T) -> Iterators.Reverse{T}` is promised by the Iterators API, since it's not documented or tested for in Base (i.e. the changes of this PR don't break that API).

### DIFF
--- a/src/deque.jl
+++ b/src/deque.jl
@@ -115,6 +115,8 @@ struct DequeIterator{T}
     q::Deque
 end
 
+Base.last(qi::DequeIterator) = last(qi.q)
+
 function Base.iterate(qi::DequeIterator{T}, (cb, i) = (qi.q.head, qi.q.head.front)) where T
     i > cb.back && return nothing
     x = cb.data[i]
@@ -130,11 +132,7 @@ end
 
 # Backwards deque iteration
 
-struct ReverseDequeIterator{T}
-    q::Deque
-end
-
-function Base.iterate(qi::ReverseDequeIterator{T}, (cb, i) = (qi.q.rear, qi.q.rear.back)) where T
+function Base.iterate(qi::Iterators.Reverse{<:Deque}, (cb, i) = (qi.itr.rear, qi.itr.rear.back))
     i < cb.front && return nothing
     x = cb.data[i]
 
@@ -148,12 +146,9 @@ function Base.iterate(qi::ReverseDequeIterator{T}, (cb, i) = (qi.q.rear, qi.q.re
     return (x, (cb, i))
 end
 
-reverse_iter(q::Deque{T}) where {T} = ReverseDequeIterator{T}(q)
-
 Base.iterate(q::Deque{T}, s...) where {T} = iterate(DequeIterator{T}(q), s...)
 
 Base.length(qi::DequeIterator{T}) where {T} = qi.q.len
-Base.length(qi::ReverseDequeIterator{T}) where {T} = qi.q.len
 
 Base.collect(q::Deque{T}) where {T} = T[x for x in q]
 
--- a/docs/src/stack_and_queue.md
+++ b/docs/src/stack_and_queue.md
@@ -1,4 +1,3 @@
-
 # Stack and Queue
 
 The `Stack` and `Queue` types are a light-weight wrapper of a deque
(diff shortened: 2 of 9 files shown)


oxinabox:
Seems like this is inline with the documentation for `Base.Iterators.reverse`

```
help?> Base.Iterators.reverse
  Iterators.reverse(itr)

  Given an iterator itr, then reverse(itr) is an iterator over the same collection but in the
  reverse order.

  This iterator is "lazy" in that it does not make a copy of the collection in order to reverse
  it; see Base.reverse for an eager implementation.

  Not all iterator types T support reverse-order iteration. If T doesn't, then iterating over
  Iterators.reverse(itr::T) will throw a MethodError because of the missing iterate methods for
  Iterators.Reverse{T}. (To implement these methods, the original iterator itr::T can be obtained
  from r = Iterators.reverse(itr) by r.itr.)
```

src/deque.jl
```diff
@@ -148,12 +144,10 @@ function Base.iterate(qi::ReverseDequeIterator{T}, (cb, i) = (qi.q.rear, qi.q.re
     return (x, (cb, i))
 end
 
-reverse_iter(q::Deque{T}) where {T} = ReverseDequeIterator{T}(q)
-
 Base.iterate(q::Deque{T}, s...) where {T} = iterate(DequeIterator{T}(q), s...)
 
 Base.length(qi::DequeIterator{T}) where {T} = qi.q.len
-Base.length(qi::ReverseDequeIterator{T}) where {T} = qi.q.len
+Base.length(qi::Iterators.Reverse{<:Deque}) = qi.itr.len
```
oxinabox, line 150: I think we can delete this line, as it is defined in `Base`
https://github.com/JuliaLang/julia/blob/5f0469b2273c8de57a26aec88468e5b5eed3a61b/base/iterators.jl#L101

nickrobinson251:
```suggestion
```

oxinabox (approved):
I think this is a good idea.
Lets give it a few days for others to chime in, and if noone says it is a bad idea lets merge it,
after bumping the patch version

nickrobinson251:
> the question is particular for `Stack` which for its `reverse` would iterate `reverse(reverse(::Deque)`
> so it makes sense to have it just iterate the `Deque`.
> https://github.com/JuliaCollections/DataStructures.jl/pull/670/files#diff-7356e908345ca352e9d4e26734784cfeR56

Yeah. I wonder: even if it's okay for `Iterators.reverse(::Stack)` to return `DequeIterator` (rather than a `Reverse{<:Stack}`), perhaps the return should still implement [all the methods the work on `Iterators.Reverse`, such as `first` and `last`](https://github.com/JuliaLang/julia/blob/5f0469b2273c8de57a26aec88468e5b5eed3a61b/base/iterators.jl#L100-L106)? We'd need to add these for `DequeIterator` i think.

stevengj:
> perhaps he can comment on if it is allowable to return something other than a Reverse{T}

Absolutely.  That's why it's called `reverse(itr)` and not `Reverse(itr)`.  For example, we already do:
```jl
julia> Iterators.reverse(1:10)
10:-1:1
```

nickrobinson251:
^ just rebased to resolve merge conflicts

oxinabox:
Merge and tag when ready.

nickrobinson251:
i don't have permissions to do that. But feel free to give me them if you want. Or please merge and tag when you've the time

oxinabox:
I gave you permissions several days ago.
I see you are still listed as pending accepting.
I think you can go to https://github.com/orgs/JuliaCollections/people
to accept.
or if not you'll need to find the email

nickrobinson251:
ah, nice, thanks! i missed that email (i get too many GitHub email and need to fix that)
But i could see the invite by jsut going to https://github.com/JuliaCollections :)

PR #672, opened 2020-09-01
Add `top` alias

YingboMa:
Removing the `top` function broke a lot of code in the SciML ecosystem.

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -112,5 +112,4 @@ module DataStructures
     include("splay_tree.jl")
 
     include("deprecations.jl")
-@deprecate top(x) = first(h)
 end
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -25,7 +25,7 @@ module DataStructures
     export AbstractHeap, compare, extract_all!, extract_all_rev!
     export BinaryHeap, BinaryMinHeap, BinaryMaxHeap, nlargest, nsmallest
     export MutableBinaryHeap, MutableBinaryMinHeap, MutableBinaryMaxHeap
-    export heapify!, heapify, heappop!, heappush!, isheap, top
+    export heapify!, heapify, heappop!, heappush!, isheap
     export BinaryMinMaxHeap, popmin!, popmax!, popall!
 
     export Trie, subtrie, keys_with_prefix, partial_path
@@ -112,5 +112,5 @@ module DataStructures
     include("splay_tree.jl")
 
     include("deprecations.jl")
-    @deprecate top(h::BinaryHeap) = first(h)
+@deprecate top(x) = first(h)
 end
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
I think it should get a comment to delete in like 6 months or may be a depwarn, but the sudden deletion was really the issue (a depwarn would be hard to avoid though)

oxinabox:
This should just be a deprecation.
Problem was as  @devmotion
 identified we screwed up the deprecation in 0.17.7

> I think the main issue was that DataStructures 0.17.17 deprecated top(x) (https://github.com/JuliaCollections/DataStructures.jl/blob/0d9cb40acc0e2898fb57022c54bf4432f495ce96/src/deprecations.jl#L3) but did not switch top(x::BinaryHeap) (and the other definitions for other heaps) to first(x::BinaryHeap) (https://github.com/JuliaCollections/DataStructures.jl/blob/0d9cb40acc0e2898fb57022c54bf4432f495ce96/src/heaps/binary_heap.jl#L151). So we never realized that it is deprecated and will be removed.


We just need to deprecate it to first  again, so people have time to update their code.

src/DataStructures.jl
```diff
@@ -112,4 +112,5 @@ module DataStructures
     include("splay_tree.jl")
 
     include("deprecations.jl")
+    const top = first
```
ChrisRackauckas, line 115:
```suggestion
    @deprecate top(h::BinaryHeap) = first(h)
```

ChrisRackauckas:
Bump please? This is as urgent as urgent gets.

src/DataStructures.jl
```diff
@@ -112,4 +112,5 @@ module DataStructures
     include("splay_tree.jl")
 
     include("deprecations.jl")
+    @deprecate top(h::BinaryHeap) = first(h)
```
devmotion, line 115: This is not sufficient since for other heaps the deprecation didn't work either:
```suggestion
@deprecate top(x) = first(h)
```
Probably should go into `src/deprecations.jl` though.

src/DataStructures.jl
```diff
@@ -25,7 +25,7 @@ module DataStructures
     export AbstractHeap, compare, extract_all!, extract_all_rev!
     export BinaryHeap, BinaryMinHeap, BinaryMaxHeap, nlargest, nsmallest
     export MutableBinaryHeap, MutableBinaryMinHeap, MutableBinaryMaxHeap
-    export heapify!, heapify, heappop!, heappush!, isheap
+    export heapify!, heapify, heappop!, heappush!, isheap, top
```
devmotion, line 28: This is not needed since `@deprecate` exports it already:
```suggestion
    export heapify!, heapify, heappop!, heappush!, isheap
```

devmotion (changes requested):

oxinabox:
Will merge and tags once CI passes

PR #673, opened 2020-09-01
Use custom orderings with heaps and nlargest/nsmallest

jlumpe:
* Added constructors to `BinaryHeap` and `MutableBinaryHeap` that take a `Base.Ordering` instance as the first argument. Previously the only way to use a specific ordering was as a type parameter, which only works if the ordering type is a singleton/has a constructor with no arguments.
* Modified `nlargest` and `nsmallest` signatures to take `by` and `lt` arguments as in `sort`. This shouldn't incur a performance penalty as I'm pretty sure Julia can fully specialize on keyword arguments, and the default argument values should statically resolve to calling `nextreme` with either `Base.Forward` or `Base.Reverse`.
* Updated docs page with instructions on how to use these features.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.18.2"
+version = "0.18.3"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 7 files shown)


oxinabox:
LGTM.
Can you change the version in the Project.toml (increment the last digit) so can tag a release?

oxinabox (approved):

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.18.2"
+version = "0.18.3"
```
```suggestion
version = "0.18.5"
```

Issue #666, opened 2020-08-28
[Tracking] MultiDict API updates before v1.0


NHDaly:
I wanted to create this meta-issue to track all the API changes that we want to do for `MultiDict`, which are currently scattered around various other threads: 
- https://github.com/JuliaCollections/DataStructures.jl/pull/656 (MuttsDict constructors operate _only_ on elements/pairs, not on the aggregated underlying view)
- https://github.com/JuliaCollections/DataStructures.jl/issues/655 (Make `iterate` iterate over the pairs, not over the aggregated view)
- https://github.com/JuliaCollections/DataStructures.jl/issues/654 (Add `delete!(d, k, v)` to delete by pair)
- https://github.com/JuliaCollections/DataStructures.jl/issues/522#issuecomment-675171633

And to use this thread to discuss other changes we might make.

----

Current suggested changes:
- [ ] Make MultiDict API consistently operate over pairs, just like a normal Dict, except that multiple entires are allowed per key.
- [ ] Use a `Set{T}`  instead of `Vector{T}` for the multi-values ([?] -- see discussion below)

StephenVavasis:
Since you are trying to clean this up before the 1.0 release, you may also want to compare the public API for MultiDict to that of SortedMultiDict to look for incompatibilities or possible improvements to either.

NHDaly:
What do you think about using a `Dict{K,Set{V}}` instead of a `Dict{K,Vector{V}}` for storing the values?

Properties that I like:
- This would make lookup by pair algorithmically optimal (prevents a linear scan)
    - For membership tests: `(2=>3) in MultiDict(2=>3)`
    - For delete by pair: `delete!(md, 2, 3)`
- It fixes weird ordering problems with equality that we currently have:
    - ```julia
       julia> MultiDict(1=>1, 1=>10) == MultiDict(1=>10, 1=>1)
       false
       ```

Properties that are maybe less good?:
- This means you can't easily store _duplicates of identical pairs_, which you currently can:
    - ```julia
       julia> MultiDict(1=>1, 1=>1)[1]
       2-element Array{Int64,1}:
        1
        1
       ```
    - But I don't know if this is something that we actually care about? For me, when I wanted to use MultiDicts, it's because of the ability to store multiple values per key, but I personally don't need the ability to store multiple _identical_ values per key... Do others want this ability?
[...]

PR #676, opened 2020-09-05
MultiDict: Have `iterate()` iterate over all pairs, as if the dict actually allowed duplicates

NHDaly:
```julia
julia> collect(MultiDict(3=>3, 1=>2, 1=>4))
3-element Array{Pair{Int64,Int64},1}:
 3 => 3
 1 => 2
 1 => 4
```

### DIFF
--- a/src/multi_dict.jl
+++ b/src/multi_dict.jl
@@ -4,6 +4,7 @@ import Base: haskey, get, get!, getkey, delete!, pop!, empty!,
              insert!, getindex, length, isempty, iterate,
              keys, values, copy, similar,  push!,
              count, size, eltype, empty
+import Base.Iterators: flatten, repeated
 
 struct MultiDict{K,V}
     d::Dict{K,Vector{V}}
@@ -42,8 +43,8 @@ end
 ## Most functions are simply delegated to the wrapped Dict
 
 @delegate MultiDict.d [ haskey, get, get!, getkey,
-                        getindex, length, isempty, eltype,
-                        iterate, keys, values]
+                        getindex, isempty,
+                        keys]
 
 sizehint!(d::MultiDict, sz::Integer) = (sizehint!(d.d, sz); d)
 copy(d::MultiDict) = MultiDict(d)
@@ -64,6 +65,46 @@ function in(pr::(Tuple{Any,Any}), d::MultiDict{K,V}) where {K,V}
     k = convert(K, pr[1])
     v = get(d,k,Base.secret_table_token)
     (v !== Base.secret_table_token) && (pr[2] in v)
+end
+
+# TODO: For efficiency, we probably want a MultiKeySet to correspond to Dict's KeySet
+
+# TODO: I'm not actually sure if we need this eachindex?
+function Base.eachindex(md::MultiDict)
+    flatten((repeated(k, length(vs)) for (k,vs) in md.d))
+end
+Base.length(md::MultiDict) = sum(length(vs) for (_,vs) in md.d)
+# TODO: what should values() do?
+Base.values(md::MultiDict) = values(md.d)
+#function Base.values(md::MultiDict)
+#    flatten(vs for (_,vs) in md.d)
+#end
+
+Base.eltype(md::MultiDict{K,V}) where {K,V} = Pair{K,V}
+function Base.iterate(md::MultiDict)
+    i = iterate(md.d)
+    if i === nothing
+        return nothing
+    end
+    ((k,vs), state) = i
+    (v, vstate) = iterate(vs)  # Should be non-empty
+    return (k=>v, (state, ((k,vs), vstate)))
+end
+function Base.iterate(md::MultiDict, md_state)
+    state, ((k,vs), vstate) = md_state
+    i = iterate(vs, vstate)
+    if i === nothing
+        # Finished iterating vs, move on to the next key
+        i = iterate(md.d, state)
+        if i === nothing
+            return nothing
+        end
+        ((k,vs), state) = i
+        (v, vstate) = iterate(vs)  # Should be non-empty
+    else
+        (v, vstate) = i
+    end
+    return (k=>v, (state, ((k,vs), vstate)))
 end
 
 function pop!(d::MultiDict, key, default)


src/multi_dict.jl
```diff
@@ -4,6 +4,7 @@ import Base: haskey, get, get!, getkey, delete!, pop!, empty!,
              insert!, getindex, length, isempty, iterate,
              keys, values, copy, similar,  push!,
              count, size, eltype, empty
+import Base.Iterators: flatten, repeated
```
oxinabox, line 7: Can we change this to not use `import` and instead overload by qualified name?

src/multi_dict.jl
```diff
@@ -66,6 +67,46 @@ function in(pr::(Tuple{Any,Any}), d::MultiDict{K,V}) where {K,V}
     (v !== Base.secret_table_token) && (pr[2] in v)
 end
 
+# TODO: For efficiency, we probably want a MultiKeySet to correspond to Dict's KeySet
```
oxinabox, line 70: Could we use an `Accumulator` for this?

oxinabox:
I like this, what can i do to help advance it?
What questions do we need to answer?

src/multi_dict.jl
```diff
@@ -66,6 +67,46 @@ function in(pr::(Tuple{Any,Any}), d::MultiDict{K,V}) where {K,V}
...
+function Base.eachindex(md::MultiDict)
+    flatten((repeated(k, length(vs)) for (k,vs) in md.d))
+end
+Base.length(md::MultiDict) = sum(length(vs) for (_,vs) in md.d)
+# TODO: what should values() do?
+Base.values(md::MultiDict) = values(md.d)
+#function Base.values(md::MultiDict)
+#    flatten(vs for (_,vs) in md.d)
+#end
+
```
NHDaly, line 82: One remaining question is what should `values(md::MultiDict)` return? Should it iterate all of the (flattened) values? Or should it iterate over the vectors?

That is, which of these options is most consistent and useful:
```julia
julia> values(MultiDict(3=>3, 1=>2, 1=>4))
Base.ValueIterator for a Dict{Int64,Array{Int64,1}} with 2 entries. Values:
  [3]
  [2, 4]
```
or
```julia
julia> values(MultiDict(3=>3, 1=>2, 1=>4))
Base.ValueIterator for a MultiDict{Int64,Int64} with 3 entries. Values:
  3
  2
  4
```

For consistency with `iterate()` I would think it should be the latter. But for consistency with `getindex`, I could see an argument for the former.

In fact, for consistency with `getindex`, I can understand why iterate might return what it did before this PR!
I'm starting to wonder if maybe we actually want to provide like two different types of interfaces over this dict? One for the aggregated view and one for the pairs view?

But, I guess, if you just want the aggregated view, you could just use a `DefaultDict{K, Vector{V}}`... So i think continuing to pursue the pairs-based interface makes sense.

NHDaly:
Thanks @oxinabox! I guess the questions are part of the overall MultiDict redesign: https://github.com/JuliaCollections/DataStructures.jl/issues/666

The main remaining question is about whether MultiDict should allow duplicates or not. It seems that different languages take different approaches, including providing both options via different implementations. I think we could do something similar, and/or control this behavior via type parameter or something.

I don't necessarily think that should impact this PR though.

src/multi_dict.jl
NHDaly: FWIW, the `golang` multimap does indeed flatten the values (the second approach), despite its `Get(k)` function returning a vector. So I think we should do this same. this is what felt most intuitive to me.

https://github.com/jwangsadinata/go-multimap/blob/c29f3d7f33b68d8bcd82880a558e9f8eed58e2b3/setmultimap/setmultimap.go#L144-L154

src/multi_dict.jl
NHDaly: Oops. thanks. Fixed in latest commit.

src/multi_dict.jl
```diff
@@ -66,6 +67,46 @@ function in(pr::(Tuple{Any,Any}), d::MultiDict{K,V}) where {K,V}
     (v !== Base.secret_table_token) && (pr[2] in v)
 end
 
+# TODO: For efficiency, we probably want a MultiKeySet to correspond to Dict's KeySet
+
+# TODO: I'm not actually sure if we need this eachindex?
+function Base.eachindex(md::MultiDict)
+    flatten((repeated(k, length(vs)) for (k,vs) in md.d))
+end
```
NHDaly, line 75: And this is the other TODO..

I don't remember what i was referring to... For Base.Dict, it looks like `eachindex` just calls `keys()`:
https://github.com/JuliaLang/julia/blob/7e8f2c05cd39aec8d4349ec915bf5c68cdf4c072/base/abstractarray.jl#L208-L209

So i'm not really sure why i wrote it?

--------

EDIT: AH, this is a similar question to the below. It's because in order for `collect()` to work, `eachindex()` has to return the same number of elements. So given `md = MultiDict(1=>2, 3=>3, 1=>4)`, `eachindex(md)` must return (unordered) `[1,1,3]`. But currently, `keys()` delegates to the dict:

```julia
julia> keys(md)
Base.KeySet for a Dict{Int64,Array{Int64,1}} with 2 entries. Keys:
  3
  1

julia> collect(eachindex(md))
3-element Array{Int64,1}:
 3
 1
 1
```

I guess i should probably just change this so that _keys_ returns `[3,1,1]`, and then just use the standard delegation from eachindex => keys. I think `keys()` should certainly include duplicates, because that's sort of the whole thing about a MultiDict.

So that needs to change too before this is finished.

NHDaly:
Also also, we need to update the tests

NHDaly:
I wrote above:

> I'm starting to wonder if maybe we actually want to provide like two different types of interfaces over this dict? One for the aggregated view and one for the pairs view?


And oh, wow, i'm only JUST NOW noticing the `enumerateall(md::MultiDict)` at the bottom of this file. It looks like this was exactly intended to provide the pair-wise view over the MultiDict. there are a few "pairwise" functions that it provided:
- `enumerateall()`: allows iterating the pairs (as we've done in this PR)
- `count()`: return the total number of inserted pairs, whereas `length()` was returning the total number of unique keys.

But `enumerateall()` no longer works. I'll probably just delete it since we're replacing it in this PR 👍 

I do see the merit in this idea, but as I stated above, I think that for a MultiDict type, we should treat everything as if it really was storing multiple entires that just happen to have the same key. That is consistent with Golang's and Python's multidicts, and i think it makes sense.

NHDaly:
K, i've updated some of the above, as long as it was all cached-in to my brain. Heading to sleep now :) Night!

oxinabox:
> , I think that for a MultiDict type, we should treat everything as if it really was storing multiple entires that just happen to have the same key. 

I agree

PR #677, opened 2020-09-12
Rename enqueue! and dequeue! to push! and popfirst!

### LINKED ISSUE #296: Use `push!` and `pop!` for `PriorityQueue` etc
We should use `push!` and `pop!` instead of `enqueue!` and `dequeue!`, 
so that different data structures are interchangeable without changing code.

cf. #289

adamslc:
This is currently annoying me, so I figured I would send a patch.

I also added `popfirst!` to `Stack` so that the core interface is consistent
between the two structures.

Fixes #296.

### DIFF
--- a/src/queue.jl
+++ b/src/queue.jl
@@ -4,6 +4,27 @@
     Queue{T}([blksize::Integer=1024])
 
 Create a `Queue` object containing elements of type `T`.
+
+# Examples
+```jldoctest
+julia> q = Queue{Int}()
+Queue{Int64}(Deque [Int64[]])
+
+julia> push!(q, 1)
+Queue{Int64}(Deque [[1]])
+
+julia> x = first(q)
+1
+
+julia> length(q)
+1
+
+julia> x = popfirst!(q)
+1
+
+julia> length(q)
+0
+```
 """
 mutable struct Queue{T}
     store::Deque{T}
@@ -20,21 +41,21 @@ Base.first(s::Queue) = first(s.store)
 Base.last(s::Queue) = last(s.store)
 
 """
-    enqueue!(s::Queue, x)
+    push!(s::Queue, x)
 
 Inserts the value `x` to the end of the queue `s`.
 """
-function enqueue!(s::Queue, x)
+function Base.push!(s::Queue, x)
     push!(s.store, x)
     return s
 end
 
 """
-    dequeue!(s::Queue)
+    popfirst!(s::Queue)
 
 Removes an element from the front of the queue `s` and returns it.
 """
-dequeue!(s::Queue) = popfirst!(s.store)
+Base.popfirst!(s::Queue) = popfirst!(s.store)
 
 Base.empty!(s::Queue) = (empty!(s.store); s)
 
--- a/src/stack.jl
+++ b/src/stack.jl
@@ -11,13 +11,13 @@ Stack{Int64}(Deque [Int64[]])
 julia> push!(s, 1) # push back a item
 Stack{Int64}(Deque [[1]])
 
-julia> x = top(s) # get a first item
+julia> x = first(s) # get a first item
 1
 
 julia> length(s)
 1
 
-julia> x = pop!(s) # get and remove a first item
+julia> x = popfirst!(s) # get and remove a first item
 1
 
 julia> length(s)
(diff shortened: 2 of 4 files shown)


src/queue.jl
```diff
@@ -4,6 +4,27 @@
...
+
+# Examples
+```jldoctest
+julia> q = Queue{Int}()
+Queue{Int64}(Deque [Int64[]])
+
+julia> push!(q, 1)
+Queue{Int64}(Deque [[1]])
+
+julia> x = top(q)
```
oxinabox, line 16: `top` is deprecated

```suggestion
julia> x = first(q)
```

adamslc: Sorry, should have caught that. I've fixed it in the Queue and Stack docstrings.

oxinabox (commented):
Ah, i am so sorry for not replying.
I wrote a reveiw a mont ago, but apparently forgot to press submit

oxinabox:
Its awkward here:

First we have that
`popfirst!` is conceptually defined on iterators to give the first element that would be iterated.
Similarly, `pop!` is conceptually defined on iterators to give the last element that would be iterated.
`popfirst!` seems the natural alias for `dequeue!` (at least on a queue).
No problem's there.

Applied to a stack though: a stack interates from the top down.
But `pop!` is defined on a stack to give the top item. Which is the first iterated, not the last

adamslc:
I agree that the whole thing is awkward. My objective in this PR is to create a pair of functions that can add and remove elements to both stacks and queues. In that light, I think `popfirst!` needs to be defined on a stack.

For `pop!` on a stack, I see three options:
1. Remove the definition, and create lots of confusion about why the standard CS terminology doesn't apply to Julia stacks.
2. Keep the definition as an alias for `popfirst!`, although this violated the conceptual definition of `pop!` in Base.
3. Re-implement `pop!` to do the right thing (remove the first item pushed), which will certainly create confusion.
Option 3 seems clearly the worst because it will generate difficult to find bugs. I would probably choose option 1, but then I won't have to deal with all of the Github issues that result.

As a radical alternative, we could create a new interface. Something like `push!`, `next!`, and `next`, that all these data structures respect. Then the data structure specific methods could be left in place.

oxinabox:
> I think popfirst! needs to be defined on a stack.

Why not `push!` on a queue?
Am i missing a thing?
I think `push!` is our more canoncial "add a thing to a collection" function for julia.
And out queue will already be a little odd because it uses `pushfirst!` and `pop!` not  `ensure!` and `dequeue!`.
I think i ma more ok with `push!(q::Queue, x) = pushfirst!(q::Queue)`
than `popfirst!(s::Stack) = pop!(s)`.

adamslc:
> Why not `push!` on a queue?
I think that this is needed as well, but seems less controversial. This PR implements this so that elements can be added to both stacks and queues using `push!`.

There also needs to be a unified interface to *get* an element these collections. That is where the `pop!` and `popfirst!` stuff come in.

oxinabox:
> There also needs to be a unified interface to get an element these collections. That is where the pop! and popfirst! stuff come in.

But here we are fine. You can always use `pop!` to get an element from a collection.
And if we did the above you can always use `push!` to put an element into a collection.
And the order if comes out again is determined by what type of collection it is.

HenriDeh:
Hello, I was about to open an issue on this matter. So instead I'm bumping this issue. I'm not familiar with the review process. Is there anything I can do to help this PR to merge?

HenriDeh:
I guess things that are missing are

1. Add tests.
2. Maybe instead of deleting enqueue! etc. It would be wiser to deprecate otherwise the changes will be breaking.
3. Update the documentation.

oxinabox:
>  1.0   Add tests.

That's done

>  2.  Maybe instead of deleting enqueue! etc. It would be wiser to deprecate otherwise the changes will be breaking.
>  3.  Update the documentation.

Correct, these need to be done.


The main thing that is blocking this PR is that it also adds `popfirst!` to stack.
Which is a much less clear cut question.
This is a classic "PR that does 2 things and one of those is less clearly right than the other, and so the whole PR is blocked"

It may be the @adamslc  has lost interest in this PR, (which is readonable).
So an alternative PR could be openned to address #296 could be made

HenriDeh:
Yeah I guess splitting into queue and stack PRs is a good idea. Okay I'll find some time to do at least queues if this one stays stale.

oxinabox:
Closed by #743

PR #687, opened 2020-09-30
Fix BinaryMaxHeap type alias

### LINKED ISSUE #686: BinaryMaxHeap{Int64}() isa BinaryMaxHeap{Int64} == false
In PR #673, the constructors of `BinaryHeap` were changed to instantiate `Base.Ordering` singleton types and then use those as parameters. This broke the following pattern:

```julia
# on 0.18.4
julia> foo(::BinaryMaxHeap) = "foo"
julia> foo(BinaryMaxHeap{Int}())
"foo"
```

```julia
# on 0.18.5
julia> foo(::BinaryMaxHeap) = "foo"
julia> foo(BinaryMaxHeap{Int}())
ERROR: MethodError: no method matching foo(::BinaryHeap{Int64,Base.Order.ReverseOrdering{Base.Order.ForwardOrdering}})
Closest candidates are:
  foo(::T) where T<:(BinaryHeap{T,Base.Order.ReverseOrdering} where T) at REPL[16]:1
Stacktrace:
 [1] top-level scope at REPL[17]:1
```

I noticed this when my tests started failing in NearestNeighborDescent.jl unexpectedly, since the compat is set to "0.18". Let me know if this was intentional / breaking.

oxinabox:
Fixed #686 

Would be cleaner if we had https://github.com/JuliaLang/julia/pull/37822

### DIFF
--- a/src/heaps/binary_heap.jl
+++ b/src/heaps/binary_heap.jl
@@ -50,9 +50,15 @@ BinaryHeap(ordering::Base.Ordering, xs::AbstractVector{T}) where T = BinaryHeap{
 BinaryHeap{T, O}() where {T, O<:Base.Ordering} = BinaryHeap{T}(O())
 BinaryHeap{T, O}(xs::AbstractVector) where {T, O<:Base.Ordering} = BinaryHeap{T}(O(), xs)
 
+const BasicReverseOrdering = Base.ReverseOrdering{Base.ForwardOrdering}
+
+# These constructors needed for BinaryMaxHeap, until we have https://github.com/JuliaLang/julia/pull/37822
+BinaryHeap{T, BasicReverseOrdering}() where {T} = BinaryHeap{T}(Base.Reverse)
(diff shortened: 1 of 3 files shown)


src/heaps/binary_heap.jl
```diff
@@ -50,9 +50,15 @@ BinaryHeap(ordering::Base.Ordering, xs::AbstractVector{T}) where T = BinaryHeap{
 BinaryHeap{T, O}() where {T, O<:Base.Ordering} = BinaryHeap{T}(O())
 BinaryHeap{T, O}(xs::AbstractVector) where {T, O<:Base.Ordering} = BinaryHeap{T}(O(), xs)
 
+const BasicReverseOrdering = Base.ReverseOrdering{Base.ForwardOrdering}
```
kmsquire, line 53: I don't feel too strongly, but how about `DefaultReverseOrdering`?

kmsquire (approved):

src/heaps/binary_heap.jl
oxinabox: Sure

PR #694, opened 2020-10-26
Rename DisjointSet and repair constructor

c-p-murphy:
Finally got around to putting this together—sorry for the delay.

### DIFF
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -1,4 +1,4 @@
-# Disjoint sets
+# Disjoint-Set
 
 ############################################################
 #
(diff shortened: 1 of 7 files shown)


c-p-murphy:
Also, I was on the fence about whether or not to hyphenate `Disjoint-Set`, but I ultimately decided to go with it ([Disjoint-set data structure](https://en.wikipedia.org/wiki/Disjoint-set_data_structure)). If people think that it would make more sense to have it as `Disjoint Set` though—just let me know.

src/disjoint_set.jl
```diff
@@ -8,41 +8,41 @@
...
 """
-mutable struct IntDisjointSets{T<:Integer}
+mutable struct IntDisjointSet{T<:Integer}
     parents::Vector{T}
     ranks::Vector{T}
     ngroups::T
 
     # creates a disjoint set comprised of n singletons
-    # IntDisjointSets(n::T) where {T<:Integer} = new(collect(Base.OneTo(n)), zeros(T, n), n)
+    # IntDisjointSet(n::T) where {T<:Integer} = new(collect(Base.OneTo(n)), zeros(T, n), n)
```
oxinabox, line 32: Can we delete this line?

oxinabox (commented):
I am going to need to review this carefuly to identify which lines changed and which were just renaming.

c-p-murphy:
Hey @oxinabox, I think this should be good to go now—thanks for taking the time to review this.

PR #707, opened 2020-12-02
New triggered tagbot

oxinabox:
trying to follow
https://discourse.julialang.org/t/ann-required-updates-to-tagbot-yml/49249
idk if the stuff with Documenter is right or not


timholy:
Closed by #715 (identical, presumably, though I don't know whether anyone had created a `DOCUMENTER_KEY` secret and corresponding deploy key.

PR #700, opened 2020-11-11
Rename DisjointSet and resolve errors

c-p-murphy:
Addresses #694 and resolves error when pushing a repeated element to a `DisjointSet`:

```julia
julia> ds = DisjointSets('a':'e')
DisjointSets{Char} with 5 elements:
  'a'
  'b'
  'c'
  'd'
  'e'

julia> ds.intmap
Dict{Char,Int64} with 5 entries:
  'a' => 1
  'c' => 3
  'd' => 4
  'e' => 5
  'b' => 2

julia> ds.internal
IntDisjointSets{Int64}([1, 2, 3, 4, 5], [0, 0, 0, 0, 0], 5)

julia> push!(ds, 'b')
'b': ASCII/Unicode U+0062 (category Ll: Letter, lowercase)

julia> ds.intmap
Dict{Char,Int64} with 5 entries:
  'a' => 1
  'c' => 3
  'd' => 4
  'e' => 5
  'b' => 6

julia> ds.internal
IntDisjointSets{Int64}([1, 2, 3, 4, 5, 6], [0, 0, 0, 0, 0, 0], 6)
```

Issue #579, opened 2020-02-10
Rename DisjointSets to DisjointSet ? 


oxinabox:
The fact that DisjointSets is a collection of sets under the hood is not relevent to the user.
The API is that of a `AbstractSet`.
The implementation is a implementation detail that is actually completely invisible to the user AFAICT

oxinabox:
@c-p-murphy did you ever make a PR to do this?
I can't find it

c-p-murphy:
Hey @oxinabox—no, sorry, things got hectic this past semester with everything going on in the world and I lost track of it. I'll try to finish putting something together over the next week or so and then submit a PR. Thanks for the reminder.

oxinabox:
Cool I might hold off releasing 0.18 for this so can get the deprecations in.
cf #577

oxinabox:
@c-p-murphy any update? I wouls still like to get this in before tagging 0.18

laborg:
Already fixed by https://github.com/JuliaCollections/DataStructures.jl/pull/700

PR #708, opened 2020-12-07
Remove disjoint_set

blegat:
Currently, both pages are available, one with `IntDisjointSet` and one with `IntDisjointSets` (notice the `s` at the end). The first one is incorrect and referenced from `docs/make.jl`, the second one is correct and referenced from `docs/src/index.md`. This PR removes the first one and reference the second one from the index.

**EDIT**: it seems to be leftovers from https://github.com/JuliaCollections/DataStructures.jl/pull/700/ cc @c-p-murphy


oxinabox:
thanks!

c-p-murphy:
@blegat thanks for catching this!

c-p-murphy:
I think this actually deleted the wrong one of the two docs pages (see #570, #578, #579, #694, #700), unless folks want to switch it back to being 'IntDisjointSets' and 'DisjointSets' again, that is. Just wanted to cite all the background for context as to how this came about. I'm just put together a new PR (#711) and bumped the version number so that this can be rolled out if there aren't any objections: 

https://github.com/JuliaCollections/DataStructures.jl/blob/e4de33937c38a820f91c836302e717300bdbb1f3/src/disjoint_set.jl#L26-L30

https://github.com/JuliaCollections/DataStructures.jl/blob/e4de33937c38a820f91c836302e717300bdbb1f3/src/disjoint_set.jl#L139-L158

https://github.com/JuliaCollections/DataStructures.jl/blob/e4de33937c38a820f91c836302e717300bdbb1f3/src/deprecations.jl#L6-L9

PR #710, opened 2020-12-09
bumped compat bounds

ExpandingMan:
Didn't change anything else and didn't get any test failures on my end.


oxinabox:
This literally doesn't change the bounds at all.

ExpandingMan:
That's confusing... so this was already allowing OrderedCollections 1.3?

ExpandingMan:
Yeah, it looks like your right.  Ok, I just find the Pkg compat bounds specifications incredibly confusing.  `"1.1.0"` really, really, realyl does not look like it allows 1.(n > 1)`.

PR #711, opened 2020-12-10
Fix docs mistake with respect to DisjointSet

c-p-murphy:
Put this together in response to #708. Original discussion didn't get much input, but I am definitely open to modifying this and/or creating another PR if others feel that changes ought to be made.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.19.0-DEV"
+version = "0.19.1-DEV"
 
 
 [deps]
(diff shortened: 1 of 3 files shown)


oxinabox:
Thanks sorry my bad about not checking carefully enough before.
We are not tagging any releases for 0.19.x until done.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.19.0-DEV"
+version = "0.19.1-DEV"
```
```suggestion
version = "0.19.0-DEV"
```
No need to increment bwhen version bus suffixes with DEV and the digit you would increment is currently 0

oxinabox (approved):

blegat:
Oops, sorry, I thought I checked that it now was plural but I apparently did it the wrong way ^^

Issue #639, opened 2020-06-29
BinaryHeap constructor performs different from heapify


AquaIndigo:
I noticed that BinaryHeap like BinaryMinHeap constructs a heap that just insert every element from an array to a empty heap. And the following `x` and `y` have the same order:

```julia
julia> nums = rand(1:20000, 2000);

julia> x = MutableBinaryMinHeap(nums);

julia> y = BinaryMinHeap(Int.([]));

julia> for i in nums 
           push!(y, i)
       end
```

However, there is an O(N) heap-building algorithm instead of O(nlogn), and function `heapify` is an implement. So why not use heapify instead of insert successively？

Sincerely.

AliMalik9599:
Hi @oxinabox @AquaIndigo I am interested in fixing this, but I am relatively new to Julia. Any tips on getting started? Thanks!

oxinabox:
I recommend asking on julialang [slack](https://slackinvite.julialang.org/) or [Discourse](https://discourse.julialang.org/).
The code change should be fairly straight forward if you are familar with the constructor in question and the `heapify` function

zerefwayne:
@AliMalik9599 Are you still working on this issue?

PR #712, opened 2020-12-10
Optimisation: Use heapify instead of iteratively inserting elements

rushabh-v:
Fixes: https://github.com/JuliaCollections/DataStructures.jl/issues/639

### DIFF
--- a/src/heaps/mutable_binary_heap.jl
+++ b/src/heaps/mutable_binary_heap.jl
@@ -1,5 +1,7 @@
 # Binary heap
 
+include("arrays_as_heaps.jl")
+
 struct MutableBinaryHeapNode{T}
     value::T
     handle::Int
@@ -131,10 +133,10 @@ end
 
 function _make_mutable_binary_heap(ord::Ordering, ty::Type{T}, values) where T
     # make a static binary index tree from a list of values
-
     n = length(values)
     nodes = Vector{MutableBinaryHeapNode{T}}(undef, n)
     nodemap = Vector{Int}(undef, n)
+    values = heapify(values, ord)
 
     i::Int = 0
     for v in values
@@ -143,9 +145,6 @@ function _make_mutable_binary_heap(ord::Ordering, ty::Type{T}, values) where T
         @inbounds nodemap[i] = i
     end
 
-    for i = 1 : n
-        _heap_bubble_up!(ord, nodes, nodemap, i)
-    end
     return nodes, nodemap
 end
 


oxinabox:
@AquaIndigo can you review

oxinabox:
CI fails so i suspect this is not correct

rushabh-v:
Turned out I haven't updated `nodemap` after calling `heapify`. I am wondering if we can do that efficiently (without using extra space). 

Or does the issue intend to use `heapify` at some other place @AquaIndigo? (That was the confusion that I mentioned [here](https://github.com/JuliaCollections/DataStructures.jl/pull/712#issue-536128749))

AquaIndigo:
> Turned out I haven't updated `nodemap` after calling `heapify`. I am wondering if we can do that efficiently (without using extra space).
> 
> Or does the issue intend to use `heapify` at some other place @AquaIndigo? (That was the confusion that I mentioned [here](https://github.com/JuliaCollections/DataStructures.jl/pull/712#issue-536128749))

I think that `heapify!` could help.

test/test_mutable_binheap.jl
```diff
@@ -52,11 +52,12 @@ end
 
 @testset "MutableBinheap" begin
 
-    vs = [4, 1, 3, 2, 16, 9, 10, 14, 8, 7]
-    vs2 = collect(enumerate(vs))
     ordering = Base.Order.By(last)
 
     @testset "construct heap" begin
+        vs = [4, 1, 3, 2, 16, 9, 10, 14, 8, 7]
+        vs2 = collect(enumerate(vs))
```
rushabh-v, line 59: The new implementation was overwriting `vs`. So, the tests were failing because of updated `vs` (The value of `vs` was not as expected because it was updated during some other test). So, I updated it to use separate `vs` and `vs2` for each test.

test/test_mutable_binheap.jl
```diff
@@ -99,17 +101,20 @@ end
...
     @testset "make mutable binary maxheap" begin
+        vs = [4, 1, 3, 2, 16, 9, 10, 14, 8, 7]
         h = MutableBinaryMaxHeap(vs)
 
         @test length(h) == 10
         @test !isempty(h)
         @test first(h) == 16
         @test isequal(list_values(h), vs)
-        @test isequal(heap_values(h), [16, 14, 10, 8, 7, 3, 9, 1, 4, 2])
+        @test isequal(heap_values(h), [16, 14, 10, 8, 7, 9, 3, 2, 4, 1])
```
rushabh-v, line 111: I think this is just due to the difference in the implementation of `heapify` and the previous method `_heap_bubble_up`. Otherwise, there is no problem. (Let me know if that's not the case)

rushabh-v:
I ran the tests locally, it was all green.

test/test_mutable_binheap.jl
oxinabox: It is not OK for `MutableBinaryMinHeap(vs)` to mutate `vs`.
We should add a `copy` to the constructor if that is the case.

(we can consider later adding a `MutableBinaryMinHeap!` or something that does do so. but lets not worry about it in this PR)

test/test_mutable_binheap.jl
oxinabox: Do these both follow the heap rule about being larger than children?
(I don't know how to read heaps when written in array form TBH)

test/test_mutable_binheap.jl
```diff
@@ -121,6 +126,7 @@ end
     end
 
     @testset "hmin / push! / pop!" begin
+        vs = [4, 1, 3, 2, 16, 9, 10, 14, 8, 7]
```
oxinabox, line 129: this is a good change anyway. Testsets hould be indipendent.
Though maybe we move this down to after `ss` is declared?

oxinabox:
Sorry for taking so long to review this.
I lost track of it

test/test_mutable_binheap.jl
rushabh-v: `list_values` related tests (i.e., [this one](https://github.com/JuliaCollections/DataStructures.jl/blob/master/test/test_mutable_binheap.jl#L96)) started failing after adding copy in the constructor. the reason for that is because `nodemap` is not updated as I mentioned [here](https://github.com/JuliaCollections/DataStructures.jl/pull/712#issuecomment-742761512). Those tests were not failing before because `vs` was being mutated.

I am trying to think if there is an efficient way to update `nodemap` after the `heapify!` call.

It might be that not being able to update `nodemap` efficiently (with O(n) implementation) is the reason they implemented it in O(nlogn) in the first place.

test/test_mutable_binheap.jl
rushabh-v: Yes, they both are valid heaps. I also confirmed it using this function [here](https://www.geeksforgeeks.org/how-to-check-if-a-given-array-represents-a-binary-heap/#:~:text=An%20Efficient%20Solution%20is%20to,%2C%20then%20x%20%3E%20z).

test/test_mutable_binheap.jl
rushabh-v: Yes, they both are valid heaps. I also confirmed it using this function [here](https://www.geeksforgeeks.org/how-to-check-if-a-given-array-represents-a-binary-heap/#:~:text=An%20Efficient%20Solution%20is%20to,%2C%20then%20x%20%3E%20z).

oxinabox: seems fine then

PR #720, opened 2021-01-18
Fix MutableBinaryHeap, issue similar to #686

dm3:
Issue with `Base.Order.ReverseOrdering`. Mirror of #686, only related to a `MutableBinaryHeap` implementation. Depends on https://github.com/JuliaLang/julia/pull/37822.

### DIFF
--- a/src/heaps/arrays_as_heaps.jl
+++ b/src/heaps/arrays_as_heaps.jl
@@ -2,6 +2,7 @@
 
 using Base.Order: Forward, Ordering, lt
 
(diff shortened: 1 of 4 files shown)


oxinabox (approved):
Good catch.
Do you need this backported so the fix is available in 0.18?
I don't know when 0.19 will be released.

dm3:
Sure, would be nice to have in 0.18!

oxinabox:
ok, I will make it so

Issue #719, opened 2021-01-11
Creating a DisjointSets forest from a single symbol results in an infinite type loop.


0x0f0f0f:
```
julia> DisjointSets(:a)
^CERROR: InterruptException:
Stacktrace:
[...]

PR #723, opened 2021-01-21
Pull Request for Issue #719

NehaDalmia:
Added the clause to check if a type is Scalar before iterating it.

### DIFF
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -1,4 +1,4 @@
-# Disjoint-Set
+#Disjoint-Set
 
 ############################################################
 #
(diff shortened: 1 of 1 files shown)


src/disjoint_set.jl
```diff
@@ -1,4 +1,4 @@
-# Disjoint-Set
+#Disjoint-Set
```
eulerkochy, line 1:
```suggestion
# Disjoint-Set
```

NehaDalmia:
The bug was arising as it was attempting to iterate over a non-iterable Symbol object. The Scalar struct declared is used to determine whether a given variable is iterable or not, if not we create a DisjointSet having just one element of the non iterable type. If it is iterable , we iterate over its members and the DisjointSet is created with these members.

PR #730, opened 2021-03-26
Document Default*Dict and add Default*Dict{K,V} constructors

mbauman:
I wanted both of these things (the documentation and these particular constructors) this evening. I didn't document _all_ the available constructors, just the four that I wanted as I think those are the most commonly tread path.


oxinabox:
LGTM just needs tests.

oxinabox:
sorry for not merging this before, it got lost.
master is currently 0.19-dev and is not planning a release until we can gut all the 1.0 changes in (modulo deprecations).
If you want this sooner, we can back port it to 0.18

PR #733, opened 2021-04-21
Add recursive `DefaultDict` example

omus:
Seems like a useful example to include. Alternatively I could see a `DefaultDictRec` being introduced which could be an abstraction on this idea and would allow for dynamically setting the type parameters.


oxinabox (approved):

oxinabox:
thanks sorry about the slow merge.

Issue #735, opened 2021-04-27
Implement constructor for Deque from iterable


jordancluts:
At present a Deque (and Stack/Queue) can only be created empty.

A constructor method should be added to enable the creation of a Deque from an iterable such as an array.

`Deque([1,2,3])` which currently errors.

A few issues to discuss
- What Deque Block Size to choose by default when given an array/iterable?
- Should a user be able to pass both an array and a blocksize? (at the moment `Deque(x::Int)` creates an empty Deque with block size of x). We could add a `Deque(array,blocksize)` method or similar to allow specification.

oxinabox:
I think
I think we should have the 2 arg one where you pass the blocksize.
and the 1 arg version that just takes it iterator,  and the default blocksize should equal to the length of the iterator.
which we probably need to get by `collect`ing it and then taking the length (since notall itrerators have length, but we are going to collect them anyway)

Something like:
```julia
Deque(x::Vector, blocksize=length(x)) =  # Main constructor
Deque(iter, blocksize) = Deque(vec(collect(iter)), blocksize)
Deque(iter) = Deque(vec(collect(iter)))
```

jordancluts:
I completely agree.
There may be some argument for making the initial block size equal to some larger fraction of the initial vector. This prevents another allocation on the first `push!`. Alternatively, the "unrolled linked list" that the Deque is currently built upon is designed for cache efficiency so choosing a block size related to cache sizes would be optimal from a performance perspective. Doing so automatically would require querying the CPU for info (l3 cache sizes and cache row sizes) when the package is loaded and then setting a variable to use in construction accordingly. That however is a much bigger task.

I was messing around with various things and arrived at basically the same solution as you but I decided to start with CircularBuffer as I thought it's implementation was good and it's a more basic data structure. Other data structures can then be updated to match and/or be built upon the more fundamental piece. Expect the PR for CircularBuffer soon.

PR #736, opened 2021-04-28
Circ buff new constructor

jordancluts:
Adds constructors to Circular Buffer allowing a vector or other iterable to be in the buffer once initialized. Capacity is still required to be provided by the user and errors if the given capacity is less than the length of the given iterable. If no type parameter is given the eltype of the provided iterable is used. If one is provided the provided iterable is converted to that type (if possible, errors otherwise).

Existing constructors have changed slightly via new type restrictions.

Allows:
`CircularBuffer(1:5,10)`
`CircularBuffer{Float64}([1,2],100)`
`CircularBuffer((1,2.0,"hi"),20)`

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -13,9 +13,9 @@ mutable struct CircularBuffer{T} <: AbstractVector{T}
     length::Int
     buffer::Vector{T}
 
-    function CircularBuffer{T}(vec,capacity::Int) where {T}
-        capacity >= length(vec) || throw(BoundsError()) # prevent overflow
-        new{T}(capacity,1,length(vec),copyto!(Vector{T}(undef,capacity),convert(Vector{T},vec)))
+    function CircularBuffer{T}(iter, capacity::Int) where {T}
+        vec = copyto!(Vector{T}(undef,capacity), vec)
+        new{T}(capacity, 1, length(vec),vec)
     end
 
 end
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -14,11 +14,16 @@ mutable struct CircularBuffer{T} <: AbstractVector{T}
     length::Int
     buffer::Vector{T}
 
+    function CircularBuffer{T}(f,len,buf) where {T}
+        f <= length(buf) || throw(ArgumentError("Value of 'first' must be inbounds of buffer"))
+        len <= length(buf) || throw(ArgumentError("Value of 'length' must be <= length of buffer"))
+        return new{T}(length(buf), f, len, buf)
+    end
 end
 
 function CircularBuffer{T}(iter, capacity::Int) where {T}
     vec = copyto!(Vector{T}(undef,capacity), iter)
-    CircularBuffer{T}(capacity, 1, length(vec),vec)
+    CircularBuffer{T}(1, length(iter),vec)
 end
 
 CircularBuffer(capacity::Int) = CircularBuffer{Any}(capacity)
@@ -29,7 +34,7 @@ CircularBuffer(iter,capacity::Int) =  CircularBuffer{eltype(iter)}(iter,capacity
 
 function CircularBuffer{T}(iter) where {T}
   vec = reshape(collect(T,iter),:) 
-  CircularBuffer{T}(length(vec), 1, length(vec), vec)
+  CircularBuffer{T}(1, length(vec), vec)
 end
 
 CircularBuffer(iter) = CircularBuffer{eltype(iter)}(iter)
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -1,11 +1,11 @@
 """
-    CircularBuffer{T}(n)
+    CircularBuffer{T}(v,n)
 
 The CircularBuffer type implements a circular buffer of fixed capacity
 where new items are pushed to the back of the list, overwriting values
 in a circular fashion.
 
-Allocate a buffer of elements of type `T` with maximum capacity `n`.
+Allocate a buffer of elements of type `T`  containing items `v` with maximum capacity `n`.
 """
 mutable struct CircularBuffer{T} <: AbstractVector{T}
     capacity::Int
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -14,7 +14,7 @@ mutable struct CircularBuffer{T} <: AbstractVector{T}
     buffer::Vector{T}
 
     function CircularBuffer{T}(iter, capacity::Int) where {T}
-        vec = copyto!(Vector{T}(undef,capacity), vec)
+        vec = copyto!(Vector{T}(undef,capacity), iter)
         new{T}(capacity, 1, length(vec),vec)
     end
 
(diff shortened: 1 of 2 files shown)


jordancluts:
Also note that per the discussion in #735 a method could be added that doesn't require a capacity to be given. Code would be 

`CircularBuffer(vec) = CircularBuffer(vec,length(vec))`

Or some other metric to automatically determine the capacity as desired.

src/circular_buffer.jl
```diff
@@ -1,22 +1,31 @@
...
+        capacity >= length(vec) || throw(BoundsError()) # prevent overflow
+        new{T}(capacity,1,length(vec),copyto!(Vector{T}(undef,capacity),convert(Vector{T},vec)))
+    end
+
 end
 
-CircularBuffer(capacity) = CircularBuffer{Any}(capacity)
+CircularBuffer(capacity::Int) = CircularBuffer{Any}(capacity)
+
+CircularBuffer{T}(capacity::Int) where {T} = CircularBuffer{T}([],capacity)
```
oxinabox, line 25: We can avoid a call to `convert` and an allocation here by intitalizing the buffer to empty of the right type
```suggestion
CircularBuffer{T}(capacity::Int) where {T} = CircularBuffer{T}(T[],capacity)
```

oxinabox (approved):
Basic concept is right, but I suggest a bit of a rework

src/circular_buffer.jl
```diff
@@ -1,22 +1,31 @@
...
 """
 mutable struct CircularBuffer{T} <: AbstractVector{T}
     capacity::Int
     first::Int
     length::Int
     buffer::Vector{T}
 
-    CircularBuffer{T}(capacity::Int) where {T} = new{T}(capacity, 1, 0, Vector{T}(undef, capacity))
+    function CircularBuffer{T}(vec,capacity::Int) where {T}
+        capacity >= length(vec) || throw(BoundsError()) # prevent overflow
```
oxinabox, line 17: We don't need to guard gainst this. `copyto!` guards against it for us.
Which si good because that means we can argange to avoid calling `length` on an iterator which might not have `length` defined`

src/circular_buffer.jl
```diff
@@ -1,22 +1,31 @@
...
+        new{T}(capacity,1,length(vec),copyto!(Vector{T}(undef,capacity),convert(Vector{T},vec)))
+    end
```
oxinabox, line 19: We don't need to guard on bounds. `copyto!` guards against it for us.
and it also calls `convert` for us, per element so we don't need to allocate for that if converting.
We want to call `length` on the `vec` rather than the iterator
as an arbitary iterator which might not have `length` defined`
```suggestion
    function CircularBuffer{T}(iter, capacity::Int) where {T}
        vec = copyto!(Vector{T}(undef,capacity), vec)
        new{T}(capacity, 1, length(vec),vec)
    end
```

jordancluts:
I have committed those changes you recommended and changed the constructor variable name to "iter" instead of "vec" at the correct locations.

I was going to add the constructor that does not take an explicit capacity but it is unclear what to do if the provided iterator does not have a known length (perhaps we just error in that situation?).

If you would like a full rework of the PR that is fine as well but I may need a few pointers as to what I should do differently for the next iteration.

Thanks for the help!

oxinabox:
> I was going to add the constructor that does not take an explicit capacity but it is unclear what to do if the provided iterator does not have a known length (perhaps we just error in that situation?).

We will know the length before it is constructed though.
Since we are going to convert it to a vector in order to store it.
and those do have known length.
So I guess for that case we would have it call `vec = collect(iter)` (or maybe `vec = convert(Vector{T}, iter)` does that work? maybe we need `vec = collect(convert(T, x) for x in iter)` ?)

oxinabox (approved):

jordancluts:
> So I guess for that case we would have it call `vec = collect(iter)` (or maybe `vec = convert(Vector{T}, iter)` does that work? maybe we need `vec = collect(convert(T, x) for x in iter)` ?)

according to the docs
```
  If T is a collection type and x a collection, the result of convert(T, x) may alias
  all or part of x.

  julia> x = Int[1, 2, 3];

  julia> y = convert(Vector{Int}, x);

  julia> y === x
  true
```

I'm fairly certain we don't want the possibility of aliasing here so we need to do more than `vec = convert(Vector{T},iter)`

Conveniently `collect` has a conversion argument already so one can do `vec = collect(T,iter)` directly. This makes the implementation easy. It'll just be

```function CircularBuffer{T}(iter)
  vec = collect(T,iter)  
  CircularBuffer{T}(vec,length(vec))
```
and `CircularBuffer(iter) = CircularBuffer{eltype(iter)}(iter)`

jordancluts:
A drat, this has an efficiency, because we are constructing the array we need but then performing a `copyto!` when we call the inner constructor. The "capacity-less" constructor needs to be able to construct directly. Maybe the best idea is to have no inner constructor so the two different methods can directly build the struct.

jordancluts:
Wasn't too bad, just need a `reshape(vec,:)` and then it should avoid a copy operation. I also added a missing `where {T}` to the capacity-less constructor. I think this should be good to review again now when you have a chance @oxinabox 

Thanks again for your help. I'm hoping that if I get one figured out it'll be easy to do the changes on the similar structures (the deques, queues, stacks, etc.)

src/circular_buffer.jl
```diff
@@ -1,22 +1,38 @@
...
+Allocate a buffer of elements of type `T`  containing items `v` with maximum capacity `n`.
+If no capacity is provided the capacity is the number of elements in `v`
 """
 mutable struct CircularBuffer{T} <: AbstractVector{T}
     capacity::Int
     first::Int
     length::Int
     buffer::Vector{T}
 
-    CircularBuffer{T}(capacity::Int) where {T} = new{T}(capacity, 1, 0, Vector{T}(undef, capacity))
```
oxinabox, line 16: Possibly we still want an inner constructor her to check that 
`first <= length <= capacity` and that `capacity==length(buffer)` (possibly capacity shouldn't be passed in here now, just `first`, `length`, and `buffer` ?

oxinabox (commented):
looks good.
One last comment on ensuring consistency via inner constructor then should be good to merge.
Thanks for working though all this

src/circular_buffer.jl
jordancluts: Apologies for the delay, work and the holiday weekend have kept me busy. I hope to get this wrapped up this week. Thanks for the mentorship throughout this process.

I agree that the "internal" interface should be passing the vector, the first index, and the length, but skipping the capacity. There isn't anything useful one can do by passing a capacity different from length(vector). Whereas atleast conceivably by carefully constructing a vector and setting first/length correctly you could directly create an instance of a buffer with any internal structure that you could arrive at via push/pop. I'll implement that internal constructor with the checks, and make the few other edits needed to the other constructors to match that new format in the next few days.

oxinabox: cool thanks, sorry for my own slowness

jordancluts:
@oxinabox I finally found an hour to hopefully finish this off. I implemented the inner constructor with guards, changed the outer constructors to match the new inner constructor, and added new tests for the inner constructor. All tests pass on my machine with these changes applied to 0.18.9.

jordancluts:
Added a test to cover the missed constructor, now should be okay on code coverage. (Although I'm not wildly familiar with code coverage so I may be wrong).

jordancluts:
@oxinabox just a reminder to take a look at this again when you have a chance and let me know if you want any other changes (if you're still interested in the changes), thanks!

oxinabox (approved):

oxinabox:
thanks! sorry it's been so busy lately

PR #747, opened 2021-08-09
inline a bit to remove an inference trigger

ChrisRackauckas:


oxinabox:
LGTM.
I will forward-port it on to master as well

PR #748, opened 2021-08-09
ForwardPort: inline a bit to remove an inference trigger

oxinabox:
This forward-ports @ChrisRackauckas  PR #747 

It does not require review as that PR was already reviewed.
Just running it here to be sure CI doesn't fail


oxinabox:
windows failure is unrelated

PR #756, opened 2021-09-11
Improvements of Trie Structure

lucidfrontier45:
This PR improved `Trie` with the following two points.

1. fix `TrieIterator` to correctly handle non-ascii characters
2. add `find_prefixes` function which is equivalent to `prefixes` method of https://github.com/pytries/marisa-trie

### DIFF
--- a/docs/src/trie.md
+++ b/docs/src/trie.md
@@ -31,3 +31,10 @@ given string, use:
 ```julia
 seen_prefix(t::Trie, str) = any(v -> v.is_key, partial_path(t, str))
 ```
+
+`find_prefixes` can be used to find all keys which are prefixes of the given string.
+
+```julia
+t = Trie(["A", "ABC", "ABCD", "BCE"])
+find_prefixes(t, "ABCDE") # "A", "ABC"
+```
--- a/docs/src/trie.md
+++ b/docs/src/trie.md
@@ -36,5 +36,5 @@ seen_prefix(t::Trie, str) = any(v -> v.is_key, partial_path(t, str))
 
 ```julia
 t = Trie(["A", "ABC", "ABCD", "BCE"])
-find_prefixes(t, "ABCDE") # "A", "ABC"
+find_prefixes(t, "ABCDE") # "A", "ABC", "ABCD"
 ```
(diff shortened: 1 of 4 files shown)


src/trie.jl
```diff
@@ -111,14 +111,27 @@ end
...
         t = t.children[it.str[i]]
-        return (t, (t, i + 1))
+        return (t, (t, nextind(it.str, i)))
     end
 end
 
 partial_path(t::Trie, str::AbstractString) = TrieIterator(t, str)
 Base.IteratorSize(::Type{TrieIterator}) = Base.SizeUnknown()
+
+function find_prefixes(t::Trie, str::AbstractString)
```
oxinabox, line 126: this requires a docstring

docs/src/trie.md
```diff
@@ -31,3 +31,10 @@ given string, use:
 ```julia
 seen_prefix(t::Trie, str) = any(v -> v.is_key, partial_path(t, str))
 ```
+
+`find_prefixes` can be used to find all keys which are prefixes of the given string.
+
+```julia
+t = Trie(["A", "ABC", "ABCD", "BCE"])
+find_prefixes(t, "ABCDE") # "A", "ABC"
```
oxinabox, line 39: why does this not return `ABCD` as well?

oxinabox:
Thanks.

Note that we are currently feature frozen pending #479 
and will not be making a 0.19 release with new features until that is done.
(something that I personally do not have time to do, so it is waiting for people who do)


The _fix TrieIterator to correctly handle non-ascii characters_ change is a good bug fix,  which we can backpoint to the 0.18.x line.
`find_prefixes` is not.

src/trie.jl
lucidfrontier45: added in the latest commit

docs/src/trie.md
lucidfrontier45: That's my mistake. I fixed it.

lucidfrontier45:
@oxinabox 

Thank you for your comment. I updated my commits according to your suggestion.
I also understood the release schedule. I don't have any problem with it.

oxinabox (approved):

PR #759, opened 2021-10-04
Allow non-string indices for Trie

jonas-schulze:
by dropping the `::AbstractString` restriction. I considered replacing that by `::Union{AbstractString,AbstractVector}` (thereby wondering why `AbstractString <: AbstractVector` doesn't hold). Which one would you prefer?

This is breaking as the original type parameter (denoting the value type) is now in second position, such that the parameters follow the common key-value pattern like e.g. `Dict` has. I noticed that you are feature-frozen (#479), but I think it would be better to consider changing the order now instead of adding this feature later in a non-breaking manner (i.e. reversing the type parameters).

This is ready for comments. If there is anything I can change to improve this PR, I'm happy to do so. 🙂

### DIFF
--- a/src/trie.jl
+++ b/src/trie.jl
@@ -1,25 +1,21 @@
-mutable struct Trie{T}
-    value::T
-    children::Dict{Char,Trie{T}}
+mutable struct Trie{K,V}
+    value::V
+    children::Dict{K,Trie{K,V}}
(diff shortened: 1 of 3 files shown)


oxinabox:
This seems sensible.
I will try and find time to review properly later this week

PR #763, opened 2021-10-16
Improve LinkedList `Base.map` performance by not having to reverse the list

NHDaly:
I was curious about the Linked List implementation so I did a bit of profiling it, and I was surprised to see that `map()` was constructing a list twice:
Screen Shot 2021-10-16 at 10 58 41 AM
Screen Shot 2021-10-16 at 10 58 54 AM

So I first implemented it as the tail recursive map for linked lists (in fa11333), which improved things quite a bit:
Before:
```julia
julia> @Btime DataStructures.map(x->x*2, $(list((1:1000)...)));
  148.476 μs (6469 allocations: 163.55 KiB)
```
After:
```julia
julia> @Btime DataStructures.map(x->x*2, $(list((1:1000)...)));
  29.705 μs (1745 allocations: 42.89 KiB)
```
But that has this silly (if adorable) recursive call stack, which can of course StackOverflow (I had forgotten that julia [doesn't have Tail Call Optimization](https://discourse.julialang.org/t/tail-call-optimization-and-function-barrier-based-accumulation-in-loops/25831/12)):
Screen Shot 2021-10-16 at 10 58 12 AM
Screen Shot 2021-10-16 at 10 58 48 AM

So I rewrote it into an iterative function, and magically that actually made it even faster still. I'm not exactly sure why, but i'll take it. :)

Before:
```julia
julia> @Btime DataStructures.map(x->x*2, $(list((1:1000)...)));
  29.705 μs (1745 allocations: 42.89 KiB)
```
After:
```julia
julia> @Btime DataStructures.map(x->x*2, $(list((1:1000)...)));
  13.139 μs (1011 allocations: 47.65 KiB)
```
Screen Shot 2021-10-16 at 11 37 21 AM
Screen Shot 2021-10-16 at 11 46 00 AM

Also, this let's us maintain the property that was added in PR #27, to allow `map()` to create a list with a different eltype.

----------------

(Note: All measurements taken on julia 1.6.3; visualizations with PProf.jl)

### DIFF
--- a/src/list.jl
+++ b/src/list.jl
@@ -79,7 +79,9 @@ function Base.map(f::Base.Callable, l::Cons{T}) where T
         push!(stack, head(l))
         l::Cons{T} = tail(l)::Cons{T}
     end
-    l2 = list(f(head(l)))  # Note this might have a different eltype than T
+    # Note the new list might have a different eltype than T
+    first = f(head(l))
+    l2 = cons(first, nil(typeof(first) <: T ? T : typeof(first)))
     for i in reverse(1:length(stack))
         l2 = cons(f(stack[i]), l2)
     end


NHDaly:
The coverage failure is flaky noise:
Screen Shot 2021-10-16 at 11 57 24 AM

PR #764, opened 2021-10-16
Remove `mutable` from LinkedList structs.

NHDaly:
These types have always been meant to be treated as immutable structs,
but I think they were created long enough ago (https://github.com/JuliaCollections/DataStructures.jl/commit/5873750e976aa572b58d74da764478ec8c69e96f) that
 there wasn't an immutable/mutable distinction? o.O

And they were converted from `type` to `mutable struct` in an automatic
conversion via femtocleaner, here: https://github.com/JuliaCollections/DataStructures.jl/pull/321

In any case, these are clearly meant to be treated as immutable structs
and if they were created today they wouldn't be marked mutable, so this
fixes that.

There's no real performance impact, since you of course have to heap
allocate all the inner nodes anyway, but one can imagine that maybe
someone maintains hundreds or thousands of very small lists, and getting
to save that one extra allocation for the `head` maybe helps some. :)

This is mainly a theoretical cleanup.

### DIFF
--- a/src/list.jl
+++ b/src/list.jl
@@ -2,10 +2,9 @@ abstract type LinkedList{T} end
 
 Base.eltype(::Type{<:LinkedList{T}}) where T = T
 
-mutable struct Nil{T} <: LinkedList{T}
-end
+struct Nil{T} <: LinkedList{T} end
 
-mutable struct Cons{T} <: LinkedList{T}
+struct Cons{T} <: LinkedList{T}
     head::T
     tail::LinkedList{T}
 end


oxinabox:
I very much want to just delete LinkedList from this package before 1.0

NHDaly:
Haha yeah, that's true. It's still an interesting / neat example of writing a data structure in julia. It's a super classic CS example, so maybe still neat to keep around? Or if you want to delete it, we could pull it into another package?

oxinabox:
I want to push  both our linked list implementations into FunctionalDataStructures.jl, which I suspect already has it.
It is code that Jeff originally wrote (it was one of the first examples in JuliaLang/julia).
Jeff gave his blessing to moving it there.

oxinabox:
Current proposal is to just delete these data structures before 1.0 release.
Pushing them out into FunctionalDataStructures.jl

KristofferC:
Ok, can this be merged in the meantime? Also, why should they be removed from the package?

KristofferC (approved):

oxinabox:
> Also, why should they be removed from the package?

 - They are almost always the wrong choice, algorithmically and practically
 - The times that they are not, is generally a part of a functional pattern, and so better fit FunctionalCollections.jl
 - They are the least mature part of the package and maturing them out would further delay being able to make a 1.0 release.

PR #771, opened 2021-11-10
Deprecate enqueue!, dequeue! and peek for PriorityQueue

### LINKED ISSUE #296: Use `push!` and `pop!` for `PriorityQueue` etc

jonas-schulze:
I took a stab at the remaining part of #296. Please note my [comment on discourse](https://discourse.julialang.org/t/taking-push-and-pop-seriously/34326/20?u=jonas-schulze) on instead using `put!` and `take!`, but I think a consistent API is more important now.

If the refactor of `popfirst!` to match `eltype` went too far (*breaking change*), I can drop the corresponding commit. Also, I for now deprecated `enqueue(pq, key, val)` in favor of `Base.push!(pq, key=>val)` instead of `Base.insert!(pq, key, val)`. Less is more, I think. `Dict` doesn't have `push!` nor `insert!`.

This is ready for comments, but there are some parts missing:

- [x] Decide on refactor of `popfirst!` (Edit: it's fine)
- [x] Decide on `push!` vs `insert!` (Edit: it's ok as it is)
- [x] Update docs

Close #296

### DIFF
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -102,7 +102,7 @@ module DataStructures
     export status
     export deref_key, deref_value, deref, advance, regress
 
-    export PriorityQueue, peek
+    export PriorityQueue
 
     include("priorityqueue.jl")
     include("sparse_int_set.jl")
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -15,7 +15,7 @@ module DataStructures
     export complement, complement!
 
     export Deque, Stack, Queue, CircularDeque
-    export enqueue!, dequeue!, dequeue_pair!, update!
+    export dequeue!, dequeue_pair!, update!
     export capacity, num_blocks, top_with_handle, sizehint!
 
     export Accumulator, counter, reset!, inc!, dec!
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -15,7 +15,7 @@ module DataStructures
     export complement, complement!
 
     export Deque, Stack, Queue, CircularDeque
-    export dequeue_pair!, update!
+    export update!
     export capacity, num_blocks, top_with_handle, sizehint!
 
     export Accumulator, counter, reset!, inc!, dec!
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -5,6 +5,10 @@ module DataStructures
                 isbitsunion, isiterable, dict_with_eltype, KeySet, Callable, _tablesz,
                 findnextnot, unsafe_getindex, unsafe_setindex!, peek
 
+    if !isdefined(Base, :popat!)
+        export popat!
+    end
+
 
     using Compat # Provides Base.Order.ReverseOrdering(). May remove this line with julia 1.4
     using OrderedCollections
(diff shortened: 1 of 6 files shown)


jonas-schulze:
The reported coverage loss is in `src/splay_tree.jl`, which I didn't even touch. 👀 
https://app.codecov.io/gh/JuliaCollections/DataStructures.jl/compare/771/changes

jonas-schulze:
@oxinabox bump 🙂

src/priorityqueue.jl
```diff
@@ -306,10 +274,16 @@ function dequeue_pair!(pq::PriorityQueue)
...
 end
 
-function dequeue_pair!(pq::PriorityQueue, key)
+if isdefined(Base, :popat!)
+    Base.popat!(pq::PriorityQueue, key) = _popat!(pq, key)
+else
+    popat!(pq::PriorityQueue, key) = _popat!(pq, key)
+end
+
+function _popat!(pq::PriorityQueue, key)
```
oxinabox, line 283: What do you think of varying the normal rule against using import?
```suggestion
if isdefined(Base, :popat!)  # We will overload if it is defined, else we define on our own
    import popat!
end

function popat!(pq::PriorityQueue, key)
```
I think it might be the lesser of two evils here?

src/DataStructures.jl
```diff
@@ -5,6 +5,10 @@ module DataStructures
                 isbitsunion, isiterable, dict_with_eltype, KeySet, Callable, _tablesz,
                 findnextnot, unsafe_getindex, unsafe_setindex!, peek
 
+    if !isdefined(Base, :popat!)
+        export popat!
+    end
```
oxinabox, line 10: What do you think of:
```suggestion
    export popat!  # Export for old version of julia where Base doesn't export this
```

I think it should work since it is either exporting something new (fine)
or it is a no-op by exporting something already exported

test/test_priority_queue.jl
```diff
@@ -110,13 +110,13 @@ import Base.Order.Reverse
     @testset "PriorityQueueMethods" begin
         pq1 = PriorityQueue('a'=>1, 'b'=>2)
 
-        @testset "peek/get/dequeue!/get!" begin
+        @testset "peek/get/popfirst!/get!" begin
             @test peek(pq1) == ('a'=>1)
```
oxinabox, line 114: Probably out of scope for this PR, but this use of `peek` should be replaces with `first`.
Which will
A) Nicely match to `popfirst`
B) Match `first` on a `Dict`
C) be consistent with `Base` using `peek` only for IO streams

oxinabox (commented):
Nice work.
This looks good to me.
Just a few comments that I will wait for you to reply to before I merge

oxinabox (approved):

src/priorityqueue.jl
jonas-schulze: I like what you proposed in https://github.com/JuliaCollections/DataStructures.jl/pull/771#discussion_r751502217, but I don't have a strong preference here. Which two evils do you mean here?

oxinabox: Evil 1: is `import Foo: bar` because it allowed unqualified overloads
Evil 2: is this code is kinda complex for what it it does.

I think yes let's conditionally import it

oxinabox (approved):

oxinabox:
thanks!

PR #775, opened 2021-11-17
Fixed long-standing bug in balanced_tree.jl that corrupted SortedMult…

StephenVavasis:
…iDict

for certain insertions.

### DIFF
--- a/src/balanced_tree.jl
+++ b/src/balanced_tree.jl
@@ -354,6 +354,7 @@ end
 ## done whether the iterm
 ## is already in the tree, so insertion of a new item always succeeds.
 
+
 function Base.insert!(t::BalancedTree23{K,D,Ord}, k, d, allowdups::Bool) where {K,D,Ord <: Ordering}
 
     ## First we find the greatest data node that is <= k.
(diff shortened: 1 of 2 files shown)


oxinabox:
It will be a little while before I can review.

Are there still tests to be added?

StephenVavasis:
I can't think of any other useful tests.  It's kind of surprising that the existing test set did not catch this error.  If it makes sense, I can add the OP's example to the tests.  Let me know if I can help with the code review.

oxinabox:
> If it makes sense, I can add the OP's example to the tests.

I think it does, just as a regression test to be sure this exact error case doesn't happen again

StephenVavasis:
I added the new test, but I am unable to load my branch to test the new test because apparently I am using Pkg incorrectly.  See the error message below. I got the same error from Pkg.develop.  Sorry, I still do not have a good grasp on Pkg.

```julia
julia> Pkg.add(url="https://github.com/StephenVavasis/DataStructures.jl#fix_sortedmultidict_bug")
     Cloning git-repo `https://github.com/StephenVavasis/DataStructures.jl#fix_sortedmultidict_bug`
    Updating git-repo `https://github.com/StephenVavasis/DataStructures.jl#fix_sortedmultidict_bug`
ERROR: could not find project file in package at `https://github.com/StephenVavasis/DataStructures.jl#fix_sortedmultidict_bug` maybe `subdir` needs to be specified
```

oxinabox:
I would just use git.
git clone your branch then run it with `julia --project="."` then `] test`.

StephenVavasis:
OK, that worked, thanks!  So the modified test set is good to go.

StephenVavasis:
Any progress on this?  The bug is serious and would affect all users of SortedMultiDict.  Thanks.

StephenVavasis:
By the way, here is the official way to use Pkg to add a branch from github:

``` 
Pkg.add(PackageSpec(url="https://github.com/StephenVavasis/DataStructures.jl", rev="fix_sortedmultidict_bug"))
```
I found out how to do this here: https://github.com/JuliaLang/Pkg.jl/issues/2852

src/balanced_tree.jl
```diff
@@ -395,134 +396,122 @@ function Base.insert!(t::BalancedTree23{K,D,Ord}, k, d, allowdups::Bool) where {
...
-                                          oldtreenode.splitkey1, oldtreenode.splitkey1)
-            righttreenodenew = TreeNode{K}(newchild, oldtreenode.child3, 0,
-                                           oldtreenode.parent,
-                                           oldtreenode.splitkey2, oldtreenode.splitkey2)
-            whichp = 2
+    ## change the 2-node to a 3-node with the new child.
+
+    while true
+
+
```
oxinabox, line 426:
```suggestion
```

src/balanced_tree.jl
```diff
@@ -395,134 +396,122 @@ function Base.insert!(t::BalancedTree23{K,D,Ord}, k, d, allowdups::Bool) where {
...
-                                           oldtreenode.parent,
-                                           minkeynewchild, minkeynewchild)
-            minkeynewchild = oldtreenode.splitkey2
-            whichp = 2
+            throw(AssertionError("Tree structure is corrupted 1"))
+        end
+
+        # Two cases: either we need to split the tree node
+        # if newchild4>0 else we convert a 2-node to a 3-node
+        # if newchild4==0
```
oxinabox, line 464:
```suggestion
        # Two cases: 
        # if newchild4>0  we need to split the tree node
        # elseif newchild4==0  we convert a 2-node to a 3-node
```

src/balanced_tree.jl
```diff
@@ -647,6 +637,7 @@ function compareInd(t::BalancedTree23, i1::Int, i2::Int)
         p1 = t.tree[i1a].parent
         p2 = t.tree[i2a].parent
         @invariant_support_statement curdepth -= 1
+        curdepth -= 1
```
oxinabox, line 640:
```suggestion
```

src/balanced_tree.jl
```diff
@@ -624,6 +613,7 @@ function compareInd(t::BalancedTree23, i1::Int, i2::Int)
     p1 = t.data[i1].parent
     p2 = t.data[i2].parent
     @invariant_support_statement curdepth = t.depth
+    curdepth = t.depth
```
oxinabox, line 616:
```suggestion
```

oxinabox (approved):
merge when happy

StephenVavasis:
I don't think I have permission to merge.

oxinabox:
Huh, I was sure you were a member.
You should be.
[Here is your invite](https://github.com/JuliaCollections)
Read and follow https://github.com/SciML/ColPrac

StephenVavasis:
Thanks!  I did a squash-and-merge.  Can we also make a new release, since this is a major bug-fix?

oxinabox:
Yes, I will backport it and release

PR #777, opened 2021-12-11
Cleaner percolate up/down syntax

MarcinKonowalczyk:
I was playing around with the arrays_as_heaps code, and I think this is ever so slightly cleaner.


MarcinKonowalczyk:
yes. the coverage went down because the line count went down 🙄

oxinabox (approved):

oxinabox:
thanks.

PR #780, opened 2021-12-17
Proposed bug fix for bad state when using CircularBuffer push!

### LINKED ISSUE #754: CircularBuffer push! increases size even if value can't be converted
Reproducible with the following steps:
```
julia> using DataStructures
julia> cb = CircularBuffer{Int}(5);
julia> size(cb)
(0,)
julia> push!(cb, 1.5)
ERROR: InexactError: Int64(1.5)
[...]
julia> size(cb)
(1,)
pkg> status
[...]
  [864edb3b] DataStructures v0.18.9
[...]
```

This seems strongly like unexpected behavior to me and should probably be fixed.

robert-wright:
closes #754

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -101,14 +101,18 @@ end
 
 Add an element to the back and overwrite front if full.
 """
-@inline function Base.push!(cb::CircularBuffer{T}, data::S) where {T, S <: T}
+@inline function Base.push!(cb::CircularBuffer{T}, data) where T
+
+    # As per the behaviour of Base.push!
+    data_converted = convert(T, data)
+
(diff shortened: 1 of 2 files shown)


simonschoelly:
Maybe it would be better, if it is not required that `typeof(data) <: T`, but rather in the first line of your function you have
```juliu
data = convert(T, data)
```
then it would just throw an exception if that conversion is not possible. This would be more in line with the behavior of `push!` on `Vector`, where one can push an `Int` to a `Vector{Float64}`.

oxinabox (approved):
LGTM

robert-wright:
Thanks @oxinabox and @simonschoelly for the review 👍

Issue #781, opened 2021-12-20
Seeking advice: SortedDict constructors


StephenVavasis:
Right now I'm making major revisions to the sorted containers to address several open issues.  Here is an issue that I was not planning to address but have stumbled across.  The file https://github.com/JuliaCollections/DataStructures.jl/blob/master/src/sorted_dict.jl  starts off with a complex sequence of constructors.  I'm having trouble understanding the code that attempts to deduce types, and  in any case it doesn't work as well as Dict. For example:
```
julia> w = SortedDict((i=>i^2) for i=1:3)
SortedDict{Any, Any, Base.Order.ForwardOrdering} with 3 entries:
  1 => 1
  2 => 4
  3 => 9

julia> Dict((i=>i^2) for i=1:3)
Dict{Int64, Int64} with 3 entries:
  2 => 4
  3 => 9
  1 => 1
```

I'm wondering: would it be OK for the constructor `SortedDict(kv::Any)` to pass `kv` to `collect`, and then build the SortedDict from the resulting Vector? The function `collect`, like `Dict`, excels at deducing types.  The disadvantage is that it is inefficient to copy the data twice, but the documentation can state that the constructor `SortedDict{K,V}(kv)` is preferred over `SortedDict(kv)`.

oxinabox:
Yes, that would be OK.
We can always improve it later.

StephenVavasis:
Closed via #787

Issue #767, opened 2021-10-25
missing some generic interfaces for sorted containers


StephenVavasis:
The sorted containers do not currently support the following interface methods described in Chap. 15 of the Julia manual: reverse iteration, firstindex, lastindex.

StephenVavasis:
Can anyone tell me: should `reverse` for the sorted containers extend `Base.reverse` or `Base.Iterators.reverse`? Or does it not matter?

oxinabox:
I think: if it is eager and allocates a data structure that holds a complete copy of the data reversed `Base.reverse`.
But if it returns an object that can be iterated to get a reversed version of the data but is not a complete copy then `Base.Iterators.reverse`.
(e.g. if it just reverses keys but shares look up of original array of values or something)

bjarthur:
related?

```
julia> d=Dict("a"=>1, "b"=>2)
Dict{String, Int64} with 2 entries:
  "b" => 2
  "a" => 1

julia> filter(isequal("a"), keys(d))
Set{String} with 1 element:
  "a"

julia> ds=SortedDict(d)
SortedDict{String, Int64, Base.Order.ForwardOrdering} with 2 entries:
  "a" => 1
  "b" => 2

julia> filter(isequal("a"), keys(ds))
ERROR: MethodError: no method matching filter(::Base.Fix2{typeof(isequal), String}, ::DataStructures.SDMKeyIteration{SortedDict{String, Int64, Base.Order.ForwardOrdering}})
Closest candidates are:
  filter(::Any, ::Tuple) at ~/bin/julia-1.7.1/share/julia/base/tuple.jl:363
  filter(::Any, ::Array{T, N}) where {T, N} at ~/bin/julia-1.7.1/share/julia/base/array.jl:2479
  filter(::Any, ::AbstractDict) at ~/bin/julia-1.7.1/share/julia/base/abstractdict.jl:431
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[105]:1
```

bjarthur:
perhaps because `keys` of a Dict is a Set while that of a SortedDict is an iterator:

```
julia> keys(d)
KeySet for a Dict{String, Int64} with 2 entries. Keys:
  "b"
  "a"

julia> keys(ds)
DataStructures.SDMKeyIteration{SortedDict{String, Int64, Base.Order.ForwardOrdering}}(SortedDict("a" => 1, "b" => 2))
```

[...]

Issue #742, opened 2021-06-16
Merge for sorted containers too slow


StephenVavasis:
A discourse user reported in [this thread](https://discourse.julialang.org/t/indexing-a-document-using-dictionary/62724/11) that `merge` for the sorted containers was too slow and memory-intensive for his application.  Indeed, the current algorithm requires N*log N operations, whereas a more complicated code could achieve linear time.

StephenVavasis:
Closed via #787

Issue #726, opened 2021-02-24
Misleading doc for `insert!(sc, k)`


SimonCoste:
The following method call should be `insert!(sc, k, v)` and not `insert!(sc, k)`. 

https://github.com/JuliaCollections/DataStructures.jl/blob/4cc80fbce85336291900f01b985806e3821abcc6/src/sorted_multi_dict.jl#L152-L166

Also, this doc entry should be merged with this duplicate:

https://github.com/JuliaCollections/DataStructures.jl/blob/4cc80fbce85336291900f01b985806e3821abcc6/src/sorted_dict.jl#L256-L270

Another duplicate was pointed in https://github.com/JuliaCollections/DataStructures.jl/issues/559 for `pop!(sc, k)`.

StephenVavasis:
Closed via #787

PR #787, opened 2022-01-20
Sortedcontainerupdates

StephenVavasis:
Major update to sorted containers.

Address issue https://github.com/JuliaCollections/DataStructures.jl/issues/239: Dict constructors on typed data but without explicit type declarations should no longer return `{Any,Any}` items because collect is now used when explicit types are omitted

Address issue https://github.com/JuliaCollections/DataStructures.jl/issues/479: Deprecate `insert!`, `pop!(s)`, introduce `popfirst!`, `poplast!`, `ss_push!`, `sd_push!`, `smd_push!`

Address issue https://github.com/JuliaCollections/DataStructures.jl/issues/528: `push!` documentation improved

Address issue https://github.com/JuliaCollections/DataStructures.jl/issues/667: Docs extensively cleaned up and much greater use of documenter.jl

Address issue https://github.com/JuliaCollections/DataStructures.jl/issues/669: Expanded use of tokens including `a:b` notation for tokens

Address issue https://github.com/JuliaCollections/DataStructures.jl/issues/671: Add reverse iteration

Address issue https://github.com/JuliaCollections/DataStructures.jl/issues/742: New union, merge, symdiff, setdiff, intersect algorithms for sorted containers to improve performance

Address issue https://github.com/JuliaCollections/DataStructures.jl/issues/726: Clarify docs on `ss_push!`, `sd_push!`, and `sdm_push!` (formerly `insert!`)

Address issue https://github.com/JuliaCollections/DataStructures.jl/issues/767: Implement firstindex, lastindex, reverse for sorted container iteration

Address issue https://github.com/JuliaCollections/DataStructures.jl/issues/781: Use collect on an intermediate step
to infer types in untyped construction of sorted containers

Also, this commit cleans up the constructors and adds two new iteration-helper functions `inclusive_key` and `exclusive_key` that were useful in my own work.


StephenVavasis:
The following test from the new testset, which is meant to exercise a new `show` method, is failing on some platforms:
```julia
    my_assert(remove_spaces(repr(MIME{Symbol("text/plain")}(), factors5)) ==
              "SortedMultiDict{Char,Int64,ForwardOrdering}" *
              "(ForwardOrdering(),['a'=>1,'b'=>2,'c'=>3])")
```
Here, `remove_spaces` is a locally written function-- see below.  So apparently on some platforms, `show` works differently?  Anyone know the differences?  Or is there a better way to write a unit test for `show`?

```julia
function remove_spaces(s::String)
    b = Vector{UInt8}()
    for c in s
        if !isspace(c)
            push!(b,UInt8(c))
        end
    end
    String(b)
end
```

StephenVavasis:
The reason the test is failing is: In Julia 1.0, printing the vector `[1=>'a',2=>'b']` prepends the output with `Pair{Int64,Char}`.  Prepending this eltype-declaration does not happen in Julia 1.7.  Apparently I need to think of a more robust unit test for a `show` method.

StephenVavasis:
I replaced the failing unit test for `show` that does not require precisely a particular string, and now all tests are successful.

oxinabox (commented):
sorry I have not finished reviewing this.
This broadly speaking looks good and if all the existing tests still pass then it should be good

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+# From tokens.jl:
+# abstract type AbstractSemiToken end
+
+# struct IntSemiToken <: AbstractSemiToken
+#     address::Int
+# end
+
+const SDMContainer = Union{SortedDict, SortedMultiDict}
+const SortedContainer = Union{SDMContainer, SortedSet}
+const Token = Tuple{SortedContainer, IntSemiToken}
```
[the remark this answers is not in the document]
StephenVavasis: Are you asking why `Token` is implemented as a tuple rather than a `struct`?   This is a legacy issue.  In pre-1.0 Julia, any struct that contained an immutable object had itself to be heap allocated.  However, in pre-1.0 Julia, operations on tuples containing an immutable object didn't always allocate.  In particular, `deref((mysorteddict, semitok))` did not heap-allocate its 2-tuple argument, whereas a hypothetical `deref(Token(mysorteddict, semitok))`, where `Token` is the constructor of a hypothetical struct, would have allocated.

src/DataStructures.jl
```diff
@@ -60,6 +65,14 @@ module DataStructures
 
     export findkey
 
+    import Base.==
+    import Base.(:)
+    import Base.+
+    import Base.-
+    import Base.isequal
+    export sd_push!, ss_push!, smd_push!, poplast!
```
oxinabox, line 73: Rather than all 3 `sd_push!, ss_push!, smd_push!` is there a single name we can overload for these that refers to what it returns, rather than hungarian notation for it's first argument?
Like `push_and_get_token!` or something? (is that what they are doing?)

src/sorted_container_iteration.jl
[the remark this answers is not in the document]
oxinabox: More my surprise was that a token included both the container and the indexing object.
I would have expected the token to just be the indexing object (but that is defined as a semitoken)
is the idea that the token is enough to fully retrieve the stored object?
But semitoken is only half of what is expected?

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+
+"""
+    token_firstindex(m::SortedContainer)
+
+
+Return the token of
+the first entry of the sorted container `m`, or the past-end token
+if the container is empty.  Time: O(log *n*)
+"""  
+token_firstindex(m::SortedContainer) = (m, firstindex(m))
```
oxinabox, line 45: do we need this to be exported?
It seems like if someone needs it is is a trivial and obvious definition.

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+"""
+    Base.lastindex(m::SortedContainer)
+    endof(m::SortedContainer)
+
+Return the semitoken of the
+last entry of the sorted container `m`, or the before-start semitoken
+if the container is empty.  Time: O(log *n*)
+"""
+Base.lastindex(m::SortedContainer) = endof(m)
+endof(m::SortedContainer) = IntSemiToken(endloc(m.bt))
```
oxinabox, line 58: as per `startof` do we need this? or can we just use `lastindex`?

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+
+
+"""
+    token_lastindex(m::SortedContainer)
+
+Return the token of the
+last entry of the sorted container `m`, or the before-start semitoken
+if the container is empty.  Time: O(log *n*)
+"""
+token_lastindex(m::SortedContainer) = (m, lastindex(m))
```
oxinabox, line 68: as per `token_firstindex` do we need to export this?
It is a trivial and obvious definition

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+
+function deref(ii::SortedDictToken)
+    has_data(ii)
+    deref_nocheck(ii)
+end
+
+@inline function deref_nocheck(ii::SortedMultiDictToken)
+    @inbounds kdrec = ii[1].bt.data[ii[2].address]
+    return kdrec.k => kdrec.d
+end
```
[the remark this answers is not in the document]
oxinabox: similar for the other `foo`/`foo_nocheck` pairs?

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+k in keys(sd)
+k in keys(smd)
+```
+
+Here, `sd` is a SortedDict, `smd` is a SortedMultiDict, and `ss` is
+a SortedSet.
+
+These five invocations of `in` use the index structure of the sorted
+container and test equality based on the order object of the keys
+rather than `isequal`. Therefore, these five are all faster than
```
oxinabox, line 1057: should we comment here that if `isequal` is defined consistently with `isless` etc this should have same behavour?

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+"""
+
+Base.empty!(m::SortedContainer) = (empty!(m.bt); m)
+Base.length(m::SortedContainer) = length(m.bt.data) - length(m.bt.freedatainds) - 2
+Base.isempty(m::SortedContainer) = length(m) == 0
+
+
+(:)(t1::Token, t2::Token) = _colon(t1,t2)
+
+function _colon(t1::Token, t2::Token)
```
oxinabox, line 1120: why not define this as `:` ?

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+Here, `t` is a token for a sorted container and `j` is an integer. 
+If `j` is negative, then `+` regresses while `-` advances.
+If the operation `t+j` or `t-j` reaches the before-start 
+or past-end positions in the container,
+then the before-start/past-end tokens are returned (and there is no error).
+Time: O(*j*+log *n*), so this function is not optimized for long jumps.
+"""
++(t1::Token, numstep::Integer) = 
+    numstep >= 0 ? stepforward(t1, numstep) : stepback(t1, -numstep)
+
```
oxinabox, line 1141: this easily fits on 1 line, no?
```suggestion
+(t1::Token, numstep::Integer) = numstep >= 0 ? stepforward(t1, numstep) : stepback(t1, -numstep)

```

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+-(t1::Token, numstep::Integer) = 
+    numstep >= 0 ? stepback(t1, numstep) : stepforward(t1, -numstep)
```
oxinabox, line 1143: this easily fits on 1 line, no?
```suggestion
-(t1::Token, numstep::Integer) = numstep >= 0 ? stepback(t1, numstep) : stepforward(t1, -numstep)
```

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+
+
+function stepforward(t1::Token, numstep::Integer)
+    m = t1[1]
+    j = t1[2].address
+    !(j in m.bt.useddatacells) && throw(BoundsError())
```
oxinabox, line 1149: what do you think of
```suggestion
    (j in m.bt.useddatacells) || throw(BoundsError())
```

src/sorted_dict.jl
```diff
@@ -3,411 +3,294 @@
...
-Returns the (key,value) type (a 2-entry pair, i.e., `Pair{K,V}`) for
-SortedDict and SortedMultiDict. Returns the key type for SortedSet.
-This function may also be applied to the type itself. Time: O(1)
-"""
-@inline Base.eltype(m::SortedDict{K,D,Ord}) where {K,D,Ord <: Ordering} =  Pair{K,D}
-@inline Base.eltype(::Type{SortedDict{K,D,Ord}}) where {K,D,Ord <: Ordering} =  Pair{K,D}
-
-"""
-    in(p, sc)
+@deprecate insert!(m::SortedDict, k, d) sd_push!(m::SortedDict, k=>d)
```
oxinabox, line 230: can we move this into `src/deprecations.jl` so we remember to delete it in next release?

test/test_sorted_containers.jl
```diff
@@ -1,18 +1,24 @@
-import Base.Ordering
-import Base.Forward
-import Base.Reverse
-import DataStructures.eq
+using Base: Ordering, Forward, Reverse, ForwardOrdering,
+            ReverseOrdering
```
oxinabox, line 2: this fits on one line
```suggestion
using Base: Ordering, Forward, Reverse, ForwardOrdering, ReverseOrdering
```

test/test_sorted_containers.jl
```diff
@@ -1,18 +1,24 @@
...
+using DataStructures: IntSemiToken
 import Base.lt
-import Base.ForwardOrdering
-import Base.ReverseOrdering
-import DataStructures.IntSemiToken
+import DataStructures.eq
```
oxinabox, line 5: We should make the imports into `using`s i think

test/test_sorted_containers.jl
```diff
@@ -1,18 +1,24 @@
-import Base.Ordering
-import Base.Forward
-import Base.Reverse
-import DataStructures.eq
+using Base: Ordering, Forward, Reverse, ForwardOrdering,
+            ReverseOrdering
+using DataStructures: IntSemiToken
 import Base.lt
```
oxinabox, line 4: isn't this called `Base.lessthan` now days?

test/test_sorted_containers.jl
```diff
@@ -1,18 +1,24 @@
...
+import DataStructures.eq
+
 
 struct CaseInsensitive <: Ordering
 end
 
 lt(::CaseInsensitive, a, b) = isless(lowercase(a), lowercase(b))
 eq(::CaseInsensitive, a, b) = isequal(lowercase(a), lowercase(b))
 
+
```
oxinabox, line 14:
```suggestion
# for testing custom ordering
```

test/test_sorted_containers.jl
```diff
@@ -32,6 +38,17 @@ function my_primes(N)
...
+function remove_spaces(s::String)
+    b = Vector{UInt8}()
+    for c in s
+        if !isspace(c)
+            push!(b,UInt8(c))
+        end
+    end
+    String(b)
+end
+
```
oxinabox, line 50:
```suggestion
remove_spaces(s::String) = replace(s, re"\s+"=>"")    
```

test/test_sorted_containers.jl
```diff
@@ -1765,3 +2093,4 @@ end
         end
     end
 end
+
```
oxinabox, line 2096: can we label the new tests with a comment including the URL of the github issues that they prove is solved?

oxinabox (commented):
Nice work.
I assume that the logic is at least as correct as before as all previous tests still pass.

Some mostly stylistic comments to resolve

src/DataStructures.jl
```diff
@@ -60,6 +65,14 @@ module DataStructures
 
     export findkey
 
+    import Base.==
+    import Base.(:)
+    import Base.+
+    import Base.-
+    import Base.isequal
```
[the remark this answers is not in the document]
StephenVavasis: I am having trouble with this for `==` and `:`.  In particular, the following constructs give me syntax errors:
```
Base.(==)(a,b) = ...
Base.==(a,b) = ...
function Base.(:)(a,b)
```
whereas there are no syntax errors with
```
import Base.==
import Base.(:)

==(a,b) = ...
function (:)(a,b)
```
Any thoughts how to fix this?

src/sorted_container_iteration.jl
StephenVavasis: The issue here is not whether `isless` and `isequal` are consistent.  Rather, the issue is whether the user has defined an ordering (such as case-insensitive for strings) for the container different from `isequal` and `isless`.

src/sorted_container_iteration.jl
StephenVavasis: The definition `function Base.(:)(a,b)` gives a syntax error (noted above).  However, `function (:)(a,b)` works, so I have fixed this (but I still have the issue with `import` noted above).

src/sorted_container_iteration.jl
StephenVavasis: I can make this change, but I have used the idiom `!(correct) && throw(...)` (rather than `correct || throw(...)`) throughout my code.  Do you recommend that I change them all.

test/test_sorted_containers.jl
StephenVavasis: I checked `ordering.jl` in `Base`, and I still see `lt`.  Where did you see `lessthan`?

src/sorted_container_iteration.jl
StephenVavasis: OK, I figured this out.

src/DataStructures.jl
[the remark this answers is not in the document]
StephenVavasis: I figured this out.

src/sorted_container_iteration.jl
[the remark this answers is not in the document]
StephenVavasis: I tried to implement this, but it is not working properly, so perhaps I am misunderstanding the rules.  Here is the new definition of `deref`:

```julia
Base.@propagate_inbounds function deref(ii::SortedDictToken)
    Base.@boundscheck has_data(ii)
    @inbounds kdrec = ii[1].bt.data[ii[2].address]
    return Pair(kdrec.k, kdrec.d)
end
```

My intention is that if a `deref` operation is prefixed by `@inbounds`, then `has_data` will not be invoked.  But this is not happening:

```julia
julia> s = SortedDict{Int,String}()
SortedDict{Int64, String, Base.Order.ForwardOrdering}()

julia> deref(token_firstindex(s))
ERROR: BoundsError
Stacktrace:
 [1] has_data
   @ c:\Users\vavasis\OneDrive - University of Waterloo\Julia_pkg_devdir\DataStructures.jl\src\sorted_container_iteration.jl:442 [inlined]
 [2] deref(ii::Tuple{SortedDict{Int64, String, Base.Order.ForwardOrdering}, DataStructures.Tokens.IntSemiToken})
   @ DataStructures c:\Users\vavasis\OneDrive - University of Waterloo\Julia_pkg_devdir\DataStructures.jl\src\sorted_container_iteration.jl:242
 [3] top-level scope
   @ REPL[24]:1

julia> @inbounds deref(token_firstindex(s))
ERROR: BoundsError
Stacktrace:
 [1] has_data
   @ c:\Users\vavasis\OneDrive - University of Waterloo\Julia_pkg_devdir\DataStructures.jl\src\sorted_container_iteration.jl:442 [inlined]
 [2] deref(ii::Tuple{SortedDict{Int64, String, Base.Order.ForwardOrdering}, DataStructures.Tokens.IntSemiToken})
   @ DataStructures c:\Users\vavasis\OneDrive - University of Waterloo\Julia_pkg_devdir\DataStructures.jl\src\sorted_container_iteration.jl:242
 [3] top-level scope
   @ REPL[25]:1
```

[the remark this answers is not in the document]
StephenVavasis: OK, nvm, I figured this out too.  Apparently `@inbounds` is not necessarily active if it is called directly from the REPL as opposed to its being called from a compiled function.

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+"""  
+Base.firstindex(m::SortedContainer) = startof(m)
+startof(m::SortedContainer) = IntSemiToken(beginloc(m.bt))
+
+
+"""
+    token_firstindex(m::SortedContainer)
+
+
+Return the token of
```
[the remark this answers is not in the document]
StephenVavasis: I would like to leave these.  See other comment.

src/sorted_container_iteration.jl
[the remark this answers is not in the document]
StephenVavasis: I think full tokens are useful, e.g., `deref(token_firstindex(s))` as opposed to `deref((s,firstindex(s)))` where the user needs to be consistent with the argument `s`.  So if it is OK with you, I would like to leave these.

src/sorted_container_iteration.jl
```diff
@@ -0,0 +1,1167 @@
...
+
+
+"""
+    Base.searchsortedfirst(m::SortedContainer, k)
+
+Return the semitoken of the first item in the 
+sorted container `m` that is greater than or equal to
+`k` in the sort order.
+If there is no
+such item, then the past-end semitoken is returned.  Time: O(*c* log *n*)
```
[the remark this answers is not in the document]
StephenVavasis: I didn't know this!  However, if you don't mind, I'll leave it as it is because there is not much benefit in changing it compared to the effort needed.

docs/Project.toml
```diff
@@ -1,5 +1,4 @@
 [deps]
+Coverage = "a2441757-f6aa-5fb2-8edb-039e3f45d037"
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-
-[compat]
-Documenter = "0.23"
```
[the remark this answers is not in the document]
StephenVavasis: OK, I put it back.  I'm not sure how it got deleted.

docs/src/sorted_containers.md
```diff
@@ -10,18 +10,19 @@ with the additional feature that the keys are stored in sorted order and
...
+Internally, SortedSet is implemented as a
+SortedDict in which the value type is `Nothing`. Finally, SortedMultiDict
 is similar to SortedDict except that each key can be associated with
 multiple values. The key=>value pairs in a SortedMultiDict are stored
 according to the sorted order for keys, and key=>value pairs with the
 same key are stored in order of insertion.
 
 The containers internally use a 2-3 tree, which is a kind of balanced
-tree and is described in many elementary data structure textbooks.
+tree and is described in data structure textbooks.
```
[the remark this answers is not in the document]
StephenVavasis: done

docs/src/sorted_containers.md
```diff
@@ -85,616 +61,378 @@ token can be regressed. A dereferencing operation on either leads to an
...
-have a documented interpretation in terms of the container.
+integers. Users should regard these integers as opaque
+since future versions of the package may change the internal indexing
+scheme.
+In certain situations it may be more costly to operate on tokens than
+semitokens because the first entry of a token (i.e., the container)
+is not a bits-type.
+If code profiling indicates that statements using tokens are allocating memory,
+then it may be advisable to rewrite the application code using semitokens
+more than tokens.
```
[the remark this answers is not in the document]
StephenVavasis: done

docs/src/sorted_containers.md
```diff
@@ -85,616 +61,378 @@ token can be regressed. A dereferencing operation on either leads to an
...
-returns the before-start semitoken. It is an error to invoke this
-function if `(sc,st)` is the before-start token. If `(sc,st)` is the
-past-end token, then this routine returns the smitoken of the last
-item in the sort order (i.e., the same semitoken returned by the
-`endof` function). Time: O(log _n_)
-
-    searchsortedfirst(sc,k)
+```@docs
+Base.first(sc::SortedContainer)
+```
```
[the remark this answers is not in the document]
StephenVavasis: I didn't know about the other syntax for documenter, but now that you told me, I have fixed it.

docs/src/sorted_containers.md
```diff
@@ -85,616 +61,378 @@ token can be regressed. A dereferencing operation on either leads to an
...
 
 ```@docs
-pop!(ss::SortedSet, k)
+sd_push!(sd::SortedDict, p::Pair)
 ```
 
 ```@docs
-pop!(ss::SortedSet)
+smd_push!(smd::SortedMultiDict, p::Pair)
 ```
```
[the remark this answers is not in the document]
StephenVavasis: Done

src/DataStructures.jl
StephenVavasis: Done.  The new function is called `push_return_token!`.  Since it is a single function name, I unified the APIs by making the return variables `Bool,IntSemiToken`. The `Bool` return variable is redundant for SortedMultiDict because push always creates a new record for SortedMultiDict (i.e., the first return variable is always `true` for a SortedMultiDict).

src/sorted_container_iteration.jl
StephenVavasis: I have deprecated `startof` and `endof` in favor of `firstindex` and `lastindex`.

src/sorted_container_iteration.jl
StephenVavasis: See other comment about `token_firstindex`.

src/sorted_container_iteration.jl
StephenVavasis: See other comment about `token_firstindex`

src/sorted_container_iteration.jl
StephenVavasis: Done.

src/sorted_container_iteration.jl
StephenVavasis: Done.

src/sorted_dict.jl
StephenVavasis: Done

test/test_sorted_containers.jl
StephenVavasis: Done

test/test_sorted_containers.jl
StephenVavasis: Done

test/test_sorted_containers.jl
StephenVavasis: I found out that it is actually `r"\s+"` rather than `re"\s+"`, but aside from that, this is done.

StephenVavasis:
@oxinabox I believe I have addressed all of your comments in this PR.  In the last commit that I just made, I have added comments to the test sets to indicate where the various issues are tested.  Let me know if you need anything else before you can merge this PR.

src/DataStructures.jl
```diff
@@ -39,12 +42,14 @@ module DataStructures
     export SortedDict, SortedMultiDict, SortedSet
     export SDToken, SDSemiToken, SMDToken, SMDSemiToken
     export SetToken, SetSemiToken
-    export startof
+    export startof, endof ## both are deprecated
```
oxinabox, line 45: We don't need to explictly export deprecated things `@deprecate` does that for us

```suggestion
```

oxinabox (approved):
Sounds good.
You have merge rights.
Merge when you are happy

PR #802, opened 2022-04-15
Add missing argument to insert! docstring

schillic:


StephenVavasis:
Thanks for noticing this, but it is preempted by #787 so I am closing it.

PR #803, opened 2022-04-16
Fix docs: switch "disjoint set" to "disjoint sets" in docs

### LINKED ISSUE #740: Disjoint sets docs are not up to date
They talk about `DisjointSet` instead of `DisjointSets`.

tmthyln:
This is a docs-only PR to change occurrences of "disjoint set" to "disjoint sets" in the docs. Should close #740.

This also happens to fix a slight bug that the table of contents didn't show the page for Disjoint Sets, since the name of link was changed (but the name of the actual doc file was not yet changed).


oxinabox (approved):
thanks

PR #806, opened 2022-04-23
Include URLs for dev and stable docs in README

mroavi:
- Replaces https://juliacollections.github.io/DataStructures.jl/latest (which actually points to the dev docs) with a link to the stable docs version.
- Adds badges for stable and dev docs.


oxinabox (approved):
Thanks

PR #807, opened 2022-04-23
Reuse handle remapping slots in MutableBinaryHeap

### LINKED ISSUE #745: `node_map` in MutableBinaryHeap grows indefinitely
If you push/pop an element in a loop 1 million times, `nodes` will be empty, but `node_map` will have length 1 million. It is never emptied. I think there could be better design using a circular buffer that would only ever require `node_map` to be as long as the maximum size of the heap?

ancapdev:
Fixes #745


ancapdev:
Considered changing handle tables to `Int32`, which if we're comfortable limiting the collection to 2 billion entries could save a lot of space (node map, node map free list, and node back handle). Could be done separately or here, happy to contribute either. Thoughts?

ancapdev:
Hi, any chance for a review and merge of this, please? Thanks

PR #800, opened 2022-04-13
Improved Deque Documentation + docstrings

lfenzo:
Hello!

This is the first of a series of PRs aiming to dramatically improve the current documentation of DataStructures.jl.

Some of the changes include the addition of actual documentation generated by Documenter.jl from docstrings (see `docs/src/deque.md`), a brief explanation of the Deque data structure and a few minor changes in source code.

The new documentation page formats will be structured in the following manner:

1. Data structure description and characteristics.
2. Constructors for the implemented data structure
3. Usage, showing first a list of available methods for the given data structure sorted alphabetically, followed by the Documenter.jl generated documentation.

Let me know in case of any suggestions.

PR #809, opened 2022-05-07
Improved Queue and Stack Documentation + docstrings

lfenzo:
Following #800. Note that this PR splits Queue and Stack documentation pages into two separate pages.


src/stack.jl
```diff
@@ -31,29 +24,69 @@ end
...
+Check if stack `s` is empty.
+"""
 Base.isempty(s::Stack) = isempty(s.store)
+
+
+"""
+    isempty(s::Stack)
+
+Return the number of elements in stack `s`.
+"""
```
oxinabox, line 39:
```suggestion
```

Per method dostrings are not needed when they add no value over the per function ones.
Plus this one is wrong

src/stack.jl
```diff
@@ -31,29 +24,69 @@ end
...
 Base.length(s::Stack) = length(s.store)
-Base.eltype(::Type{Stack{T}}) where T = T
+
+
+"""
+    eltype(::Type{Stack{T}}) where {T}
+
+Return the type of the elements in the stack.
+"""
```
oxinabox, line 47:
```suggestion
```

Per method dostrings are not needed when they add no value over the per function ones.
Plus this one is wrong

src/stack.jl
```diff
@@ -31,29 +24,69 @@ end
...
+
+Remove and return the top element from stack `s`.
+"""
 Base.pop!(s::Stack) = pop!(s.store)
 
+"""
+    empty!(s::Stack)
+
+Remove all elements from stack `s`.
+"""
```
oxinabox, line 80:
```suggestion
```

Per method dostrings are not needed when they add no value over the per function ones.

src/stack.jl
```diff
@@ -31,29 +24,69 @@ end
...
 
 Base.iterate(st::Stack, s...) = iterate(Iterators.reverse(st.store), s...)
 
 Iterators.reverse(s::Stack{T}) where {T} = DequeIterator{T}(s.store)
 
+"""
+    ==(x::Stack, y::Stack)
+
+Check if stacks `x` and `y` are equal in term of their contents.
+"""
```
oxinabox, line 91:
```suggestion
```

Per method dostrings are not needed when they add no value over the per function ones.

lfenzo:
Sorry for taking so long..

Thanks for the comments! I see the merit of keeping the docstrings succinct (as it seems to be the trend around most of Julia packages), but in the case I've decided to be a little more verbose and add docstring examples for each of the exported functions (except for the iterators which have their own example at the beginning of the `stack.md` doc file). Ideally, for 1.0, I believe we should have at least one example for each function regardless of how trivial they may look.

Let me know what you think about these examples! So far I only added them to the `Stack` type, with your feedback I can make the requested changes and do the same for the `Queue`.

oxinabox (approved):

oxinabox:
I am neural on most of the examples. 
Low key against them for ones that I don't think we should have method docstrings for in the first place (because of the method already being well described by the overall function documentation).
But i do see the utility of showing examples for `first` and `last` for `Queue` and `Stack`
I do think if we want to keep them it is possibly worth hiding them behind `## Extended Help`

> For long docstrings, consider splitting the documentation with an # Extended help header. The typical help-mode will show only the material above the header; you can access the full help by adding a '?' at the beginning of the expression (i.e., "??foo" rather than "?foo").

(https://docs.julialang.org/en/v1/manual/documentation/)


All in all this is looking really nice. Clear improvement over status quo

lfenzo:
I've trimmed most of the code examples leaving only `first`, `last` (as you suggested) and I also took liberty of leaving the examples of `==`; it occurred to me that the previous example and docstrings for that method were not precise enough. Besides that, I also made a couple of minor changes in the docstrings in general of `Stack`, see how that reads to you.

lfenzo:
ping @oxinabox

oxinabox (approved):
Thanks. LGTM

PR #816, opened 2022-06-10
new method implementations for Stack

baumgold:


oxinabox:
Why are these wanted?
Stacks are only supposed to be accessed from the top.

PR #818, opened 2022-07-12
add MutableLinkedList functionality

tmcgrath325:
Related to #443

In addition to the methods that have already been implemented for MutableLinkedList, it might also be helpful to have ways to mutate "the middle" of a list rather than only make changes at the beginning and end.

To that end, I've added `popat!`, `insert!`, and `splice!`.


skygering:
I would love to bump this! I need this functionality as well!

PR #824, opened 2022-08-29
Make binary heap tests slightly stricter

LilithHafner:


oxinabox (approved):

oxinabox:
thanks, these are indeed better

PR #826, opened 2022-09-08
remove unbound type parameters

nsajko:
I didn't check, but unbound type parameters often cause performance issues, so this isn't merely cosmetic.

FTR, there's a relevant function in Test in stdlib for checking this stuff:
https://docs.julialang.org/en/v1/stdlib/Test/#Test.detect_unbound_args


StephenVavasis:
Yes, it makes sense to remove these unused parameters.  Thanks for catching this.

oxinabox (approved):

PR #832, opened 2022-10-21
remove resizing of internal fields in int set tests

KristofferC:
This apparently breaks the tests on 1.9. No idea why they are there in the first place.


nickrobinson251 (approved):

KristofferC:
@oxinabox, can a new release be made with this maybe? Makes this package green on PkgEval.

PR #837, opened 2022-12-11
ready Accumulator for v1.0

oxinabox:
checks off accumulator.jl on https://github.com/JuliaCollections/DataStructures.jl/issues/479

One thing I am not sure about is the names `nlargest` and `nsmallest`
those could be changed

### DIFF
--- a/src/accumulator.jl
+++ b/src/accumulator.jl
@@ -33,9 +33,8 @@ function counter(seq)
 end
 
 eltype_for_accumulator(seq::T) where T = eltype(T)
-function eltype_for_accumulator(seq::Base.Generator)
-    Base.@default_eltype(seq)
-end
+eltype_for_accumulator(seq::Base.Generator) = Base.@default_eltype(seq)
+
 
 
 Base.copy(ct::Accumulator) = Accumulator(copy(ct.map))
@@ -44,9 +43,9 @@ Base.length(a::Accumulator) = length(a.map)
 
 ## retrieval
 
-Base.get(ct::Accumulator, x, default) = get(ct.map, x, default)
 # need to allow user specified default in order to
 # correctly implement "informal" AbstractDict interface
+Base.get(ct::Accumulator, x, default) = get(ct.map, x, default)
 
 Base.getindex(ct::Accumulator{T,V}, x) where {T,V} = get(ct.map, x, zero(V))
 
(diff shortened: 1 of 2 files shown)


nickrobinson251 (approved):

src/accumulator.jl
```diff
@@ -76,12 +75,11 @@ inc!(ct::Accumulator, x, v::Number) = (ct[x] += v)
 inc!(ct::Accumulator{T, V}, x) where {T, V} = inc!(ct, x, one(V))
 
 # inc! is preferred over push!, but we need to provide push! for the Bag interpreation
-# which is used by classified_collections.jl
 Base.push!(ct::Accumulator, x) = inc!(ct, x)
 Base.push!(ct::Accumulator, x, a::Number) = inc!(ct, x, a)
 
 # To remove ambiguities related to Accumulator now being a subtype of AbstractDict
-Base.push!(ct::Accumulator, x::Pair)  = inc!(ct, x)
+Base.push!(ct::Accumulator{P}, x::P) where P<:Pair  = inc!(ct, x)
```
oxinabox, line 82:
```suggestion
Base.push!(ct::Accumulator, x::Pair)  = inc!(ct, x)
```

src/accumulator.jl
```diff
@@ -92,7 +90,9 @@ Decrements the count for `x` by `v` (defaulting to one)
 dec!(ct::Accumulator, x, v::Number) = (ct[x] -= v)
 dec!(ct::Accumulator{T,V}, x) where {T,V} = dec!(ct, x, one(V))
 
-#TODO: once we are done deprecating `pop!` for `reset!` then add `pop!` as an alias for `dec!`
```
Krastanov, line 95: I assume it was decided at some point that `pop!` is not being deprecated so this TODO is incorrect and that is why it is deleted and disobeyed?

Krastanov: When `pop!` becomes an alias for `dec!` as seen here, then this will be happening:

```
julia> a = Accumulator(:k=>1); dec!(a, :k); dec!(a, :k)
-1
```

Just making sure that this is the desired behavior. Asking because I would have expected `KeyError` when the count goes negative.

src/accumulator.jl
```diff
@@ -92,7 +90,9 @@ Decrements the count for `x` by `v` (defaulting to one)
...
+Base.pop!(ct::Accumulator, x) = dec!(ct, x)
+Base.pop!(ct::Accumulator, x, default) = haskey(ct, x) ? dec!(ct, x) : default
```
Krastanov, line 94: similarly to what I mentioned above, here I would have expected checking for "value equal zero", not only "has key"

Krastanov (approved):
Hi, folks,

given that I depend on this package, I tried to give a bit back by reviewing. Please feel free to disregard if the community standards permit only reviews from vetted individuals.

The changes look good to me. I had some minor concerns about `pop!`, which behaves a bit surprisingly to me, but it seems self-consistent. Maybe `Counter` and `Accumulator` have different behavior for their `pop!` in which case everything is fine.

LGTM, but please let me know how I should understand `pop!`

Krastanov:
In https://github.com/JuliaCollections/DataStructures.jl/issues/479#issuecomment-682449805  there is an admonition to not use `push!(container, key, value)`, but here it says `# inc! is preferred over push!, but we need to provide push! for the Bag interpreation`. Just putting it out here in case this contradiction is important.

src/accumulator.jl
oscardssmith: IMO `KeyError` is also correct.

PR #844, opened 2022-12-30
get rid of Compat dependency and bump Julia compat to 1.6

KristofferC:


andyferris:
Bump.

(I think this makes a lot of sense)

laborg:
This PR could also remove the export of  `peek`, `popat!`, `sizehint!`, `filter`, `reverse`, `cat`. Right? All of those only add methods to Base.[function]

This part in `priorityqueue.jl`
```
if isdefined(Base, :popat!)  # We will overload if it is defined, else we define on our own
    import Base: popat!
end

function popat!(pq::PriorityQueue, key)
```
will need to change to 
```
function Base.popat!(pq::PriorityQueue, key)
```

laborg:
The changes have been incorporated into master.

PR #858, opened 2023-06-10
fix: check key values, not `get` return for haskey(k, orderedrobindict)

### LINKED ISSUE #857: Base.in(x, KeySet{OrderedRobinDict}) should not use `get`
MWE:

```julia
## Incorrect answer if value is a negative Int
julia> d = OrderedRobinDict(:a => -3)
OrderedRobinDict{Symbol, Int64} with 1 entry:
  :a => -3

julia> :a in keys(d1)
false

# Completely errors out if value is not comparable against Int with >=

julia> d = OrderedRobinDict(1 => [1,2,3], 2 => [4,5,6])
OrderedRobinDict{Int64, Vector{Int64}} with 2 entries:
  1 => [1, 2, 3]
  2 => [4, 5, 6]

julia> 1 in keys(d)
ERROR: MethodError: no method matching isless(::Int64, ::Vector{Int64})

Closest candidates are:
  isless(::Real, ::ColorTypes.AbstractGray)
   @ ColorTypes C:\Users\mrufsvold\.julia\packages\ColorTypes\1dGw6\src\operations.jl:39
  isless(::AbstractVector, ::AbstractVector)
   @ Base abstractarray.jl:2911
  isless(::Real, ::AbstractFloat)
   @ Base operators.jl:178
  ...

Stacktrace:
 [1] <(x::Int64, y::Vector{Int64})
   @ Base .\operators.jl:343
 [2] <=(x::Int64, y::Vector{Int64})
   @ Base .\operators.jl:392
 [3] >=(x::Vector{Int64}, y::Int64)
   @ Base .\operators.jl:416
 [4] in(key::Int64, v::Base.KeySet{Int64, OrderedRobinDict{Int64, Vector{Int64}}})
   @ DataStructures C:\Users\mrufsvold\.julia\packages\DataStructures\59MD0\src\ordered_robin_dict.jl:306
 [5] top-level scope
   @ REPL[74]:1
```

Base.in is using `get` and checks if the value is greater than zero, causing all these problems.

mrufsvold:
Resolves #857 

I am not sure if checking if the key appears in the key vector property is the right way to go. Open to another solution!

### DIFF
--- a/src/ordered_robin_dict.jl
+++ b/src/ordered_robin_dict.jl
@@ -305,8 +305,8 @@ julia> haskey(D, 'c')
 false
 ```
 """
-Base.haskey(h::OrderedRobinDict, key) = key in h.keys
-Base.in(key, v::Base.KeySet{K,T}) where {K,T<:OrderedRobinDict{K}} = key in v.dict.keys
+Base.haskey(h::OrderedRobinDict, key) = (get(h.dict, key, -1) > 0)
+Base.in(key, v::Base.KeySet{K,T}) where {K,T<:OrderedRobinDict{K}} = (get(v.dict.dict, key, -1) >= 0)
 
 """(get(h.dict, key, -2) > 0)
     getkey(collection, key, default)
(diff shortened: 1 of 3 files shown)


mrufsvold:
I discovered the same problem in `setindex!` last night, so I'll push a couple more fixes tomorrow.

oscardssmith (approved):

Issue #859, opened 2023-07-10
remove the `<:Number` constraint in `Accumulator`


Krastanov:
Not all number-like types in the julia ecosystem subtype number, in particular various symbolic types in AbstractAlgebra and Nemo. I happen to need to accumulate over that type of "number". May we drop the type constraint?

PR #860, opened 2023-07-10
Permit non-number types in Accumulator

Krastanov:
See #859


oxinabox:
This usecase is wild, but ok

PR #870, opened 2023-09-01
Update ordered_containers.md

SyxP:
1) Fixed incorrect example code in `OrderedSet` (Note that this code doesn't run as the various irrational numbers are not defined.)

2) Added clarification on equality. The behaviour was raised as a point of confusion in a Slack thread. I just documented the current existing behaviour.

### DIFF
--- a/docs/src/ordered_containers.md
+++ b/docs/src/ordered_containers.md
@@ -12,7 +12,7 @@ for c in 'a':'e'
 end
 collect(d) # => [('a',1),('b',2),('c',3),('d',4),('e',5)]
 
-s = OrderedSet(π,e,γ,catalan,φ)
+s = OrderedSet([π,e,γ,catalan,φ])
 collect(s) # => [π = 3.1415926535897...,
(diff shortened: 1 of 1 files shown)


jariji:
This is documenting rather than fixing bad behavior IMO. 

For any type `T`, `x::T` and `y::T` should not have `x == y` unless `x` and `y` agree on all aspects of their public API. Since order is part of the public API of `Ordered` containers, they should not be `==(::T, ::T)` unless they agree on order.

SyxP:
I agree, I don't like how this works. But as we had a discussion on Slack, this is actually how someone might prefer the behaviour to be. 

I make no judgement on whether this is the "correct" behaviour. But as of now this is what is inside the library, and I document it to prevent confusion. I will close the PR if there is another successfully merged PR augmenting the behaviour.

StephenVavasis:
I don't think ordered_set is part of this repository.  I believe that it is in the JuliaCollections/OrderedCollections.jl repository.

Issue #739, opened 2021-06-09
MutableLinkedList append! iterating infinitely


Sinfaen:
When I run the following segment of code:
```
julia> using DataStructures
julia> a = MutableLinkedList{Int}(1,2,3)
MutableLinkedList{Int64}(1, 2, 3)
julia> b = MutableLinkedList{Int}(3,4,5)
MutableLinkedList{Int64}(3, 4, 5)
julia> append!(a, b)
```
This seemingly infinite object gets printed to my screen:
```
MutableLinkedList{Int64}(1, 2, 3, 3, 4, 5, 0, 3, 4, 5, 0, 3, 4, 5, 0, 3, 4, 5, ... # the 0,3,4,5 repeats endlessly until I hit Ctrl-C
```

`length(a)` gives me the correct size of 6 though. I also tried running `collect(a)` and that seg-faulted my Julia REPL. Am I just stupid and missed something completely obvious here?

I'm running on macOS with `DataStructures` v0.18.9.

Sinfaen:
I tried fixing this issue on my windows machine by changing this function:
```
function Base.append!(l1::MutableLinkedList{T}, l2::MutableLinkedList{T}) where T
    l1.node.prev.next = l2.node.next # l1's last's next is now l2's first
    l2.node.prev.next = l1.node # l2's last's next is now l1.node
    l1.len += length(l2)
    return l1
end
```

to this:
```
function Base.append!(l1::MutableLinkedList{T}, l2::MutableLinkedList{T}) where T
    l1.node.prev.next = l2.node.next # l1's last's next is now l2's first
    l2.node.prev.next = l1.node  # l2's last's next is now l1.node
    l2.node.next.prev = l1.node.prev # l2's first's prev is now l1's last
    l1.node.prev      = l2.node.prev # l1's first's prev is now l2's last
    l1.len += length(l2)
    # make l2 empty
    l2.node.prev = l2.node
    l2.node.next = l2.node
    l2.len = 0
    return l1
end
```

It boils down to updating the `prev` references in the `l1` elements and making  `l2` empty to make it clear that `l1` took the elements (and also because otherwise printing the var passed as `l2` afterwards did the endless print to the screen).

Does this seem like the right sort of fix? Here's an example on windows after the change:
```julia
julia> using DataStructures

julia> a = MutableLinkedList{Int}(1,2,3)
MutableLinkedList{Int64}(1, 2, 3)

[...]

PR #871, opened 2023-09-01
Fix for MutableLinkedList append!

Sinfaen:
See https://github.com/JuliaCollections/DataStructures.jl/issues/739 for example of the segfault that I'm seeing

- Updates `Base.append!(l1::MutableLinkedList{T}, l2::MutableLinkedList{T}) where T` so that the resulting object doesn't print endlessly to the screen, and that `collect` doesn't segfault the julia session
   - adds more lines to update node `next` and `prev` pointers
- Adds a test in the append section so that the second argument is checked for being empty

with these changes, I'm seeing this work without segfaulting
```julia
(@v1.9) pkg> activate .
  Activating project at `~/repos/DataStructures.jl`

julia> using DataStructures

julia> l1, l2 = MutableLinkedList{Int}(1,2), MutableLinkedList{Int}(3,4)
(MutableLinkedList{Int64}(1, 2), MutableLinkedList{Int64}(3, 4))

julia> append!(l1, l2)
MutableLinkedList{Int64}(1, 2, 3, 4)

julia> collect(l1)
4-element Vector{Int64}:
 1
 2
 3
 4

julia> l2
MutableLinkedList{Int64}()
```

### DIFF
--- a/src/mutable_list.jl
+++ b/src/mutable_list.jl
@@ -153,7 +153,13 @@ end
 function Base.append!(l1::MutableLinkedList{T}, l2::MutableLinkedList{T}) where T
     l1.node.prev.next = l2.node.next # l1's last's next is now l2's first
     l2.node.prev.next = l1.node # l2's last's next is now l1.node
+    l2.node.next.prev = l1.node.prev # l2's first's prev is now l1's last
+    l1.node.prev      = l2.node.prev # l1's first's prev is now l2's last
     l1.len += length(l2)
+    # make l2 empty
+    l2.node.prev = l2.node
+    l2.node.next = l2.node
+    l2.len = 0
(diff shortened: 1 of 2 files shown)


laborg (approved):
Fixes the problematic behaviour. For 1.0 it should be decided if append! should mutate the second list...

oxinabox:
It should not, since we want to match the behavour of appending two `Vector`s

Sinfaen:
I'll change it to do a copy of the second list

oxinabox:
It's unfortunate since that's very expensive with a linked list.
Its too bad these are double-y linked lists...

Linked lists are such a can of worms

oxinabox (approved):

Issue #795, opened 2022-03-24
Ill logic of MutableLinkedList.node 


algorithmx:
https://github.com/JuliaCollections/DataStructures.jl/blob/f90dd8ceb1808123136d96599fbf41407ef4fadc/src/mutable_list.jl#L19

Code
------
````
A = MutableLinkedList{Int}(1,2,3)
node = A.node
for i=1:8
    println(node.data)
    node = node.next
end
````

Result
------
````
140155178173488
1
2
3
140155178173488
1
2
3
````

Version
------

````
julia> versioninfo()
Julia Version 1.7.2
Commit bf53498635 (2022-02-06 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 

julia> 

````

algorithmx:
related issue 
https://github.com/JuliaCollections/DataStructures.jl/issues/794

Issue #794, opened 2022-03-24
leakage in append! functions for two `MutableLinkedList`'s


algorithmx:
https://github.com/JuliaCollections/DataStructures.jl/blob/f90dd8ceb1808123136d96599fbf41407ef4fadc/src/mutable_list.jl#L153

Code
------
````
A = MutableLinkedList{Int}(1,2,3)
B = MutableLinkedList{Int}(1,2,3)
append!(A,B) # endless loop
````

Version
------

````
julia> versioninfo()
Julia Version 1.7.2
Commit bf53498635 (2022-02-06 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 

julia> 

````

algorithmx:
related issue 
https://github.com/JuliaCollections/DataStructures.jl/issues/795

algorithmx:
The local files on my computer is different from the files on GitHub, but the version number is the same ...

algorithmx:
I have examined the local files, I found that the fix wasn't merged into the master branch correctly. Can anyone check the status of the repo?
I have installed v0.18.10 first, this gives me a folder `nBjdy` in the package folder. Then I update to v0.18.11, got one more foledr with name `vSp4s`. The file `mutable_list.jl` in both versions are different from the file on GitHub, leaving the bug unfixed ...

algorithmx:
@grahamas

algorithmx:
@oxinabox

Sinfaen:
this is also what I encountered in https://github.com/JuliaCollections/DataStructures.jl/issues/739

Issue #704, opened 2020-11-18
Bug: delete!(::MutableLinkedList, ::UnitRange)


poldavezac:
If a `UnitRange` with `first == last` is used then a boundary error is thrown. See https://github.com/JuliaCollections/DataStructures.jl/blob/master/src/mutable_list.jl#L182

I can't think of a reason for forbidding such situations thus I would suggest lowering the constraints to:

`@boundscheck 0 < first(r) <= last(r) <= l.len || throw(BoundsError(l, r))`

PR #873, opened 2023-10-13
Improve MutableLinkedList for 1.0

laborg:
It started as a simple inspection into MutableLinkedList and ended into a thorough refactoring of the code. As a consequence bugs have been fixed, and the API is now in alignment with Base.

 - refactored: reduce code duplication by introducin a couple of functions basic (`_insert`, `_remove!`, `_traverse`) 
 - refactored: removed `@boundscheck`: traversal (=indexing) is O(n), so the `@boundscheck` are not going to be a big deal.
 - improvement: depending on the index it will either traverse the list from front or from the back 
 - improvement: `show()` is now limited to the space given and has the same apperance as `Vector`
 - added: verbose, readable tests for functions and a comparison of all functions against `Vector{T}()`
 - added: `append!(l,collections...)` 
 - added: `filter!(f,l)`
 - added: `reverse!(l)`
 - added: `empty!(l)`
 - added: `splice!(l,idx)` 
 - added: `splice!(l,range)` there is another PR for splice! but I found that it was flawed 
 - added: constructor `MutableLinkedList(iter)` without a type parameter infers the eltype from `iter`
 - changed: `last` and `first` return a BoundsError if the list is empty instead of an `ArgumentError` to mirror `Vector`
 - changed: `popat!(l,idx)` throws a `BoundsError` if the list is emtpy instead of an `ArgumentError` to mirror `Vector`
 - deprecated: `delete!` has been deprecated to `deleteat!`
 - fixed: allow i:j with j <= i for `delete`
 - fixed: `append!()` will not mutate the last element 

Fixes 
 - https://github.com/JuliaCollections/DataStructures.jl/issues/794
 - https://github.com/JuliaCollections/DataStructures.jl/issues/795
 - https://github.com/JuliaCollections/DataStructures.jl/issues/739
 - https://github.com/JuliaCollections/DataStructures.jl/issues/704

Supersedes:
 - https://github.com/JuliaCollections/DataStructures.jl/pull/871
 - https://github.com/JuliaCollections/DataStructures.jl/pull/818

In case this gets merge it needs to be squashed, the single commits aren't useful on their own.

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -23,6 +23,18 @@ Base.@deprecate_binding IntDisjointSets IntDisjointSet
 @deprecate insert!(m::SortedDict, k, d) push_return_semitoken!(m::SortedDict, k=>d)
 @deprecate insert!(m::SortedMultiDict, k, d) (push_return_semitoken!(m::SortedMultiDict, k=>d))[2]
 
+function Base.delete!(l::MutableLinkedList, idx)
+    Expr(:meta, :noinline)
+    Base.depwarn("`delete!(l::MutableLinkedList,idx::Int)` is deprecated, use `deleteat!(l,idx)` instead.", :delete!)
+    deleteat!(l,idx)
+end
+
+function Base.delete!(l::MutableLinkedList, r::UnitRange)
+    Expr(:meta, :noinline)
+    Base.depwarn("`delete!(l::MutableLinkedList,r::UnitRange)` is deprecated, use `deleteat!(l,idx)` instead.", :delete!)
+    deleteat!(l,r)
+end
+
 function Base.peek(q::PriorityQueue)
     Expr(:meta, :noinline)
     Base.depwarn("`peek(q::PriorityQueue)` is deprecated, use `first(q)` instead.", :peek)
--- a/src/mutable_list.jl
+++ b/src/mutable_list.jl
@@ -2,16 +2,14 @@ mutable struct ListNode{T}
     data::T
     prev::ListNode{T}
     next::ListNode{T}
-    function ListNode{T}() where T
+    function ListNode{T}() where {T}
         node = new{T}()
-        node.next = node
-        node.prev = node
+        node.next = node.prev = node
         return node
     end
-    function ListNode{T}(prev,data,next) where T
-        node = new{T}(data)
-        node.next = next
-        node.prev = prev
+    function ListNode{T}(prev, data, next) where {T}
+        node = new{T}(data, prev, next)
+        prev.next = next.prev = node
         return node
     end
 end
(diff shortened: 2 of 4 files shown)


src/deprecations.jl
```diff
@@ -23,6 +23,18 @@ Base.@deprecate_binding IntDisjointSets IntDisjointSet
...
+    Expr(:meta, :noinline)
+    Base.depwarn("`delete!(l::MutableLinkedList,idx::Int)` is deprecated, use `deleteat!(l,idx)` instead.", :delete!)
+    deleteat!(l,idx)
+end
+
+function Base.delete!(l::MutableLinkedList, r::UnitRange)
+    Expr(:meta, :noinline)
+    Base.depwarn("`delete!(l::MutableLinkedList,r::UnitRange)` is deprecated, use `deleteat!(l,idx)` instead.", :delete!)
+    deleteat!(l,r)
+end
```
oxinabox, line 36: Why can't this be done with the `@deprecate` macro?

src/mutable_list.jl
```diff
@@ -2,71 +2,95 @@ mutable struct ListNode{T}
...
+MutableLinkedList{T}(elts...) where {T} = append!(MutableLinkedList{T}(), elts)
 
 Base.iterate(l::MutableLinkedList) = l.len == 0 ? nothing : (l.node.next.data, l.node.next.next)
 Base.iterate(l::MutableLinkedList, n::ListNode) = n === l.node ? nothing : (n.data, n.next)
 
 Base.isempty(l::MutableLinkedList) = l.len == 0
 Base.length(l::MutableLinkedList) = l.len
-Base.collect(l::MutableLinkedList{T}) where T = T[x for x in l]
-Base.eltype(::Type{<:MutableLinkedList{T}}) where T = T
+Base.collect(l::MutableLinkedList{T}) where {T} = T[x for x in l]
```
oxinabox, line 34: Do we need this?
I would have thought it would fall back to this if we didn't define it.

src/mutable_list.jl
```diff
@@ -2,71 +2,95 @@ mutable struct ListNode{T}
...
+function Base.:(==)(l1::MutableLinkedList{T}, l2::MutableLinkedList{T}) where {T}
     length(l1) == length(l2) || return false
     for (i, j) in zip(l1, l2)
         i == j || return false
     end
     return true
 end
 
-function Base.map(f::Base.Callable, l::MutableLinkedList{T}) where T
+function Base.map(f::Base.Callable, l::MutableLinkedList{T}) where {T}
```
oxinabox, line 93: It would be good to strip out a lot of these constraints that do not follow modern practices.

```suggestion
function Base.map(f, l::MutableLinkedList{T}) where {T}
```

src/mutable_list.jl
```diff
@@ -89,162 +113,227 @@ function Base.map(f::Base.Callable, l::MutableLinkedList{T}) where T
     end
 end
 
-function Base.filter(f::Function, l::MutableLinkedList{T}) where T
+function Base.filter(f::Function, l::MutableLinkedList{T}) where {T}
```
oxinabox, line 116:
```suggestion
function Base.filter(f, l::MutableLinkedList{T}) where {T}
```

oxinabox (approved):
minor feedback.
I have given you merge rights so merge this when you are happy and mark this file done in the issue

laborg:
> minor feedback. I have given you merge rights so merge this when you are happy and mark this file done in the issue

Thanks for the review. As soon as I've time I'll look into your recommendations. In the meantime: I can't merge the PR via user interface. Shouldn't this work since you gave me merge rights?

JordiManyer:
Hi! Is this PR alive? 

I am running into issues, where `append!` for linked lists is still broken on the relase version v0.18.20 but not broken on the master branch. Is there any intention of merging this at some point? Or releasing a version of the package where the `append!` but is fixed? 

Thank you!

PR #880, opened 2023-11-09
Impl, docs, & tests for searchsortedbefore

### LINKED ISSUE #879: searchsortedbefore - reverse of searchsortedafter
For sorted containers, it'd be nice if there was a function analogous to searchsortedafter that instead got the token of the last item that is < a given key. In more detail, searchsortedfirst (first item >= key) is to searchsortedafter (first item > key) as searchsortedlast (last item <= key) would be to... searchsortedbefore? That'd probably be an appropriate name. 
I'm willing to write a PR if there's interest in this. Seems like it'd be very simple considering prevloc0 is available in the internals. Thoughts?

devinallen:
Fixes #879 
I pretty much just looked to searchsortedafter as a template for what tests to write
Otherwise, pretty straightforward
Thanks :D


StephenVavasis (approved):
This looks good-- fits into the existing codebase well.

PR #883, opened 2023-12-04
fix append! iterable to MutableLinkedList

chengchingwen:
Currently `append!(MutableLinkedList{Int}(), [3])` would try to push the vector into the linked list instead of pushing elements into the linked list.


chengchingwen:
@oxinabox Could you help review this PR?

PR #888, opened 2023-12-31
Empty head block's data when the block is reset

### LINKED ISSUE #884: Deque leaks memory

eulerkochy:
Fixes #884

### DIFF
--- a/src/deque.jl
+++ b/src/deque.jl
@@ -39,7 +39,7 @@ isrear(blk::DequeBlock) =  blk.next === blk
 
 function reset!(blk::DequeBlock{T}, front::Int) where T
     empty!(blk.data)
-    blk.data = Vector{T}(undef, blk.capa)
+    resize!(blk.data, blk.capa)
     blk.front = front
     blk.back = front - 1
     blk.prev = blk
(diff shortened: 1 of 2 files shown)


src/deque.jl
```diff
@@ -38,6 +38,8 @@ isrear(blk::DequeBlock) =  blk.next === blk
 # reset the block to empty, and position
 
 function reset!(blk::DequeBlock{T}, front::Int) where T
+    empty!(blk.data)
+    blk.data = Vector{T}(undef, blk.capa)
```
oxinabox, line 42: Why empty it if its just going to be over written with a newly allocated array?
I would think just:
```suggestion
    empty!(blk.data)
```
would be enough.
(and it's nonallocating)

eulerkochy: The underlying the assumption of a block is that it's data field will have already `data` field allocated with `capa` number of elements. Just emptying and not allocating block's `data` leads to the problem that once you have cleared the deque, you can't push elements into it again. That's why I've added new testset `push! after empty!` to check this scenario.

oxinabox: So we want:
```suggestion
    # Fill it with undefs back to capacity
    empty!(blk.data); resize!(blk.data, blk.capa)
```
?

eulerkochy: yes. made this change

oxinabox (approved):

PR #890, opened 2024-01-06
Robindict with metadata

eulerkochy:


fingolfin:
From the PR description it is not clear to me what this is about. Could you perhaps elaborate a bit on that?

Issue #891, opened 2024-01-21
Specify buffer length as Integer of any size


KronosTheLate:
This is easiest to show in code. I will make a fresh environment and try to initialize a circular buffer:
```julia
(@v1.10) pkg> activate --temp
  Activating new project at `/tmp/jl_7Wbc1W`

(jl_7Wbc1W) pkg> add DataStructures
    Updating registry at `~/.julia/registries/General.toml`
   Resolving package versions...
    Updating `/tmp/jl_7Wbc1W/Project.toml`
⌃ [864edb3b] + DataStructures v0.18.15
    Updating `/tmp/jl_7Wbc1W/Manifest.toml`
⌃ [34da2185] + Compat v4.10.1
⌃ [864edb3b] + DataStructures v0.18.15
  [bac558e1] + OrderedCollections v1.6.3
  [2a0f44e3] + Base64
  [b77e0a4c] + InteractiveUtils
  [d6f4376e] + Markdown
  [9a3f8284] + Random
  [ea8e919c] + SHA v0.7.0
  [cf7118a7] + UUIDs
        Info Packages marked with ⌃ have new versions available and may be upgradable.

julia> using DataStructures
```

But using `Int32` or `Int16` errors:
```julia
julia> CircularBuffer{Float32}(Int32(10))
ERROR: MethodError: no method matching CircularBuffer{Float32}(::Int32)

Closest candidates are:
  CircularBuffer{T}(::Int64) where T
   @ DataStructures ~/.julia/packages/DataStructures/MKv4P/src/circular_buffer.jl:16

Stacktrace:
 [1] top-level scope
   @ REPL[10]:1

julia> CircularBuffer{Float32}(Int16(10))
ERROR: MethodError: no method matching CircularBuffer{Float32}(::Int16)

Closest candidates are:
  CircularBuffer{T}(::Int64) where T
   @ DataStructures ~/.julia/packages/DataStructures/MKv4P/src/circular_buffer.jl:16

Stacktrace:
 [1] top-level scope
   @ REPL[11]:1
```

Int64 works as expected:
```julia
julia> CircularBuffer{Float32}(Int64(10))
0-element CircularBuffer{Float32}
```

This could easily be fixed, by I wanted to ask some questions before making a PR:
- Do we agree that this is unwanted behaviour?
- Is it possible that this should be fixed for a class of structs defined in this package, or is it likely to only affect CircularBuffers?

oxinabox:
yes we want this.
[...]

PR #893, opened 2024-01-25
Allow other int types for type container size

KronosTheLate:
This PR aims to solve https://github.com/JuliaCollections/DataStructures.jl/issues/891, and check if this issue occurs for other types as well. For the types listed [in the documentation](), the results are seen in the drop-down below.

<details>
Deque (implemented with an unrolled linked list)
errors on `Deque{Float64}(Int32(10))`. Fixed

CircularBuffer
Silent error on CircularBuffer{Float32}(Int32(10)). Fixed

CircularDeque (based on a circular buffer)
errors on CircularDeque{Float32}(Int32(10)). Fixed

Stack
errors on Stack{Float64}(Int32(5)). Got fixed, I think with Deque

Queue
errors on Queue{Float64}(Int32(5)). Got fixed, I think with Deque

Priority Queue
No length parameter upon initialization

Fenwick Tree
Works for FenwickTree{Float64}(Int32(5)) and FenwickTree{Float64}(Int16(5))

No length parameter upon initialization for the following types
    Accumulators and Counters (i.e. Multisets / Bags)
    Disjoint Sets
    Binary Heap
    Mutable Binary Heap
    Ordered Dicts and Sets
    RobinDict and OrderedRobinDict (implemented with Robin Hood Hashing)
    SwissDict (inspired from SwissTables)
    Dictionaries with Defaults
    Trie
    Linked List and Mutable Linked List
    Sorted Dict, Sorted Multi-Dict and Sorted Set
    DataStructures.IntSet
    SparseIntSet
    DiBitVector
    Red Black Tree
    AVL Tree
    Splay Tree
</details>

### DIFF
--- a/src/circ_deque.jl
+++ b/src/circ_deque.jl
@@ -12,6 +12,7 @@ end
 Create a double-ended queue of maximum capacity `n`, implemented as a circular buffer. The element type is `T`.
 """
 CircularDeque{T}(n::Int) where {T} = CircularDeque(Vector{T}(undef, n), n, 0, 1, n)
+CircularDeque{T}(n::Integer) where {T} = CircularDeque(Vector{T}(undef, Int(n)), Int(n), 0, 1, Int(n))
 
 Base.length(D::CircularDeque) = D.n
 Base.eltype(::Type{CircularDeque{T}}) where {T} = T
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -19,18 +19,21 @@ mutable struct CircularBuffer{T} <: AbstractVector{T}
         len <= length(buf) || throw(ArgumentError("Value of 'length' must be <= length of buffer"))
         return new{T}(length(buf), f, len, buf)
     end
+
+    # Convert any `Integer` to whatever `Int` is on the relevant machine
+    CircularBuffer{T}(f::Integer, len::Integer, buf::Integer) where {T} = CircularBuffer{T}(Int(f), Int(len), Int(buf))
 end
 
-function CircularBuffer{T}(iter, capacity::Int) where {T}
+function CircularBuffer{T}(iter, capacity::Integer) where {T}
     vec = copyto!(Vector{T}(undef,capacity), iter)
     CircularBuffer{T}(1, length(iter),vec)
 end
 
-CircularBuffer(capacity::Int) = CircularBuffer{Any}(capacity)
+CircularBuffer(capacity::Integer) = CircularBuffer{Any}(capacity)
 
-CircularBuffer{T}(capacity::Int) where {T} = CircularBuffer{T}(T[],capacity)
+CircularBuffer{T}(capacity::Integer) where {T} = CircularBuffer{T}(T[],capacity)
 
-CircularBuffer(iter,capacity::Int) =  CircularBuffer{eltype(iter)}(iter,capacity)
+CircularBuffer(iter,capacity::Integer) =  CircularBuffer{eltype(iter)}(iter,capacity)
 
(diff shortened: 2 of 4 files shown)


src/circ_deque.jl
```diff
@@ -12,6 +12,7 @@ end
 Create a double-ended queue of maximum capacity `n`, implemented as a circular buffer. The element type is `T`.
 """
 CircularDeque{T}(n::Int) where {T} = CircularDeque(Vector{T}(undef, n), n, 0, 1, n)
+CircularDeque{T}(n::Integer) where {T} = CircularDeque(Vector{T}(undef, Int(n)), Int(n), 0, 1, Int(n))
```
oxinabox, line 15: I think we can just do
```suggestion
CircularDeque{T}(n::Integer) where {T} = CircularDeque(Vector{T}(undef, n), n, 0, 1, n)
```
since constructors automatically call `convert` on all arguments.

src/circular_buffer.jl
```diff
@@ -19,18 +19,21 @@ mutable struct CircularBuffer{T} <: AbstractVector{T}
         len <= length(buf) || throw(ArgumentError("Value of 'length' must be <= length of buffer"))
         return new{T}(length(buf), f, len, buf)
     end
+
+    # Convert any `Integer` to whatever `Int` is on the relevant machine
+    CircularBuffer{T}(f::Integer, len::Integer, buf::Integer) where {T} = CircularBuffer{T}(Int(f), Int(len), Int(buf))
```
oxinabox, line 24: Per above i don't think this is needed since constructors automatically convert.
Even if you have another inner constructor.
```suggestion
```

src/deque.jl
```diff
@@ -20,13 +20,16 @@ mutable struct DequeBlock{T}
         blk.next = blk
         return blk
     end
+
+    # Convert any `Integer` to whatever `Int` is on the relevant machine
+    DequeBlock{T}(capa::Integer, front::Integer) where T = DequeBlock{T}(Int(capa), Int(front))
```
oxinabox, line 25: Per above
```suggestion
```

src/circ_deque.jl
KronosTheLate: I though about that option, but it somehow seemed more clear to me to explicitly do what will happen automatically. It somehow makes the code more readable to me, because it shows it's purpose more clearly. But I do not have strong opinions, so if you prefer this, it is fine by me ^_^

KronosTheLate:
Regarding your reviews, I though you were mainly changing my ecplixit conversion via `Int(arg)`. But I had to add the lines in the first place to make it work. Have you tested your suggestions?

src/circ_deque.jl
oxinabox: I see your point (getting rid of this behavour has actually been talked about for julia 2.0).

On the other hand, code that doesn't exist can't have bugs.

oxinabox:
> Have you tested your suggestions?

I have not.
If you say they proved nesc then I will go with that.
As my opinon is not strong

oxinabox (approved):

timholy:
Fun fact: sometimes it's useful to use the `Foo(n::Integer, args...) = Foo(Int(n), args...)` pattern even for constructors that perform automatic conversion. The reason? It confines specialization-diversity to a short "stub" function, and the "long function" (the one that does the actual work) gets compiled for a narrow range of types. I.e., less latency.

That said, here the constructors are short so I don't think it's a big deal either way.

KronosTheLate:
> Do you need this backported to 0.18?

If you are asking me, then no ^_^

Issue #894, opened 2024-02-02
Base might export `isfull` in non-breaking release


KronosTheLate:
I just submitted [this](https://github.com/JuliaLang/julia/pull/53159) PR to julia base, which will be breaking for DataStructures.jl if merged.

In order to maintain compatability with both 1.10 and 1.11, I propose we implement something like the following:
```
# A function `isfull` was introduced in Julia 1.10, 
# despite being already exported here. The code 
# below makes a conditional definition of `isfull` to 
# maintain simultaneous compatability with 1.10 
# and later versions. Particularly relevant if 1.10 
# indeed becomes next LTS
if isdefined(Base, :isfull)
    @eval DataStructures begin
        # import and extend `Base.isfull`
    end
else
    @eval DataStructures begin
        # insert old definition here
    end
end
```

vtjnash:
Can you make that a PR? I think all you need is:
```julia
if isdefined(Base, :isfull)
    import Base: isfull
end
```

CameronBieganek:
This might qualify as punning. The docstring for `Base.isfull` is very specific to Channels---it does not provide a generic definition of `isfull` that could also apply to `DataStructures.CircularBuffer`

KronosTheLate:
I am not familiar with the concept of "punning", but you do have a point. However I feel as if the meaning of `isfull` is quite clear from the name, even if it has currently only been deemed usefull for channels in Base.

CameronBieganek:
"Punning" is using the same generic function in two different contexts with two different meanings. "Is full?" seems clear enough, but it would still need a docstring with a generic definition. However, you have to ask yourself if there are ever any circumstances where it makes sense to write a generic function `foo`, like this,

```julia
function foo(x)
    # ...
    ... isfull(x) ...
    # ....
end
```

that works on both `Channel` and `DataStructures.CircularBuffer`. (I doubt that there are.) If there are no plausible generic use cases like that, then `Base.isfull` and `DataStructures.isfull` should probably be separate functions. (In other words, DataStructures should not overload `Base.isfull`.)

All that being said, the name `isfull(c::Channel)` seems to be in line with the names of the other functions in Base that operate on channels, so it's probably ok to add `Base.isfull`. It might have been nice if the Channel functions had been hidden inside a `Channel` namespace, but that ship sailed a long time ago.

vtjnash:
[...]

PR #896, opened 2024-02-06
Conditional import of Base.isfull

KronosTheLate:
As suggested in [this](https://github.com/JuliaCollections/DataStructures.jl/issues/894#issuecomment-1925922689) comment in #894

### DIFF
--- a/src/circular_buffer.jl
+++ b/src/circular_buffer.jl
@@ -207,6 +207,11 @@ Return capacity of CircularBuffer.
 """
 capacity(cb::CircularBuffer) = cb.capacity
 
+# Base might define `isfull` from julia 1.11 onwards, see julia PR #53159
+if isdefined(Base, :isfull)
+    import Base: isfull
+end
+
 """
     isfull(cb::CircularBuffer)
 


oxinabox:
Will merge once  https://github.com/JuliaLang/julia/pull/53159 is merged

CameronBieganek:
This might qualify as punning. The docstring for `Base.isfull` is very specific to Channels---it does not provide a generic definition of `isfull` that could also apply to `DataStructures.CircularBuffer`.

vtjnash:
Do you want to suggest a change on the Base PR? I think this function is very much the same concept as a general queue/stack implementation concept

CameronBieganek:
I wouldn't recommend overloading `Base.isfull`. There is no indication that code written for `Base.isfull(c::Channel)` would have any applicability to `DataStructures.CircularBuffer`. They should remain separate functions.

The introduction of `Base.isfull` does not break `DataStructures` per se, it just introduces ambiguity errors in downstream code like this:

```julia
using DataStructures
x = # ...
isfull(x)
```

That's unfortunate, but the real issue is with the "import all" pattern. For code that needs to survive minor updates to dependencies, you need to write it like this instead:

```julia
using DataStructures: isfull
x = # ...
isfull(x)
```

KronosTheLate:
> Will merge once [JuliaLang/julia#53159](https://github.com/JuliaLang/julia/pull/53159) is merged

That PR got merged yesterday.

oxinabox:
merged and backported

PR #897, opened 2024-02-16
Use `Base._unsetindex!` in `pop!` and `popfirst!`

### LINKED ISSUE #884: Deque leaks memory

kpamnany:
For `Deque` and `CircularDeque`. So that popped elements are not rooted by the deque and can be GCed when they drop out of caller scope.

Completes the fix of #884.

### DIFF
--- a/src/circ_deque.jl
+++ b/src/circ_deque.jl
@@ -64,6 +64,7 @@ end
 
 @inline Base.@propagate_inbounds function Base.pop!(D::CircularDeque)
     v = last(D)
+    Base._unsetindex!(D.buffer, D.last) # see issue/884
     D.n -= 1
     tmp = D.last - 1
     D.last = ifelse(tmp < 1, D.capacity, tmp)
@@ -91,6 +92,7 @@ Remove the element at the front.
 """
 @inline Base.@propagate_inbounds function Base.popfirst!(D::CircularDeque)
     v = first(D)
+    Base._unsetindex!(D.buffer, D.first) # see issue/884
     D.n -= 1
     tmp = D.first + 1
     D.first = ifelse(tmp > D.capacity, 1, tmp)
--- a/src/deque.jl
+++ b/src/deque.jl
@@ -297,6 +297,7 @@ function Base.pop!(d::Deque{T}) where T
     @assert rear.back >= rear.front
 
     @inbounds x = rear.data[rear.back]
+    Base._unsetindex!(rear.data, rear.back) # see issue/884
     rear.back -= 1
     if rear.back < rear.front
         if d.nblocks > 1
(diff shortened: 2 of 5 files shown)


src/deque.jl
```diff
@@ -322,6 +323,7 @@ function Base.popfirst!(d::Deque{T}) where T
     @assert head.back >= head.front
 
     @inbounds x = head.data[head.front]
+    Base._unsetindex!(head.data, head.front)
```
nickrobinson251, line 326: i wonder if these lines are worthy of a comment? maybe even just 
```suggestion
    Base._unsetindex!(head.data, head.front) # see issue/884
```

src/deque.jl
```diff
@@ -297,6 +297,7 @@ function Base.pop!(d::Deque{T}) where T
     @assert rear.back >= rear.front
 
     @inbounds x = rear.data[rear.back]
+    Base._unsetindex!(rear.data, rear.back)
```
nickrobinson251, line 300:
```suggestion
    Base._unsetindex!(rear.data, rear.back) # see issue/884
```

nickrobinson251 (approved):
Thanks, @kpamnany ! LGTM

@oxinabox are you still somewhat of a maintainer here? It'd be very helpful for us (at RAI) to get a new release with this fix 🙏  Or perhaps you could give me permissions on this repo? (I used to have them but never asked for them back when i lost all perms because of a MFA change) -- thanks, Frames!

oxinabox:
@nickrobinson251   I am still maintaining things here, but I am happy to hand this one over to you.
Go to https://github.com/orgs/JuliaCollections/ to accept your invite back to the org.

Usual procedure once happy with it: merge it into master, and backport it into `#0.18-release`,

nickrobinson251:
Also i wonder if we could add a test for this? Something based on checking `summarysize` perhaps?

src/circ_deque.jl
```diff
@@ -64,6 +64,7 @@ end
 
 @inline Base.@propagate_inbounds function Base.pop!(D::CircularDeque)
     v = last(D)
+    Base._unsetindex!(D.buffer, D.last)
```
kpamnany, line 67:
```suggestion
    Base._unsetindex!(D.buffer, D.last) # see issue/884
```

src/circ_deque.jl
```diff
@@ -91,6 +92,7 @@ Remove the element at the front.
 """
 @inline Base.@propagate_inbounds function Base.popfirst!(D::CircularDeque)
     v = first(D)
+    Base._unsetindex!(D.buffer, D.first)
```
kpamnany, line 95:
```suggestion
    Base._unsetindex!(D.buffer, D.first) # see issue/884
```

src/circ_deque.jl
nickrobinson251, line 67:
```suggestion
    Base._unsetindex!(D.buffer, D.last)
```

PR #898, opened 2024-02-20
Backport deque memory fix

### LINKED ISSUE #884: Deque leaks memory

nickrobinson251:
- backport #897 to v0.18 (fixes #884)

cc @kpamnany

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataStructures"
 uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
-version = "0.18.16"
+version = "0.18.17"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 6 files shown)


nickrobinson251:
lol the `release-0.18` branch has different CI, and tests against Julia v1.10... which fails! (so that's a good catch)

nickrobinson251:
oooohhh, it's not Julia v1.10 is v1.0 that's failing!

nickrobinson251:
`_unsetindex!` isn't defined til v1.3

https://github.com/JuliaLang/julia/commit/6f5cefb1d231a267eb5e0aa46469c0ef53593e28

nickrobinson251:
i've done the lazy fix and just defined `_unsertindex!` as a no-op on Julia v1.0 - v1.2

if an alternative (like drop support for pre-v1.3) or a better fix (actually implement `_unsetindex!` for pre-v1.3) is preferred, please let me know (cc @oxinabox)

PR #907, opened 2024-04-17
Improve performance of sizehint! on BinaryHeap

Tortar:
Noticed that sizehint! was actually worsening a lot of the performance of some of my heaps, with this change it is much faster:

e.g. with

```julia
using DataStructures, BenchmarkTools
function with_sizehint(T, n)
    value = BinaryHeap(Base.Order.ForwardOrdering(), T[])
    sizehint!(value, n)
end
```

before:

```julia
julia> @btime with_sizehint(Int, 1);
  861.250 ns (5 allocations: 256 bytes)
```

after:

```julia
julia> @btime with_sizehint(Int, 1);
  64.790 ns (4 allocations: 224 bytes)
```

### DIFF
--- a/src/heaps/arrays_as_heaps.jl
+++ b/src/heaps/arrays_as_heaps.jl
@@ -75,7 +75,7 @@ end
 
 In-place [`heapify`](@ref).
 """
-function heapify!(xs::AbstractArray, o::Ordering=Forward)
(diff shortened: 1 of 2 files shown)


Tortar:
Oh actually 1.6 doesn't support `@inline` at callsite, so I will inline it in the definition

PR #915, opened 2024-10-05
fix MutableLinkedList when appending empty list

foraspell:
Currently `append!(MutableLinkedList{Int}(1), MutableLinkedList{Int}())` results in an infinite iteration. This is due to the order of sentinel-node reassignments and the second list being empty. One way to fix that is by using something like `isempty(l2) && return l1` at the top of the `append!` method. Another (proposed in this PR) is to just reorder the reassignments. The former is more straightforward, the latter adds no new code.


yohmercy:
@oxinabox I apologize for tagging you directly: I'm new to contributing, so I'm not sure if that's OK. The changes here are minimal, yet having an infinite iteration every time I try to append an empty list is troublesome. I resort to using my own local version of the package in the meantime. Would you mind reviewing the PR? I'm not sure which of the two options is better. Thank you.

oxinabox:
tagging me is fine.
I am very busy right now.
I will try and find time for this next week.
If another maintainer doesn't come along and look first.

In general I hate the linked list code and want to just delete all of it.
In almost no real world senario is a linked list the right datastructure to use

fingolfin:
@yohmercy could you perhaps add a test that demonstrates the bug fix works (i.e. something that fails now and works correctly with your patch)

foraspell:
@fingolfin I added a test that checks if the last element in a nonempty list references itself as its successor. It fails without the fix.

PR #921, opened 2025-02-12
Update disjoint_set.jl

amontoison:
Improve `sizehint!(DisjointSets)` such that we can also preallocate the vectors needed by the internal `IntDisjointSets`.

### DIFF
--- a/src/disjoint_set.jl
+++ b/src/disjoint_set.jl
@@ -182,6 +182,8 @@ Base.empty(s::DisjointSet{T}, ::Type{U}=T) where {T,U} = DisjointSet{U}()
 function Base.sizehint!(s::DisjointSet, n::Integer)
     sizehint!(s.intmap, n)
     sizehint!(s.revmap, n)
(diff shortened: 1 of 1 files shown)


src/disjoint_set.jl
```diff
@@ -182,6 +182,8 @@ Base.empty(s::DisjointSet{T}, ::Type{U}=T) where {T,U} = DisjointSet{U}()
 function Base.sizehint!(s::DisjointSet, n::Integer)
     sizehint!(s.intmap, n)
     sizehint!(s.revmap, n)
+    sizehint!(s.internal.parents, n)
+    sizehint!(s.internal.ranks, n)
```
oxinabox, line 186: Can't we just do:
```suggestion
    sizehint!(s.internal, n)
```

amontoison: We can't because `sizehint!` is not defined for a `IntDisjointSet`.

oxinabox: We should define it, and then do this?

PR #922, opened 2025-02-12
Update TagBot.yml

amontoison:


oxinabox:
what is the motivation?

amontoison:
We don't see anymore the new releases on the main page of this github repository.
I am wondering if it's not related to the old version of `TagBot.yml`.
I just copy-pasted the new one.

PR #924, opened 2025-02-22
Fix splay tree amortized time issues (resolves #923)

### LINKED ISSUE #923: Splay tree amortized time issues
Hello,

Thanks for your work on this package !

The documentation at https://juliacollections.github.io/DataStructures.jl/stable/splay_tree/#Splay-Tree-1 indicates the splay tree implementation should guarantee

> Operations such as search, insert and delete can be done in O(log n) amortized time, where n is the number of nodes in the SplayTree

But I think this can be violated by the current implementation because it doesn't splay on unsuccessful searches, or on redundant insertions.

Here's a short test script to show the behavior; I have a particularly slow computer so you might need to increase the "10000" to see useful timing info:
```
using DataStructures

function test(N)
    if true
        t = SplayTree{Int}()
    else
        t = AVLTree{Int}()
    end

    for i in 1:N
        push!(t, i)
    end

    for i in 1:N
        if true
            push!(t, 1)
        else
            push!(t, -i)
        end
    end
end

for i in 1:10
    @time test(i * 10000)
end
```

In its default configuration, this:
1. Pushes 1, 2, 3, ..., N, which (correctly) creates a splay tree that's just one long list (root is N, left child N-1, left child N-2, etc.).
2. Tries to repeatedly re-push 1, which is already in the data structure. This causes it to traverse the entire length-N chain, but since the implementation doesn't splay on redundant pushes no reorganization happens and so this expensive operation happens every single time we search.

The result (at least on my machine) is very poor scaling behavior.

Notably, it is significantly faster if an AVL tree is used or if we push nonredundant elements, so I think the explanation above of the cause of the slowdown is correct. And I think it also affects, e.g., sequences of unsuccessful `haskey`s (replace `push!(t, 1)` with `haskey(t, 0)`).

I think the basic fix is to have `search_node` always splay the very last node it encounters, as suggested in the splay tree paper. Alternatively, if this is expected behavior, it would be nice IMO to document that. If this package is still maintained + there's interest in either of those two changes I'm happy to send a PR.

Also sorry if I've misunderstood something about the implementation or use of the package; I'm new to Julia.

matthewsot:
Per discussion in #923, this PR ensures that the splay tree's `search_node!` now always splays the last-seen node. This means that `push!` will reorganize the tree even if the push is already present, and `haskey` will reorganize the tree even if the key is not present. This resolves a performance issue.

In addition, I think I found a correctness bug in `delete!`, where it doesn't check that the node returned by `search_node!` actually has the requested key (`search_node!` seems to return the last visited node during the traversal, whether or not it was actually the node we searched for). I fixed this and added a regression test. You can check that it fails on the old splay tree implementation using this branch: https://github.com/matthewsot/DataStructures.jl/tree/regression

A few things that came up in the process that I'm not 100% sure how you all want to handle:
1. It seems like `search_node` was part of the public API, but now it's been renamed to `search_node!` for splay trees as suggested in the discussion in #923, which could be a breaking change. Is that intended?
2. `splay_tree.jl` had a few places using `== nothing` and a few other places using `=== nothing`. I'm not totally sure what the difference is, so I tried to basically just do whatever the most similar code was doing before.

Feedback welcome!

### DIFF
--- a/benchmark/bench_trees.jl
+++ b/benchmark/bench_trees.jl
@@ -1,0 +1,69 @@
+module BenchTrees
+
+using DataStructures
+using BenchmarkTools
+using Random
+
+suite = BenchmarkGroup()
+
+rand_setup =  (
+	Random.seed!(1234);
+	idx1 = rand(1:30000, 1000);
+	didx1 = rand(1:1000, 500);
+)
+
+# insert a bunch of keys, search for all of them, delete half of them randomly,
+# then search for all of them again.
+function test_rand(T)
+    t = T{Int}()
+    for i in idx1
+        push!(t, i)
+    end
+    for i in idx1
+        haskey(t, i)
+    end
+    for i in didx1
+        delete!(t, idx1[i])
+    end
+    for i in idx1
+        haskey(t, i)
+    end
+end
+
+# insert 1, ..., N, then push 1 many times. tests a regression from an older
+# splay tree implementation where splays didn't happen on redundant pushes.
+function test_redundant(T, N=100000)
+    t = T{Int}()
+    for i in 1:N
+        push!(t, i)
+    end
+    for i in 1:N
+        push!(t, 1)
+    end
+end
+
+# insert 1, ..., N, then access element 1 for N iterations. splay trees should
+# perform best here.
+function test_biased(T, N=100000)
+    t = T{Int}()
+    for i in 1:N
+        push!(t, i)
+    end
+    for _ in 1:N
+        haskey(t, 1)
+    end
+end
+
+trees = Dict("splay" => SplayTree, "avl" => AVLTree, "rb" => RBTree)
+for (T_name, T) in trees
+    suite[T_name]["rand"] =
+        @benchmarkable test_rand($(T)) setup=rand_setup
+    suite[T_name]["redundant"] =
+        @benchmarkable test_redundant($(T))
+    suite[T_name]["biased"] =
+        @benchmarkable test_biased($T)
+end
+
+end  # module
+
+BenchTrees.suite
--- a/src/splay_tree.jl
+++ b/src/splay_tree.jl
@@ -126,7 +126,7 @@ function _join!(tree::SplayTree, s::Union{SplayTreeNode, Nothing}, t::Union{Spla
     end
 end
 
-function search_node(tree::SplayTree{K}, d::K) where K
+function search_node!(tree::SplayTree{K}, d::K) where K
     node = tree.root
     prev = nothing
     while node != nothing && node.data != d
@@ -137,7 +137,9 @@ function search_node(tree::SplayTree{K}, d::K) where K
             node = node.leftChild
         end
     end
-    return (node == nothing) ? prev : node
+    last = (node == nothing) ? prev : node
+    (last == nothing) || splay!(tree, last)
+    return last
 end
 
 function Base.haskey(tree::SplayTree{K}, d::K) where K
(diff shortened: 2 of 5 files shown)


src/splay_tree.jl
```diff
@@ -126,7 +126,7 @@ function _join!(tree::SplayTree, s::Union{SplayTreeNode, Nothing}, t::Union{Spla
     end
 end
 
-function search_node(tree::SplayTree{K}, d::K) where K
+function search_node!(tree::SplayTree{K}, d::K) where K
```
eulerkochy, line 129: I gave this a thought. And perhaps now I recall the reason why I didn't splay while searching. 

Making `search_node` as a modifying function essentially means we are modifying the tree on every access. This means simple lookup using `in` or `haskey` - all such functions need to be appended with `!` indicating the change in stucture of the underlying tree. This will be a breaking change (and quite rightly so).

The context for use of splay trees implies you want recently accessed elements near to the root, so that the subsequent access takes less time. 

@oxinabox are we okay for introducing a breaking change in the package?

src/splay_tree.jl
```diff
@@ -137,33 +137,31 @@ function search_node(tree::SplayTree{K}, d::K) where K
...
+        node = search_node!(tree, d)
         (node === nothing) && return false
-        is_found = (node.data == d)
-        is_found && splay!(tree, node)
-        return is_found
+        return (node.data == d)
     end
 end
 
 Base.in(key, tree::SplayTree) = haskey(tree, key)
```
eulerkochy, line 156: Inline with my above comment, every function calling `search_node!` becomes modifying in nature.

benchmark/bench_trees.jl
eulerkochy, line 1: What are the results for this benchmark suite? Do post on the main PR.

eulerkochy (changes requested):

src/splay_tree.jl
oxinabox: Yes into the master branch that is fine. We are still prepping 0.19

matthewsot:
Thanks for the feedback, everyone!

Regarding benchmark results, here they are on my machine for the current `master` branch (without this PR): https://github.com/matthewsot/DataStructures.jl/blob/regression/old.md

Notice `["trees", "splay", "redundant"]` takes almost `45s`, whereas AVL and RB are in the millisecond range.

I'll post final benchmark results for this PR after resolving the other issues, but early results it seems like it brings the splay tree performance into the millisecond range as well.

src/splay_tree.jl
matthewsot: I tend to agree with @fingolfin here: none of these operations actually modify the tree in any way that's visible with user-facing set operations. Even an in-order traversal would not change after a splay (though pre/post-orders would, but I don't think those are exposed?).

I've pushed up a new version that reverts this change and leaves `search_node` un-`!`d. Thoughts on this direction @eulerkochy ? Or do you strongly prefer changing all of the accessors to `!`?

eulerkochy: I am happy with the un-`!` with being defined as something not changing the "user visible state". Let's  add a comment to that effect, so that it's remember why this decision was made, even though the "internal state" of the tree is changing.

src/splay_tree.jl
eulerkochy: This becomes moot, because of "user-visible state" definition.

eulerkochy (approved):
LGTM. Add the comment regarding the un-`!` being the user-visible state, and then we can merge. Also post the final benchmark results.

matthewsot:
Thanks for all the feedback!

In the latest commit I've added a comment about how `search_node` doesn't change the user-facing state of the set. I've also attached both the old and new benchmark runs to this comment. The main thing that changed is the ~1000x speedup in the `["trees", "splay", "redundant"]` test.

[old_benchmarks.md](https://github.com/user-attachments/files/19405324/old_benchmarks.md)
[new_benchmarks.md](https://github.com/user-attachments/files/19405325/new_benchmarks.md)

PR #929, opened 2025-03-17
Switch to Documenter 1.x

fingolfin:


docs/make.jl
```diff
@@ -5,6 +5,7 @@ DocMeta.setdocmeta!(DataStructures, :DocTestSetup, :(using DataStructures); recu
 
 makedocs(
     sitename = "DataStructures.jl",
+    warnonly = true,  # FIXME: address all warnings and resolve them
```
fingolfin, line 8: Had to turn off warnings, as there are "132 docstrings not included in the manual". If one does `warnonly = :exported` this goes down to 70+ but that's still quite a bit.

PR #933, opened 2025-03-24
Avoid unnecessary hash lookups in trie.jl

AzamatB:


src/trie.jl
```diff
@@ -53,11 +50,8 @@ end
 function subtrie(t::Trie, prefix)
     node = t
     for char in prefix
-        if !haskey(node.children, char)
-            return nothing
-        else
-            node = node.children[char]
-        end
+        node = get!(node.children, char, nothing)
```
jonas-schulze, line 53: You do not want to store `nothing` here.

```suggestion
        node = get(node.children, char, nothing)
```

AzamatB: Good catch! Fixed, thank you @jonas-schulze

oxinabox (approved):

PR #936, opened 2025-04-30
Refactor `not_iterator_of_pairs` to stop using `InteractiveUtils`

asinghvi17:
Jeff and Oscar agree this should be equivalent (and precompilable!).  It uses Base.isiterable instead.

This also adds a fast path checking `eltype` if at all possible.  Only if the eltype is not Any do we use the fast path.  But that should ideally also cut down the time taken to construct a priority queue.

### DIFF
--- a/src/dict_support.jl
+++ b/src/dict_support.jl
@@ -1,9 +1,6 @@
 # support functions
 
-using InteractiveUtils: methodswith
-
-function not_iterator_of_pairs(kv)
-    return any(x->isempty(methodswith(typeof(kv), x, true)),
-               [iterate]) ||
+function not_iterator_of_pairs(kv::T) where T
+    return Base.hasmethod(Base.iterate, Tuple{T}) &&
            any(x->!isa(x, Union{Tuple,Pair}), kv)
 end
--- a/src/dict_support.jl
+++ b/src/dict_support.jl
@@ -1,6 +1,6 @@
 # support functions
 
 function not_iterator_of_pairs(kv::T) where T
-    return Base.hasmethod(Base.iterate, Tuple{T}) &&
+    return Base.isiterable(kv) &&
            any(x->!isa(x, Union{Tuple,Pair}), kv)
 end
--- a/src/dict_support.jl
+++ b/src/dict_support.jl
@@ -1,6 +1,6 @@
 # support functions
 
 function not_iterator_of_pairs(kv::T) where T
-    return Base.isiterable(kv) &&
+    return Base.isiterable(T) &&
            any(x->!isa(x, Union{Tuple,Pair}), kv)
 end
--- a/src/dict_support.jl
+++ b/src/dict_support.jl
@@ -1,6 +1,16 @@
 # support functions
 
 function not_iterator_of_pairs(kv::T) where T
-    return !(Base.isiterable(T)) || # if the object is not iterable, return true, else check the eltype of the iteration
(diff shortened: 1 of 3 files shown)


src/dict_support.jl
```diff
@@ -1,9 +1,6 @@
 # support functions
 
-using InteractiveUtils: methodswith
-
-function not_iterator_of_pairs(kv)
-    return any(x->isempty(methodswith(typeof(kv), x, true)),
-               [iterate]) ||
+function not_iterator_of_pairs(kv::T) where T
+    return Base.hasmethod(Base.iterate, Tuple{T}) &&
```
asinghvi17, line 4:
```suggestion
    return Base.isiterable(kv) &&
```

src/dict_support.jl
```diff
@@ -1,9 +1,6 @@
 # support functions
 
-using InteractiveUtils: methodswith
-
-function not_iterator_of_pairs(kv)
-    return any(x->isempty(methodswith(typeof(kv), x, true)),
-               [iterate]) ||
+function not_iterator_of_pairs(kv::T) where T
+    return Base.isiterable(kv) &&
```
asinghvi17, line 4:
```suggestion
    return Base.isiterable(T) &&
```

src/dict_support.jl
```diff
@@ -1,9 +1,6 @@
 # support functions
 
-using InteractiveUtils: methodswith
-
-function not_iterator_of_pairs(kv)
-    return any(x->isempty(methodswith(typeof(kv), x, true)),
-               [iterate]) ||
+function not_iterator_of_pairs(kv::T) where T
+    return Base.isiterable(T) &&
```
asinghvi17, line 4:
```suggestion
    return !(Base.isiterable(T)) || # if the object is not iterable, return true, else check the eltype of the iteration
```

oxinabox (approved):

test/test_priority_queue.jl
```diff
@@ -105,6 +105,40 @@ import Base.Order.Reverse
...
+                Base.IteratorSize(i::EltypeUnknownIterator) = Base.IteratorSize(i.x)
+                Base.length(i::EltypeUnknownIterator) = Base.length(i.x)
+                Base.size(i::EltypeUnknownIterator) = Base.size(i.x)
+    
+                @test_nowarn PriorityQueue(Dict(zip(1:5, 2:6)))
+                @test_nowarn PriorityQueue(EltypeUnknownIterator(Dict(zip(1:5, 2:6))))
+                @test_throws ArgumentError PriorityQueue(EltypeUnknownIterator(['a']))
+            end
+
+            @testset "Eltype any"
```
oxinabox, line 125:
```suggestion
            @testset "Eltype any" begin
```

oxinabox:
It would be good if the coverage was complete

src/dict_support.jl
```diff
@@ -1,9 +1,16 @@
...
 
-using InteractiveUtils: methodswith
-
-function not_iterator_of_pairs(kv)
-    return any(x->isempty(methodswith(typeof(kv), x, true)),
-               [iterate]) ||
-           any(x->!isa(x, Union{Tuple,Pair}), kv)
+function not_iterator_of_pairs(kv::T) where T
+    # if the object is not iterable, return true, else check the eltype of the iteration
+    Base.isiterable(T) && return true
```
asinghvi17, line 5:
```suggestion
    Base.isiterable(T) || return true 
```

PR #949, opened 2025-08-20
Fix CircularBuffer inner constructor

rossviljoen:


src/circular_buffer.jl
```diff
@@ -14,29 +14,29 @@ mutable struct CircularBuffer{T} <: AbstractVector{T}
...
 
-    function CircularBuffer{T}(f,len,buf) where {T}
+    function CircularBuffer{T}(f::Int, len::Int, buf) where {T}
         f <= length(buf) || throw(ArgumentError("Value of 'first' must be inbounds of buffer"))
         len <= length(buf) || throw(ArgumentError("Value of 'length' must be <= length of buffer"))
         return new{T}(length(buf), f, len, buf)
     end
 
     # Convert any `Integer` to whatever `Int` is on the relevant machine
-    CircularBuffer{T}(f::Integer, len::Integer, buf::Integer) where {T} = CircularBuffer{T}(Int(f), Int(len), Int(buf))
```
rossviljoen, line 24: This constructor could be removed entirely, as it was just never called in the previous version. However it seems that https://github.com/JuliaCollections/DataStructures.jl/pull/893 decided to keep it

oxinabox:
What is this fixing?
Can it have a test added that was failing before and now passes?

PR #953, opened 2025-08-25
Prep for v0.19.1

odow:
Requires #951


odow:
I actually find it useful to leave the changelog until I come to do a release. Makes me have a second read over the changes

PR #959, opened 2025-10-24
Fix naming of DisjointSet in documentation and tests

m-fila:
`DisjointSet` and `IntDisjointSet` were renamed in #700 but the documentation was still referring to the old name


oxinabox (approved):
thanks

PR #954, opened 2025-09-08
Improve bounds checks in heap operations

Drvi:

PR #960, opened 2025-11-04
Don't require the element to be provided to `percolate_down!` and `percolate_up!`

Drvi:
Follow-up to #954, which made the third argument, the element (or the Ordering), required. By providing a default value for the Ordering, we no longer hit the edge case.


Drvi:
Thanks for the review! I can't merge the PR, so please, someone do it

PR #977, opened 2026-06-06
docs: document array-as-heap functions

### LINKED ISSUE #957: No documentation for `heappush!` and `heappop!` etc
https://github.com/JuliaCollections/DataStructures.jl/blob/c86628e81eddcb849d4cf733c8e002101906e95e/src/heaps/arrays_as_heaps.jl#L44-L65

these are not rendered in the docs apparently.

JohnCobbler:
`heappush!`, `heappop!`, `heapify!`, `heapify`, and `isheap` are exported with
docstrings but no manual page lists them, so they never render. Add an `@docs`
block for the five under a new "Array-as-heap functions" heading in `heaps.md`.

`heapify` also had no attached docstring at all: a `# Todo` comment between its
docstring and definition detached it, so the `@docs` entry and `heapify!`'s
`@ref` to it could not resolve. Move the comment above the docstring and refresh
its jldoctest output for current Julia.

Fixes #957

### DIFF
--- a/docs/src/heaps.md
+++ b/docs/src/heaps.md
@@ -153,6 +153,20 @@ this ordering:
 DataStructures.nextreme(Base.Forward, n, a) # Equivalent to nsmallest(n, a)
 ```
 
+# Array-as-heap functions
+
+These functions treat an `AbstractArray` directly as a binary heap, mirroring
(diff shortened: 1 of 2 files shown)


oxinabox (approved):
Nice. thankls

PR #983, opened 2026-07-25
Subtype IntSet and SparseIntSet to AbstractSet{Int}

### LINKED ISSUE #908: Intset doesn't subtype AbstractSet

JohnCobbler:
Neither `IntSet` nor `SparseIntSet` subtyped `AbstractSet`, so generic set
algorithms and APIs that dispatch on `AbstractSet` ignored them.

Make both `<: AbstractSet{Int}` and add the helpers Base's set fallbacks need:

- `emptymutable` / `copymutable` so `filter` and `setdiff` return the same
  concrete type instead of falling back to `Set`
- `filter!` via `unsafe_filter!` (BitSet pattern) so
  `intersect!(s::AbstractSet, s2::AbstractSet)` works
- `delete!` for `SparseIntSet` (was only `pop!`) for the filter! path
- `symdiff!(::IntSet, ::AbstractSet)` and `intersect!(::SparseIntSet, ::AbstractSet)`
  for unambiguous dispatch

Same-type `==` / `<` / `<=` methods remain more specific (unchanged). Cross-type
equality with `Set` / between the two types becomes available via AbstractSet
fallbacks.

Docs note the `AbstractSet{Int}` relationship.

Fixes #908


src/int_set.jl
```diff
@@ -133,6 +136,7 @@ end
 
 Base.symdiff(s::IntSet, ns) = symdiff!(copy(s), ns)
 Base.symdiff!(s::IntSet, ns) = (for n in ns; symdiff!(s, n); end; s)
+Base.symdiff!(s::IntSet, ns::AbstractSet) = (for n in ns; symdiff!(s, n); end; s)
```
oxinabox: is this needed just to break ambiguity?

src/sparse_int_set.jl
```diff
@@ -170,6 +174,7 @@ Base.intersect!(s1::SparseIntSet, ss...) = intersect!(s1, intersect(ss...))
 
 #Is there a more performant way to do this?
 Base.intersect!(s1::SparseIntSet, ns) = copy!(s1, intersect(s1, ns))
+Base.intersect!(s1::SparseIntSet, ns::AbstractSet) = copy!(s1, intersect(s1, ns))
```
oxinabox: again is this needed just to break ambiguity?

src/int_set.jl
oxinabox: It is tempting to write this as:

```suggestion
for T in (AbstractSet, Any)
    Base.symdiff!(s::IntSet, ns::T) = (for n in ns; symdiff!(s, n); end; s)
end
```
To make it clear that they are actually identical

JohnCobbler:
Yes — those twins exist only to break the diamond with Base's `AbstractSet` methods once we subtype (`(IntSet, AbstractSet)` is more specific on the left than `symdiff!(::AbstractSet, ::AbstractSet)` and more specific on the right than `symdiff!(::IntSet, ns)`). Same pattern as `BitSet`.

Pushed a small follow-up that writes each pair as `for T in (AbstractSet, Any)` so the identical bodies are obvious (and the same for `SparseIntSet` `intersect!`).

oxinabox:
Can we add a comment saying that we define both to break method errors for ambiguity?

JohnCobbler:
Done - comments on both the IntSet `symdiff!` and SparseIntSet `intersect!` twin loops.

oxinabox (approved):

